PR #43, opened 2020-03-20
fix: Malformed content app URL

cutwater:
* Use CONTENT_PATH_PREFIX setting from pulpcore to
  build content app URL.
* Fix missing X_PULP_CONTENT_HOST setting for development
  environment.

### DIFF
--- a/galaxy_ng/app/settings.py
+++ b/galaxy_ng/app/settings.py
@@ -32,6 +32,5 @@ X_PULP_API_USER = "admin"
 X_PULP_API_PASSWORD = "admin"
 X_PULP_API_PREFIX = "pulp_ansible/galaxy/automation-hub/api"
 
-X_PULP_CONTENT_HOST = "pulp-content"
+X_PULP_CONTENT_HOST = "localhost"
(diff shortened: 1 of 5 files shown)


galaxy_ng/app/settings.py
```diff
@@ -32,6 +32,5 @@
 X_PULP_API_PASSWORD = "admin"
 X_PULP_API_PREFIX = "pulp_ansible/galaxy/automation-hub/api"
 
-X_PULP_CONTENT_HOST = "pulp-content"
+X_PULP_CONTENT_HOST = "localhost"
```
alikins: Not sure I understand why X_PULP_CONTENT_HOST is "localhost" now?  I would have guessed 'content-app'.

Does the X_PULP_CONTENT_HOST defined here in settings.py always get overridden by the value in common/galaxy_ng.env ? And 'localhost' is just a default?

If it works, it works. Just curious.

cutwater: It is to have environment agnostic default configuration. Actual setting for development environment is set in `dev/common/galaxy_ng.env`

PR #55, opened 2020-03-24
On sha256 mismatch, raise ValidationError

alikins:
So the error is returned "insights style"
instead of exactly as returned from pulp.

Fixes ansible/galaxy-dev#290


alikins:
[WIP] because I haven't tested it really while doing some yak shaving


(if only yak hair was N95 compliant...)

chouseknecht:
This is over a year old. Assuming it's no longer relevant.

Issue #52, opened 2020-03-24
Infra: Generate custom Pulp bindings


chouseknecht:
Subtask of #51 

Generate bindings from https://github.com/ansible/galaxy_ng/blob/master/bindings/openapi.yaml and add to source control.

newswangerd:
Moved to issues.redhat.com/

PR #56, opened 2020-03-25
Add pre-generated galaxy_pulp bindings

cutwater:
This change adds pre-generated galaxy_pulp bindings to the
source control and removes building bindings at the
container image build step.

Issue: #52


.travis/install.sh
```diff
@@ -63,7 +63,6 @@ images:
       pulpcore: pulpcore
       plugins:
         - ./galaxy_ng
-        - ./galaxy_ng/bindings/galaxy-pulp
```
fabricio-aguiar: you have to remove it couple lines below also
https://github.com/ansible/galaxy_ng/blob/master/.travis/install.sh#L79

cutwater: missed that. now it's done.

setup.py
```diff
@@ -2,12 +2,23 @@
 
 from setuptools import find_packages, setup
 
-requirements = [
+# NOTE(cutwater): Because bindings are statically generated, requirements list
+#   from pulp-galaxy/setup.py has to be copied here and manually maintained.
+galaxy_pulp_requirements = [
```
alikins: +1 for separating the reqs

setup.py
```diff
@@ -2,12 +2,23 @@
 
 from setuptools import find_packages, setup
 
-requirements = [
+# NOTE(cutwater): Because bindings are statically generated, requirements list
+#   from pulp-galaxy/setup.py has to be copied here and manually maintained.
```
alikins: Would it make any sense to add a setup.py to galaxy_pulp and install as it's own package?

(Probably not worth it considering end goal is to not have galaxy_pulp at all, so no need to change)

alikins:
Also seems to work with some minimal testing. ie, `./compose build && ./compose up` and poking around the UI and api a little bit. +1

Issue #10, opened 2020-03-11
Infra: Migrate automated deployments to galaxy_ng


chouseknecht:
Subtask of #9 

Action items:

- [x] Prepare container images for being executed in OpenShift environment.
- [x] Upgrade deployment scripts to build container images from ansible/galaxy_ng repository.
- [x] Setup GitHub web hooks to trigger automatic builds of CI and QA environments.
- [x] Migrate CI and QA environments to galaxy_ng.

chouseknecht:
Remaining tasks:

- [ ] Enable metrics https://github.com/ansible/galaxy_ng/pull/74
- [ ] Enable Cloudwatch logs

PR #60, opened 2020-03-27
Container image build improvements for OpenShift

cutwater:
Improve container image build process for compatibility with
OpenShift execution environment.

Changes:

* Create linux user and group for container. User and group
  names as well as ID can be passed as parameters at the container
  build time. Group is not created if it already exists.

* Fix container build issues caused by combination of rpm cache and
  overlay2 filesystem layers and rpm cache. See [1] and [2] for more
  details.

* Rename docker/scripts to docker/bin and remove .sh extension
  from scripts intendet to be executed as commands.

* Add `wait-for-migrations` script, that waits until all database
  migrations are applied.

* Extract wait for TCP port logic from `entrypoint.sh` to standalone
  script `docker/bin/wait-for-tcp`.

* Disable by defailt second run of editable install at container startup
  time, which is used when container is running against local source
  code. It can be enabled by setting environment variable
  `WITH_DEV_INSTALL=1`.

* Disable execution migrations by default at container startup time.
  Automatic migrations execution may cause issue if two or more
  container instances trying to execute migrations concurrently.
  Automatic migrations execution can be enabled by setting environment
  variable `WITH_MIGRATIONS=1`.

References:
[1] https://bugs.centos.org/view.php?id=16934
[2] https://bugzilla.redhat.com/show_bug.cgi?id=1680124#c6

Issue: #10


cutwater:
@awcrosby It _may_ need to run editable install for the second time to generate `*.egg-info` directory on host if it's missing.

PR #63, opened 2020-03-27
Reproducible builds for container images

cutwater:
Lock project dependencies in `requirements.txt` file which is
used during container image build process.
    
Issue: #10
DependsOn: #60


release_requirements.txt
```diff
@@ -0,0 +1,75 @@
+#
+# This file is autogenerated by pip-compile
+# To update, run:
+#
+#    pip-compile --output-file=requirements.txt
```
awcrosby, line 5: s/requirements.txt/release_requirements.txt here and in `README.md`

PR #66, opened 2020-03-30
Introducing snippets for galaxy ui

fabricio-aguiar:
Adding entries on nginx/apache for UI URL
Depends on:
*  https://github.com/pulp/ansible-pulp/pull/262
*  https://github.com/pulp/pulp_installer/pull/271

### DIFF
--- a/galaxy_ng/app/webserver_snippets/apache.conf
+++ b/galaxy_ng/app/webserver_snippets/apache.conf
@@ -1,0 +1,9 @@
+ProxyPass /static/ http://${pulp-api}/static/
+ProxyPassReverse /static/ http://${pulp-api}/static/
+
+RewriteEngine  on
+RewriteCond "%{REQUEST_FILENAME}"       !-f
+RewriteCond "%{REQUEST_FILENAME}"       !-d
+RewriteRule "^/[a-zA-Z0-9_]*$" "http://${pulp-api}/static/galaxy_ng/index.html" [P]
+ProxyPass "/" "http://${pulp-api}/static/galaxy_ng/index.html"
+ProxyPassReverse "/" "http://${pulp-api}/static/galaxy_ng/index.html"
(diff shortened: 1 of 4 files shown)


galaxy_ng/app/webserver_snippets/apache.conf
```diff
@@ -0,0 +1 @@
+Alias / /var/lib/pulp/assets/galaxy_ng/
```
fabricio-aguiar, line 1: Just a placeholder, as it is not working and I haven't figured out how to do it on apache

galaxy_ng/app/webserver_snippets/nginx.conf
```diff
@@ -0,0 +1,3 @@
+location / {
+    alias /var/lib/pulp/assets/galaxy_ng/;
```
fabricio-aguiar, line 2: It works only with the change in https://github.com/pulp/pulp_installer/pull/268/files

bmbouter:
Adding in the snippets also requires entries in the MANIFEST. We have some docs on it, around step 3 here https://docs.pulpproject.org/plugins/plugin-writer/concepts/index.html#configuring-reverse-proxy-with-custom-urls

bmbouter:
Here's a recap of the working plan @newswangerd and I put together:

### Installer changes
* Update the installer to have `/` route to Django in all cases so Django can both serve static media and have plugin views/viewsets serve routes as short as `/`.

### Galaxy_ng changes
* Remove the snippets since this plan doesn't need them
* keep galaxy_ng/settings.py to set `STATIC_URL = "/static/"`
* Add a new galaxy_ng view that 302 redirects `/` to `/ui/`
* Add a new galaxy_ng view that matches `/ui/*` and 302 redirects to `/static/galaxy_ng/index.html`

fabricio-aguiar:
Closing in favor of https://github.com/ansible/galaxy_ng/pull/116

PR #69, opened 2020-04-01
Fix dev install for compose environment

cutwater:


alikins:
Ah, better!  That at least gets /api/automation-hub responding

PR #70, opened 2020-04-02
Fix pulp content url in dev env

shanemcd:
When following along with https://pulp-ansible.readthedocs.io/en/latest/workflows/collections.html, I hit a few bugs that required these variables to be set.


dev/common/settings.py
```diff
@@ -11,3 +11,6 @@
         'PASSWORD': os.environ.get('PULP_DB_PASSWORD', ''),
     }
 }
+
+ANSIBLE_API_HOSTNAME = 'http://localhost:5001'
+ANSIBLE_CONTENT_HOSTNAME = 'http://localhost:24816/api/automation-hub/v3/artifacts/collections'  # noqa: E501
```
cutwater, line 16: These settings are specific to automation hub environment. Though it is the only deployment scenario supported at the moment, I would appreciate if you put all automation-hub specific settings to `dev/automation-hub/galaxy_ng.env`.

Please note that all environment variables for pulp must begin with prefix `PULP_`.

Issue #64, opened 2020-03-27
Importer: ansible-test error inside pylint on openshift


awcrosby:
When `ansible-test sanity` runs during import, it runs in an openshift container and the `pylint` test is run. For collection `ibm.spectrum_virtualize`, `pylint` calls `multiprocessing.connection.recv` which hits an EOF error when unpickling an object.

Environment(s) seen with issue: AH Prod and AH QA
Environment(s) seen without issue: running importer in cli mode, running `ansible-test sanity -t pylint --docker` locally

Traceback from AH QA import:
```
...
Running sanity test 'pep8' with Python 3.6 
Running sanity test 'pylint' with Python 3.6 
ERROR: Command "/usr/bin/python3.6 -m pylint --jobs 0 --reports n --max-line-length 160 --rcfile /usr/local/lib/python3.6/dist-packages/ansible_test/_data/sanity/pylint/config/collection.cfg --output-format json --load-plugins blacklist,deprecated,string_format tests/unit/plugins/module_utils/test_ibm_svc_utils.py tests/unit/plugins/modules/test_ibm_svc_host.py tests/unit/plugins/modules/test_ibm_svc_info.py tests/unit/plugins/modules/test_ibm_svc_mdisk.py tests/unit/plugins/modules/test_ibm_svc_mdiskgrp.py tests/unit/plugins/modules/test_ibm_svc_vdisk.py tests/unit/plugins/modules/test_ibm_svc_vol_map.py" returned exit status 1. 
>>> Standard Error 
Process ChildLinter-2: 
Traceback (most recent call last): 
File "/usr/lib/python3.6/multiprocessing/process.py", line 258, in _bootstrap 
self.run() 
File "/usr/local/lib/python3.6/dist-packages/pylint/lint.py", line 241, in run 
for file_or_module in iter(tasks_queue.get, "STOP"): 
File "<string>", line 2, in get 
File "/usr/lib/python3.6/multiprocessing/managers.py", line 757, in _callmethod 
kind, result = conn.recv() 
File "/usr/lib/python3.6/multiprocessing/connection.py", line 250, in recv 
buf = self._recv_bytes() 
File "/usr/lib/python3.6/multiprocessing/connection.py", line 407, in _recv_bytes 
buf = self._recv(4) 
File "/usr/lib/python3.6/multiprocessing/connection.py", line 383, in _recv 
raise EOFError 
EOFError 
Process ChildLinter-4: 
Traceback (most recent call last): 
[...]

PR #71, opened 2020-04-03
Add a pinned version of galaxy-importer

awcrosby:
`galaxy_ng` uses `pulp_ansible` which requires `galaxy-importer` unpinned

This change uses a specific version of `galaxy-importer` for `galaxy_ng`

Supports #64

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,9 +14,9 @@ galaxy_pulp_requirements = [
 requirements = galaxy_pulp_requirements + [
     "pulpcore>=3.1.1",
     "pulp-ansible>=0.2.0b8",
-    "django-prometheus>=2.0.0"
+    "django-prometheus>=2.0.0",
+    "galaxy-importer==0.2.0",
 ]
-
 
 
 setup(


setup.py
```diff
@@ -14,11 +14,11 @@
 requirements = galaxy_pulp_requirements + [
     "pulpcore>=3.1.1",
     "pulp-ansible>=0.2.0b8",
-    "django-prometheus>=2.0.0"
+    "django-prometheus>=2.0.0",
+    "galaxy-importer==0.2.0",
```
cutwater, line 18: `galaxy-importer` is a transitive dependency for this project. It's defined as a direct dependency in `pulp_ansible` project.
Also normally you should avoid pinning to a specific version in `setup.py` unless for a reason.

awcrosby:
I see https://github.com/ansible/galaxy_ng/pull/63 pins `release_requirements.txt` which satisfies the need to specify the galaxy-importer version for a particular release. Closing this.

PR #74, opened 2020-04-07
Configure prometheus metrics for django app

cutwater:
* Enable django_prometheus middlewares.
* Add `/metrics` URL.

Issue: #10

### DIFF
--- a/galaxy_ng/app/settings.py
+++ b/galaxy_ng/app/settings.py
@@ -1,3 +1,18 @@
+MIDDLEWARE = [
+    'django_prometheus.middleware.PrometheusBeforeMiddleware',
+    # BEGIN: Pulp standard middleware
+    'django.middleware.security.SecurityMiddleware',
+    'whitenoise.middleware.WhiteNoiseMiddleware',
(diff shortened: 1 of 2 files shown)


galaxy_ng/app/settings.py
```diff
@@ -1,3 +1,18 @@
...
+    'django.contrib.sessions.middleware.SessionMiddleware',
+    'django.middleware.common.CommonMiddleware',
+    'django.middleware.csrf.CsrfViewMiddleware',
+    'django.contrib.auth.middleware.AuthenticationMiddleware',
+    'django.contrib.messages.middleware.MessageMiddleware',
+    'django.middleware.clickjacking.XFrameOptionsMiddleware',
+    # END: Pulp standard middleware
+    'django_prometheus.middleware.PrometheusAfterMiddleware',
+]
+
```
alikins, line 15: Would like to figure out if there is a way to alter the existing MIDDLEWARE instead of replacing it, but I didn't have much luck with dynaconf.

So this seems reasonable alternative.

PR #83, opened 2020-04-09
Include UI as static files during build package build.

newswangerd:
- [x] Add the compiled UI to the build process
- [x] Automatic releases to PyPI

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,6 +27,7 @@ RUN set -ex; \
         python3-devel \
         libpq \
         libpq-devel \
+        git \
     && dnf clean all \
     && rm -rf /var/cache/dnf/ \
     && rm -f /var/lib/rpm/__db.* \
--- a/setup.py
+++ b/setup.py
@@ -4,6 +4,7 @@ from setuptools import find_packages, setup
 import setuptools.command.build_py
 import tarfile
 import urllib.request
+from galaxy_ng.app.common import version
 
 # NOTE(cutwater): Because bindings are statically generated, requirements list
 #   from pulp-galaxy/setup.py has to be copied here and manually maintained.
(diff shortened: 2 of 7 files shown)


setup.py
```diff
@@ -20,6 +23,28 @@
 ]
 
 
+class BuildPyCommand(setuptools.command.build_py.build_py):
+    """Custom build command."""
+
+    def run(self):
+        print('Downloading UI static files')
+
+        download_url = 'https://github.com/ansible/ansible-hub-ui/' + \
```
cutwater, line 32: Please move `download_url` outside the function (class level \ module level) and make it a constant.

newswangerd:
Temporarily closing to prevent travis from running while I test travis releases on my personal fork.

setup.py
```diff
@@ -20,9 +24,32 @@
...
+            path='galaxy_ng/app/static/galaxy_ng'
+        )
+
+        setuptools.command.build_py.build_py.run(self)
+
+
 setup(
     name="galaxy-ng",
-    version="0.1.0a1",
+    version=version.get_git_version(),
```
cutwater, line 52: I'm not sure that using git for versioning is a good idea, though it's acceptable for now.

setup.py
```diff
@@ -1,6 +1,10 @@
 #!/usr/bin/env python3
 
 from setuptools import find_packages, setup
+import setuptools.command.build_py
+import tarfile
+import urllib.request
```
cutwater, line 6: Minor: Please organize imports according to PEP8.

Dockerfile
```diff
@@ -27,6 +27,7 @@ RUN set -ex; \
         python3-devel \
         libpq \
         libpq-devel \
+        git \
```
cutwater, line 30: According to our last discussion `git` is not going to be used from within container. Is this dependency still required?

newswangerd: Yes. The script throws an error if git isn't available that's different from the one if git is run outside of a repo, and I don't think it's a good idea to ignore that error.

newswangerd:
The travis build for this pr doesn't seem to be getting linked here correctly, but it is passing

https://travis-ci.com/github/ansible/galaxy_ng/builds/159633401

PR #84, opened 2020-04-14
Use pulp viewsets for 'v3' and add per repo/distro v3 style endpoints

alikins:
This replaces most of galaxy_ng/app/api/v3/ with viewsets based on the pulp_ansible v3 viewsets eliminating the use of galaxy_pulp for those views.

This also adds a new set of api endpoints that are exposed on urls based on each available
pulp Distribution's base_path. ie, there is an api endpoint for each pulp Repository

The new 'per Distribution' endpoints:

```
/api/automation-hub/content/<str:path>/v3/collections/	galaxy_ng.app.api.v3.viewsets.CollectionViewSet	galaxy:api:per_distribution:collection-list
/api/automation-hub/content/<str:path>/v3/collections/<str:namespace>/<str:name>/	galaxy_ng.app.api.v3.viewsets.CollectionViewSet	galaxy:api:per_distribution:collection
/api/automation-hub/content/<str:path>/v3/collections/<str:namespace>/<str:name>/versions/	galaxy_ng.app.api.v3.viewsets.CollectionVersionViewSet	galaxy:api:per_distribution:collection-version-list
/api/automation-hub/content/<str:path>/v3/collections/<str:namespace>/<str:name>/versions/<str:version>/	galaxy_ng.app.api.v3.viewsets.CollectionVersionViewSet	galaxy:api:per_distribution:collection-version
/api/automation-hub/content/<str:path>/v3/imports/collections/<str:pk>/	galaxy_ng.app.api.v3.viewsets.CollectionImportViewSet	galaxy:api:per_distribution:collection-import
/api/automation-hub/content/<str:path>/v3/artifacts/collections/	galaxy_ng.app.api.v3.viewsets.CollectionUploadViewSet	galaxy:api:per_distribution:collection-artifact-upload
/api/automation-hub/content/<str:path>/v3/	galaxy_ng.app.api.v3.viewsets.ApiRootView	galaxy:api:per_distribution:root-v3
/api/automation-hub/content/<str:path>/	galaxy_ng.app.api.v3.viewsets.ApiRootView	galaxy:api:root-per-distro
/api/automation-hub/content/<str:path>/api/	galaxy_ng.app.api.v3.viewsets.SlashApiRedirectPerDistroView	galaxy:api:api-redirect
```

The 'default distribution' endpoints are equivalent to the old 'v3/' endpoints (excluding '_ui' for now).
They both use the same viewset implementation, but the 'default distribution' sets a default
distribution base_path.

```
[...]

PR #96, opened 2020-04-20
Get download_url to point at pulp content app

alikins:
Let pulp_ansible / galaxy_ng point to the default pulp content urls for
'download_url' in CollectionVersion. This lets ansible-galaxy download
directly from the pulp content-app.

This avoids galaxy_ng having to provide a viewset that provides the 
'GET /v3​/artifacts​/collections​/{filename}' endpoint. That is the endpoint
that was previously provided by galaxy_ng's CollectionArtifactDownloadView.

Since there isn't a pulp_ansible equivalent to that viewset, there is nothing
to expose in the per-repo urls.

For the standalone cases, using the pulp content app urls provided in
'download_url' for ansible-galaxy downloads makes sense.

For hosted/c.r.c scenario, we may need a custom download endpoint, depending
on what it needs to do. (For ex, download metrics).

This is related to https://github.com/ansible/galaxy_ng/pull/84

### DIFF
--- a/dev/insights/galaxy_ng.env
+++ b/dev/insights/galaxy_ng.env
@@ -1,9 +1,7 @@
-PULP_CONTENT_PATH_PREFIX=/api/automation-hub/v3/artifacts/collections/
-
 PULP_GALAXY_API_PATH_PREFIX=/api/automation-hub/
 PULP_GALAXY_AUTHENTICATION_CLASSES=['galaxy_ng.app.auth.auth.RHIdentityAuthentication']
 PULP_GALAXY_PERMISSION_CLASSES=['rest_framework.permissions.IsAuthenticated', 'galaxy_ng.app.auth.auth.RHEntitlementRequired']
 PULP_RH_ENTITLEMENT_REQUIRED=insights
 
 PULP_ANSIBLE_API_HOSTNAME=http://localhost:5001
-PULP_ANSIBLE_CONTENT_HOSTNAME=http://localhost:24816/api/automation-hub/v3/artifacts/collections
+PULP_ANSIBLE_CONTENT_ORIGIN=http://localhost:24816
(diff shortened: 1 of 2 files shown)


dev/insights/galaxy_ng.env
```diff
@@ -1,9 +1,7 @@
-PULP_CONTENT_PATH_PREFIX=/api/automation-hub/v3/artifacts/collections/
```
cutwater, line 1: Why this parameter got removed. Removing it breaks `download_url` returned by `api/automation-hub/v3/collections/<namespace>/<name>/versions/<version>/` endpoint.

dev/insights/galaxy_ng.env
```diff
@@ -1,9 +1,7 @@
...
-
 PULP_GALAXY_API_PATH_PREFIX=/api/automation-hub/
 PULP_GALAXY_AUTHENTICATION_CLASSES=['galaxy_ng.app.auth.auth.RHIdentityAuthentication']
 PULP_GALAXY_PERMISSION_CLASSES=['rest_framework.permissions.IsAuthenticated', 'galaxy_ng.app.auth.auth.RHEntitlementRequired']
 PULP_RH_ENTITLEMENT_REQUIRED=insights
 
 PULP_ANSIBLE_API_HOSTNAME=http://localhost:5001
-PULP_ANSIBLE_CONTENT_HOSTNAME=http://localhost:24816/api/automation-hub/v3/artifacts/collections
```
cutwater, line 9: Same here. This parameter was requested by Towen team, because they were using pulp API directly.

chouseknecht:
>For the standalone cases, using the pulp content app urls provided in
'download_url' for ansible-galaxy downloads makes sense.

>For hosted/c.r.c scenario, we may need a custom download endpoint, depending
on what it needs to do. (For ex, download metrics).

I don't understand the above comments. As much as possible, local AH and cloud AH should look the same and provide the same API endpoints and collect the same metrics. How we collect metrics may be different, as Prometheus might not be available locally, but that's a different problem.

PR #99, opened 2020-04-21
Add ansible tmp dir to fix running ansible-doc in importer

awcrosby:
Error occurs in openshift env, a permission error on creating the tmp dir

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -55,16 +55,21 @@ RUN set -ex; \
 RUN set -ex; \
     mkdir -p /var/lib/pulp/artifact \
              /var/lib/pulp/tmp \
+             /tmp/ansible \
+             /etc/ansible \
     && chown -R ${USER_NAME}:${USER_GROUP} \
         /app \
         /venv \
         /var/lib/pulp \
+        /tmp/ansible \
+        /etc/ansible \
     && chmod -R 0775 /var/lib/pulp \
                      /app/docker/entrypoint.sh \
                      /app/docker/bin/* \
     && mv /app/docker/entrypoint.sh /entrypoint.sh \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -55,10 +55,12 @@ RUN set -ex; \
 RUN set -ex; \
     mkdir -p /var/lib/pulp/artifact \
              /var/lib/pulp/tmp \
+             /.ansible/tmp \
     && chown -R ${USER_NAME}:${USER_GROUP} \
         /app \
         /venv \
         /var/lib/pulp \
+        /.ansible/tmp \
```
cutwater, line 63: Can you configure it to use different directory?
I think it should be declared as a VOLUME.

awcrosby: in galaxy-dev we added `ansible.cfg` to make it use `/tmp/ansible`, we can do the same and make that a VOLUME

cutwater: Please do it.

awcrosby: I made the updates - I only added the volume to the pulp worker - let me know how these changes look

dev/docker-compose.yml
```diff
@@ -48,6 +48,7 @@ services:
       - "./common/settings.py:/etc/pulp/settings.py:z"
       - "${COMPOSE_CONTEXT}/..:/src:z"
       - "pulp_artifact:/var/lib/pulp/artifact"
+      - "ansible_tmp:/tmp/ansible"
```
cutwater, line 51: This is unnecessary. Normally you don't need persistent volume for temporary files.

awcrosby: updated

PR #101, opened 2020-04-21
Allow for running custom init scripts via volume mounted dir

shanemcd:
Ping @cutwater - here's the idea we chatted about.

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -30,6 +30,23 @@ install_local_deps() {
     done
 }
 
+process_init_files() {
+    local file
+    for file; do
+        case "$file" in
+            *.sh)
+                if [[ -x "$file" ]]; then
+                    log_message "$0: running $file"
+                    "$file"
+                else
+                    log_message "$0: sourcing $file"
+                    source "$file"
+                fi
+                ;;
+            *) log_message "$0: ignoring $file" ;;
+        esac
+    done
+}
 
 run_service() {
     if [[ "$#" -eq 0 ]]; then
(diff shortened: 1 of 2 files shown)


docker/entrypoint.sh
```diff
@@ -30,6 +30,23 @@ install_local_deps() {
     done
 }
 
+process_init_files() {
+    local f
```
cutwater, line 34: Please use more verbose variable naming.

docker/entrypoint.sh
```diff
@@ -30,6 +30,23 @@ install_local_deps() {
...
+    for f; do
+        case "$f" in
+            *.sh)
+                if [ -x "$f" ]; then
```
cutwater, line 38: We're using bash in docker scripts. It's safe and preferred to use bash `[[` operator.

docker/entrypoint.sh
```diff
@@ -30,6 +30,23 @@ install_local_deps() {
...
+                    echo "$0: running $f"
+                    "$f"
+                else
+                    echo "$0: sourcing $f"
+                    . "$f"
```
cutwater, line 43: Please use `source` alias instead of `.`

docker/entrypoint.sh
```diff
@@ -30,6 +30,23 @@ install_local_deps() {
     done
 }
 
+process_init_files() {
+    local f
+    for f; do
+        case "$f" in
+            *.sh)
+                if [ -x "$f" ]; then
+                    echo "$0: running $f"
```
cutwater, line 39: please use `log_message()` function for consistency with the rest of the script.

cutwater:
@shanemcd Thank you for submitting this PR. There are minor issues, but overall looks good.

shanemcd:
@cutwater I believe I've addressed all of your feedback.

cutwater:
@shanemcd Thank you. LGTM.

Issue #76, opened 2020-04-07
API: Create and manage users


chouseknecht:
Subtask of #75 

For local AH, add endpoints for creating and managing users. On c.rh.c AH will continue to rely on RH SSO and related services.

PR #121, opened 2020-04-27
Add user management API

newswangerd:
Adds 
- `/api/automation-hub/v3/_ui/users/` endpoint which allows users in the `partner-engineer` group to view, create and update existing using.

Updates
- `/api/automation-hub/v3/_ui/me/` endpoint to show more details about the current user as well as give them the ability to update their information.
- Moves `CurrentUser` views and serializers into `user.py`

#76


galaxy_ng/app/api/ui/urls.py
```diff
@@ -8,6 +8,7 @@
 router.register('namespaces', viewsets.NamespaceViewSet, basename='namespaces')
 router.register('my-namespaces', viewsets.MyNamespaceViewSet, basename='my-namespaces')
 router.register('collections', viewsets.CollectionViewSet, basename='collections')
+router.register('users', viewsets.UserViewSet, basename='users')
```
chouseknecht, line 11: How do we remove or disable the `users` endpoints when running on c.rh.com?

galaxy_ng/app/api/ui/urls.py
```diff
@@ -22,5 +23,5 @@
     path('', include(router.urls)),
     # NOTE: Using path instead of SimpleRouter because SimpleRouter expects retrieve
     # to look up values with an ID
-    path('me/', viewsets.CurrentUserViewSet.as_view({'get': 'retrieve'}))
+    path('me/', viewsets.CurrentUserViewSet.as_view({'get': 'retrieve', 'put': 'update'}))
```
chouseknecht, line 26: Same question here. I'm guessing we don't want the `put` method to be accessible when on c.rh.com.

galaxy_ng/app/api/ui/urls.py
newswangerd: This is a good question. On the UI we have a flag to toggle between on prem and cloud. I'm not a huge fan of this approach, but we were sort of forced into doing it because of some of the limitations with webpack. @cutwater do you have any more pythonic ideas?

https://github.com/ansible/ansible-hub-ui/blob/master/src/api/user.ts#L13

galaxy_ng/app/api/ui/serializers/user.py
```diff
@@ -0,0 +1,80 @@
+from rest_framework import serializers
+from rest_framework.exceptions import ValidationError
+from galaxy_ng.app.api import permissions
```
cutwater, line 3: Please group imports according to PEP8.

galaxy_ng/app/api/ui/serializers/user.py
```diff
@@ -0,0 +1,80 @@
...
+    class Meta:
+        model = auth_models.Group
+        fields = (
+            'id',
+            'name'
+        )
+
+
+class UserSerializer(serializers.ModelSerializer):
+    password_confirm = serializers.CharField(write_only=True, allow_blank=True, required=False)
```
cutwater, line 19: While confirmation field is useful when working with HTML forms and user interface, I don't understand why it has to be a part of the API.

galaxy_ng/app/api/ui/serializers/user.py
```diff
@@ -0,0 +1,80 @@
...
+        if len(password) <= 10:
+            raise ValidationError(detail={
+                'password': "Password must be 10+ characters long"})
+        if password != self.initial_data['password_confirm']:
+            raise ValidationError(detail={
+                'password': "Passwords do no match"})
+        return password
+
+    def update(self, instance, data):
+        if 'password' in data:
```
cutwater, line 49:
```
password = data.pop('password')
if password:
    instance.set_password(password)
```

galaxy_ng/app/api/ui/serializers/user.py
```diff
@@ -0,0 +1,80 @@
...
+        extra_kwargs = {
+            'password': {'write_only': True, 'allow_blank': True, 'required': False}
+        }
+
+    def validate_password(self, password):
+        if not password:
+            return
+        if len(password) <= 10:
+            raise ValidationError(detail={
+                'password': "Password must be 10+ characters long"})
```
cutwater, line 42: Error message looks inconsistent with actual check. Does `10+` mean range 10 and higher (including 10 itself)?
This code will fail if length of the password is exactly 10 characters.

galaxy_ng/app/api/ui/serializers/user.py
```diff
@@ -0,0 +1,80 @@
...
+            'password_confirm'
+        )
+        extra_kwargs = {
+            'password': {'write_only': True, 'allow_blank': True, 'required': False}
+        }
+
+    def validate_password(self, password):
+        if not password:
+            return
+        if len(password) <= 10:
```
cutwater, line 40: This value must not be hardcoded. Please move it to project-level constants as it can be used elsewhere.

galaxy_ng/app/api/ui/serializers/user.py
```diff
@@ -0,0 +1,80 @@
...
+            'username',
+            'first_name',
+            'last_name',
+            'email',
+            'groups',
+            'password',
+            'password_confirm'
+        )
+        extra_kwargs = {
+            'password': {'write_only': True, 'allow_blank': True, 'required': False}
```
cutwater, line 34: You could've add `min_length=` here instead of writing custom validation.

galaxy_ng/app/api/ui/serializers/user.py
```diff
@@ -0,0 +1,80 @@
...
+
+            # password doesn't get set the same as other data, so delete it
+            # before the serializer saves
+            del data['password']
+
+        return super().update(instance, data)
+
+    def to_representation(self, instance):
+        representation = super().to_representation(instance)
+        representation['groups'] = \
```
cutwater, line 61: Please use parentheses for wrapping long lines.
Also please note, that the maximum line length for this project  complies with pulp coding style and is 100 characters.

galaxy_ng/app/api/ui/viewsets/user.py
```diff
@@ -0,0 +1,41 @@
+from rest_framework import mixins
+from rest_framework.response import Response
```
cutwater, line 2: Flake8:

galaxy_ng/app/api/ui/viewsets/user.py:2:1: F401 'rest_framework.response.Response' imported but unused


It doesn't seem that flake8 is executed against PRs on Travis. Reported it as a bug #129

galaxy_ng/app/api/ui/viewsets/user.py
```diff
@@ -0,0 +1,41 @@
...
+class CurrentUserViewSet(
+    api_base.GenericViewSet,
+    mixins.RetrieveModelMixin,
+    mixins.UpdateModelMixin,
+):
+    serializer_class = serializers.CurrentUserSerializer
+    model = auth_models.User
+
+    def get_object(self):
+        if not self.request.user.is_authenticated:
```
cutwater, line 36: Is there a reason for checking authentication manually instead of using a permission class `rest_framework.permissions.IsAuthenticated`

galaxy_ng/app/api/ui/serializers/user.py
cutwater: UPD: Password validation is configured globally for Django projects via `AUTH_PASSWORD_VALIDATORS` settings. In addition to that Django provides builtin facilities to validate user password [1]
This is the preferred way which you should use to validate user passwords.


[1] https://docs.djangoproject.com/en/3.0/topics/auth/passwords/#module-django.contrib.auth.password_validation

galaxy_ng/app/api/ui/serializers/user.py
newswangerd: Oh, that's a good point. We can just validate this on the UI.

cutwater:
In addition to comments above. Can you please add unit tests as we discussed earlier. This is a good time to start adding them.

galaxy_ng/app/api/ui/serializers/user.py
newswangerd: Adding this might break the initial data since the default admin password will be invalid. I'll look into it

newswangerd: also, I believe anything we put in settings can potentially get overridden by other pulp plugins

cutwater: My point is that we should rely on a password policy configured at the project level, because we use `django.auth`.

galaxy_ng/app/api/ui/serializers/user.py
cutwater: `if 'password' in data` is not needed. Forgot that `dict.pop` raises `KeyError` if called without a second argument`, should be:
```
password = data.pop('password', None)
```

galaxy_ng/app/api/ui/urls.py
cutwater: Should this router be disabled for insights mode?

galaxy_ng/app/api/ui/viewsets/user.py
```diff
@@ -0,0 +1,43 @@
...
+    mixins.UpdateModelMixin,
+):
+    serializer_class = serializers.CurrentUserSerializer
+    model = auth_models.User
+    permission_classes = [
+        drf_permissions.IsAuthenticated,
+        permissions.RestrictUnsafeOnCloudDeployments]
+
+    def get_object(self):
+        obj, created = self.model.objects.get_or_create(
```
cutwater, line 40: Why it's `get_or_create`?

galaxy_ng/app/api/ui/viewsets/user.py
```diff
@@ -0,0 +1,43 @@
...
+
+class CurrentUserViewSet(
+    api_base.GenericViewSet,
+    mixins.RetrieveModelMixin,
+    mixins.UpdateModelMixin,
+):
+    serializer_class = serializers.CurrentUserSerializer
+    model = auth_models.User
+    permission_classes = [
+        drf_permissions.IsAuthenticated,
```
cutwater, line 36: This will override default permission classes set in project settings. This is probably not that is wanted.
Better override `get_permissions()` method and append permissions to the default ones.

galaxy_ng/app/api/ui/urls.py
newswangerd: I don't think that's necessary. People shouldn't be using the `_ui` api anyway.

galaxy_ng/app/api/ui/viewsets/user.py
newswangerd: Oh. Good catch. I based the view off of https://github.com/ansible/galaxy/blob/devel/galaxy/api/views/active_user.py#L26 and missed that it was using `get_or_create`

galaxy_ng/app/api/ui/viewsets/user.py
```diff
@@ -0,0 +1,49 @@
...
+
+class UserViewSet(
+    mixins.ListModelMixin,
+    mixins.RetrieveModelMixin,
+    mixins.CreateModelMixin,
+    mixins.UpdateModelMixin,
+    api_base.GenericViewSet,
+):
+
+    def get_serializer_class(self):
```
cutwater, line 20: minor: Because it just returns static class, it can be defined as `serializer_class = ` class attribute.

galaxy_ng/app/api/ui/viewsets/user.py
```diff
@@ -0,0 +1,49 @@
...
+        return serializers.UserSerializer
+
+    def get_queryset(self):
+        return auth_models.User.objects.all()
```
cutwater, line 24: minor: ditto.

galaxy_ng/app/api/ui/viewsets/user.py
```diff
@@ -0,0 +1,49 @@
...
+from galaxy_ng.app.api import base as api_base
+
+
+class UserViewSet(
+    mixins.ListModelMixin,
+    mixins.RetrieveModelMixin,
+    mixins.CreateModelMixin,
+    mixins.UpdateModelMixin,
+    api_base.GenericViewSet,
+):
```
cutwater, line 18:
```
model = auth_models.User
serializer_class = serializers.UserSerializer
```

galaxy_ng/app/api/ui/viewsets/user.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def get_serializer_class(self):
+        return serializers.UserSerializer
+
+    def get_queryset(self):
+        return auth_models.User.objects.all()
+
+    def get_permissions(self):
+        permission_list = super().get_permissions()
+        permission_list.append(permissions.IsPartnerEngineer())
```
cutwater, line 28: IMPORTNANT: You **must not** modify instance returned by `get_permissions()` method, as it will modify global setting parameter returned by reference from `get_permissions()` method.
```
return super().get_permissions() + [
   permissions.IsPartnerEngineer(),
   permissionsRestrictOnCloudDeployments(),
]
```

galaxy_ng/app/api/ui/viewsets/user.py
```diff
@@ -0,0 +1,49 @@
...
+
+class CurrentUserViewSet(
+    api_base.GenericViewSet,
+    mixins.RetrieveModelMixin,
+    mixins.UpdateModelMixin,
+):
+    serializer_class = serializers.CurrentUserSerializer
+    model = auth_models.User
+
+    def get_permissions(self):
```
cutwater, line 41: ditto

galaxy_ng/app/api/ui/viewsets/user.py
```diff
@@ -0,0 +1,49 @@
...
+        permission_list = super().get_permissions()
+        permission_list.append(drf_permissions.IsAuthenticated())
+        permission_list.append(permissions.RestrictUnsafeOnCloudDeployments())
+        return permission_list
+
+    def get_object(self):
+        obj = get_object_or_404(self.model, pk=self.request.user.pk)
```
cutwater, line 48: minor: `return get_obj_or_404(...)`

galaxy_ng/app/api/ui/viewsets/user.py
newswangerd: In that case we need to update the namespace viewsets too

https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/api/ui/viewsets/namespace.py#L45

cutwater: UPD: DRF returns new list object in `get_permissions()` method. So technically it's valid.
However generally it's considered a bad practice to modify objects with lifetime you don't control.

PR #124, opened 2020-04-28
Updating README with user install

fao89:
### Blocked by: 
https://github.com/ansible/galaxy_ng/pull/125


bmbouter:
This looks great! Thank you @fao89!

My only request is to add a warning block somewhere near the playbook indicating users should set the `pulp_default_admin_password` field. Otherwise lgtm!

PR #128, opened 2020-04-29
Update release requirements

cutwater:


alikins:
@cutwater Looks like the commit needs to be signed.

cutwater:
Looks like git was using wrong key. Signed with appropriate key now.

PR #135, opened 2020-04-30
Rename flake8.cfg to .flake8 (default location)

alikins:
Per https://flake8.pycqa.org/en/latest/user/configuration.html#configuration-locations
flake8 will look in per project setup.cfg, tox.ini, or .flake8
for configuration.

Invoking plain 'flake8' would fail to find or use flake8.cfg,
so rename it to .flake8


.travis/before_install.sh
```diff
@@ -49,7 +49,7 @@ pip install -r dev_requirements.txt
 
 
 # Lint code.
-#flake8 --config flake8.cfg
+#flake8 --config .flake8
```
cutwater, line 52: Looks like `--config .flake8` is redundant here, isn't it?

Issue #117, opened 2020-04-24
Importer: Custom galaxy_ng code on import, promotion to staging


awcrosby:
Subtask of #20

Work with pulp-ansible to trigger promotion from inbound repo to staging

Issue #41, opened 2020-03-19
API: Content promotion move endpoint


chouseknecht:
Subtask of #20 

-  [x] Migrate POC promotion task code from `pulp_ansible` to `galaxy_ng`
https://github.com/pulp/pulp_ansible/blob/master/pulp_ansible/tests/performance/promote.py
-  [x] Enable RH partner engineers to promote artifacts from the 'unreviewed' repository to the 'rejected' or 'certified' repositories. 
- [x] Enable RH partner engineers to promote artifacts from the 'rejected' repository to the 'certified' repository
- [x] When an artifact is 'promoted', it will be copied to the target repository and removed from the source repository. For example, when promoting from 'uncertified' to 'certified', the artifact is copied to 'certified' and removed from 'uncertified'.

**Notes:**
- Repository names should not be hardcoded. See #39

PR #140, opened 2020-05-01
Promotion collection version move/ endpoint

awcrosby:
`move/` endpoint on a CollectionVersion that starts 2 pulp tasks... 1 to make a new RepositoryVerison of source repo without CollectionVersion, and 1 to make a new RepositoryVersion of destination repo with CollectionVersion. Works #41, #117

Example:
`api/automation-hub/content/automation-hub/v3/collections/<str:namespace>/<str:name>/versions/<str:version>/move/<str:source_path>/<str:dest_path>/`

Testing:
1. Confirm a version (i.e. `1.0.74`) in list http://localhost:5001/api/automation-hub/content/staging/v3/collections/awcrosby/collection_test/versions/
2. POST to http://localhost:5001/api/automation-hub/v3/collections/awcrosby/collection_test/versions/1.0.74/move/staging/automation-hub/ and see task ids returned
3. Confirm version no longer in source_repo list: http://localhost:5001/api/automation-hub/content/staging/v3/collections/awcrosby/collection_test/versions/
4. Confirm version now in destination_repo list: http://localhost:5001/api/automation-hub/content/automation-hub/v3/collections/awcrosby/collection_test/versions/

### DIFF
--- a/galaxy_ng/app/api/v3/urls.py
+++ b/galaxy_ng/app/api/v3/urls.py
@@ -53,4 +53,10 @@ urlpatterns = [
         viewsets.CollectionArtifactDownloadView.as_view(),
         name='collection-artifact-download'
     ),
+    path(
+        'collections/<str:namespace>/<str:name>/versions/<str:version>/move/'
+        '<str:source_path>/<str:dest_path>/',
+        viewsets.CollectionVersionMoveViewSet.as_view({'post': 'move_content'}),
+        name='collection-version-move',
+    ),
 ]
--- a/galaxy_ng/app/api/v3/viewsets/collection.py
+++ b/galaxy_ng/app/api/v3/viewsets/collection.py
@@ -2,7 +2,7 @@ import logging
 
 import requests
 
-from django.core.exceptions import ValidationError
+from django.core.exceptions import ValidationError, ObjectDoesNotExist
 
 
 from django.conf import settings
@@ -16,10 +16,12 @@ from pulpcore.plugin.models import ContentArtifact, Task
 from pulpcore.plugin.tasking import enqueue_with_reservation
 from pulp_ansible.app.tasks.collections import import_collection
 from pulp_ansible.app.galaxy.v3 import views as pulp_ansible_views
+from pulp_ansible.app.models import CollectionVersion, AnsibleDistribution
 
 from galaxy_ng.app.api.base import (
     GALAXY_PERMISSION_CLASSES,
     APIView,
+    ViewSet,
     LocalSettingsMixin,
 )
 
@@ -29,7 +31,11 @@ from galaxy_ng.app.api import permissions
 from galaxy_ng.app.api.v3.serializers import CollectionVersionSerializer, CollectionUploadSerializer
 
 from galaxy_ng.app.common import metrics
-from galaxy_ng.app.tasks import import_and_auto_approve
+from galaxy_ng.app.tasks import (
+    import_and_auto_approve,
+    add_content_to_repository,
+    remove_content_from_repository,
+)
 
 # hmm, not sure what to do with this
 # from galaxy_ng.app.common.parsers import AnsibleGalaxy29MultiPartParser
@@ -180,3 +186,65 @@ class CollectionArtifactDownloadView(APIView):
         metrics.collection_artifact_download_failures.labels(status=response.status_code).inc()
         raise APIException('Unexpected response from content app. '
                            f'Code: {response.status_code}.')
+
+
+class CollectionVersionMoveViewSet(ViewSet):
+    permission_classes = GALAXY_PERMISSION_CLASSES + [
+        permissions.IsNamespaceOwner
+    ]
+
+    def move_content(self, request, *args, **kwargs):
+        """Remove content from source repo and add to destination repo.
+
+        Creates new RepositoryVersion of source repo without content included.
+        Creates new RepositoryVersion of destination repo with content included.
+        """
+
+        version_str = '-'.join([self.kwargs[key] for key in ('namespace', 'name', 'version')])
+        try:
+            collection_version = CollectionVersion.objects.get(
+                namespace=self.kwargs['namespace'],
+                name=self.kwargs['name'],
+                version=self.kwargs['version'],
+            )
+        except ObjectDoesNotExist:
+            raise NotFound(f'Collection {version_str} not found')
+
+        try:
+            src_repo = AnsibleDistribution.objects.get(
+                base_path=self.kwargs['source_path']).repository
+            dest_repo = AnsibleDistribution.objects.get(
+                base_path=self.kwargs['dest_path']).repository
+        except ObjectDoesNotExist:
+            raise NotFound(f'Repo(s) for moving collection {version_str} not found')
+
+        src_versions = CollectionVersion.objects.filter(pk__in=src_repo.latest_version().content)
+        if collection_version not in src_versions:
+            raise NotFound(f'Collection {version_str} not found in source repo')
+
+        dest_versions = CollectionVersion.objects.filter(pk__in=dest_repo.latest_version().content)
+        if collection_version in dest_versions:
+            raise APIException(f'Collection {version_str} already found in destination repo')
+
+        add_task = self._add_content(collection_version, dest_repo)
+        remove_task = self._remove_content(collection_version, src_repo)
+
+        return Response(
+            data={
+                'add_task_id': add_task.id,
+                'remove_task_id': remove_task.id,
+            },
+            status='202'
+        )
+
+    @staticmethod
+    def _add_content(collection_version, repo):
+        locks = [repo]
+        task_args = (collection_version.pk, repo.pk)
+        return enqueue_with_reservation(add_content_to_repository, locks, args=task_args)
+
+    @staticmethod
+    def _remove_content(collection_version, repo):
+        locks = [repo]
+        task_args = (collection_version.pk, repo.pk)
+        return enqueue_with_reservation(remove_content_from_repository, locks, args=task_args)
--- a/openapi/openapi.yaml
+++ b/openapi/openapi.yaml
@@ -191,6 +191,28 @@ paths:
         'default':
           $ref: '#/components/responses/Errors'
 
+  '/v3/collections/{namespace}/{name}/versions/{version}/move/{source_path}/{destination_path}':
+    parameters: #tag
+      - $ref: '#/components/parameters/CollectionNamespaceName'
+      - $ref: '#/components/parameters/CollectionName'
+      - $ref: '#/components/parameters/SemanticVersion'
+      - $ref: '#/components/parameters/DistributionBasePath'
+      - $ref: '#/components/parameters/DistributionBasePath'
+    post:
+      summary: >
+        Queues tasks to move CollectionVersion
+        from source Repository to destination Repository
+        using the Repository DistributionBasePath.
+      operationId: moveCollectionVersion
+      tags:
+        - Collections
+      parameters: []
+      responses:
+        '202':
+          $ref: '#/components/responses/CollectionVersionMove'
+        'default':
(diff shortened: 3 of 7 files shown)


galaxy_ng/app/api/v3/viewsets/collection.py
```diff
@@ -180,3 +186,64 @@ def get(self, request, *args, **kwargs):
...
+        if collection_version in dest_versions:
+            raise APIException('Collection version already found in destination repo')
+
+        add_task = self._add_content(collection_version, dest_repo)
+        remove_task = self._remove_content(collection_version, src_repo)
+
+        return Response(
+            data={
+                'add_task_id': add_task.id,
+                'remove_task_id': remove_task.id,
```
alikins, line 234: Where would these task_id's be looked up?  Is there the equilivent of the the /imports/ endpoints or would client tooling need to use the pulp api task endpoints to check status?

galaxy_ng/app/api/v3/viewsets/collection.py
```diff
@@ -180,3 +186,64 @@ def get(self, request, *args, **kwargs):
...
+    def _add_content(collection_version, repo):
+        locks = [repo]
+        task_args = (collection_version.pk, repo.pk)
+        return enqueue_with_reservation(add_content_to_repository, locks, args=task_args)
+
+    @staticmethod
+    def _remove_content(collection_version, repo):
+        locks = [repo]
+        task_args = (collection_version.pk, repo.pk)
+        return enqueue_with_reservation(remove_content_from_repository, locks, args=task_args)
```
alikins, line 249: Do these tasks change the RepositoryVersion 'atomically'?  Will a /move/ bump the RepositoryVersion twice (once for add, once for remove) ?

openapi/openapi.yaml
```diff
@@ -191,6 +191,28 @@ paths:
...
+      summary: >
+        Queues tasks to move CollectionVersion
+        from source Repository to destination Repository
+        using the Repository DistributionBasePath.
+      operationId: moveCollectionVersion
+      tags:
+        - Collections
+      parameters: []
+      responses:
+        '200':
```
alikins, line 211: Posting to this endpoint creates async tasks that (possibly) have not finished at the end of the request?

Could be more appropriate to return 202 for that case  (https://geemus.gitbooks.io/http-api-design/content/en/responses/return-appropriate-status-codes.html)

galaxy_ng/app/api/v3/urls.py
```diff
@@ -53,4 +53,10 @@
         viewsets.CollectionArtifactDownloadView.as_view(),
         name='collection-artifact-download'
     ),
+    path(
+        'collections/<str:namespace>/<str:name>/versions/<str:version>/move/'
+        '<str:source_path>/<str:dest_path>/',
```
alikins, line 58: Wondering if there is any utility to making it explicit that this is an 'action'. 

https://geemus.gitbooks.io/http-api-design/content/en/requests/actions.html suggests that. I can see the reasoning behind the suggestion and generally like the rest of the guidelines/suggestions in that doc. 

But it also seems like it doesn't matter that much.

alikins: I'm less sure about the /move/{src}/{dest}  path args. On one hand, it's straightforward and elegant. On the other hand, /move/{src}/  kind of prevents there ever being /move/ sub actions  (/move/recursive/src/dest maybe?).  Probably not a big deal, since there could always be another 'action'  (ie, /move_recursive/{src}/{dest}/).

I guess one concern would be if the 'src' and 'dest' would ever need to be 'pulp_href's which could be complicated
(ie, /move/pulp/api/v3/collections/ansible/ansible/XXXXXX/pulp/api/v3/collections/ansible/ansible/YYYY).

galaxy_ng/app/api/v3/viewsets/collection.py
```diff
@@ -180,3 +186,64 @@ def get(self, request, *args, **kwargs):
...
+
+    def move_content(self, request, *args, **kwargs):
+        """Remove content from source repo and add to destination repo.
+
+        Creates new RepositoryVersion of source repo without content included.
+        Creates new RepositoryVersion of destination repo with content included.
+        """
+
+        try:
+            collection_version = CollectionVersion.objects.get(
```
alikins, line 204: At the moment, I'm not sure if a CollectionVersion can exist in multiple repositories, but seems like that would eventually be the case. Does this need to filter on the Repository as well? (To avoid multiple matches)

galaxy_ng/app/api/v3/viewsets/collection.py
awcrosby: i think it would need to use a pulp api task endpoint. client should not need to check status for the same reasons as an import which could take a long time - this operation should be quick unless there is a system issue like locked repo or all workers busy.

galaxy_ng/app/api/v3/viewsets/collection.py
awcrosby: source_repo gets a new RepositoryVersion, and destination_repo gets a new RepositoryVersion, but this shouldn't be happening on same Repository object.

openapi/openapi.yaml
awcrosby: yeah a 202 makes sense

galaxy_ng/app/api/v3/viewsets/collection.py
awcrosby: I think eventually a CollectionVersion could exist in more than one Repository, but I don't think this query needs a Repository filter.... this is simply looking up the CollectionVersion whether it exists in a particular Repository or not. Below it is the check to see if it actually exists in the source_repo.

galaxy_ng/app/api/v3/viewsets/collection.py
alikins: I don't know of a particular use case for the task_ids currently. I suspect the first thing that would want/need the task_ids and task status would be web ui stuff just as progress indicators.

Mainly just want to verify that data is returned in a useful format which for this case would just mean the task_ids can be used to find the task status if needed. Particularly if the tasks fail for some reason.

galaxy_ng/app/api/v3/viewsets/collection.py
alikins: Ah, that makes sense. 

Was going to ask about /copy/, but we discussed it in the pulp meeting. 

(Adding some links to copy impls, mostly just for my reference)

pulp_docker
https://github.com/pulp/pulp_docker/blob/master/pulp_docker/app/tasks/recursive_add.py
https://github.com/pulp/pulp_docker/blob/master/pulp_docker/app/viewsets.py#L317
pulp_rpm
https://github.com/pulp/pulp_rpm/blob/master/pulp_rpm/app/viewsets.py#L250
https://github.com/pulp/pulp_rpm/blob/master/pulp_rpm/app/tasks/copy.py

galaxy_ng/app/api/v3/viewsets/collection.py
alikins: Ah, cool.

galaxy_ng/app/api/v3/urls.py
alikins: Looking at pulp_rpm's 'copy' endpoint (https://pulp-rpm.readthedocs.io/en/latest/restapi.html#tag/rpm:-copy)  and it is just `/pulp/api/v3/rpm/copy/`.

So, /move/ seems okay.

But then again, some pulp plugin API's like
https://petstore.swagger.io/?url=https://pulp-python.readthedocs.io/en/latest/_static/api.json#/repositories%3A%20python%20versions/repositories_python_python_versions_repair

If my translation from pulp-speak is correct, is something like `POST /pulp/api/v3/repositories/python/python/aaeab0e1-baf1-482c-b083-fba26f122043`

The other pulp plugin endpoint seem to use the body for the data.  May be more consistent with
pulp to make the src/dest body params.

galaxy_ng/app/api/v3/viewsets/collection.py
awcrosby: yeah these are the same task id formats as our other pulp tasks

galaxy_ng/app/api/v3/viewsets/collection.py
```diff
@@ -180,3 +186,64 @@ def get(self, request, *args, **kwargs):
         metrics.collection_artifact_download_failures.labels(status=response.status_code).inc()
         raise APIException('Unexpected response from content app. '
                            f'Code: {response.status_code}.')
+
+
+class CollectionVersionMoveViewSet(ViewSet):
```
alikins, line 191: Wondering if this needs to be a separate viewset, of if it could be considered custom "actions" for CollectionVersionViewSet ala https://www.django-rest-framework.org/api-guide/viewsets/#marking-extra-actions-for-routing

I don't have a strong feeling about it either way, so was wondering if custom actions were considered.

galaxy_ng/app/api/v3/viewsets/collection.py
```diff
@@ -180,3 +186,64 @@ def get(self, request, *args, **kwargs):
...
+        """
+
+        try:
+            collection_version = CollectionVersion.objects.get(
+                namespace=self.kwargs['namespace'],
+                name=self.kwargs['name'],
+                version=self.kwargs['version'],
+            )
+        except ObjectDoesNotExist:
+            raise NotFound('Collection version not found')
```
alikins, line 210: Do user facing errors or log messages include info about which collection version was not found?

If not, that info should be added to the messages.

galaxy_ng/app/api/v3/viewsets/collection.py
```diff
@@ -180,3 +186,64 @@ def get(self, request, *args, **kwargs):
...
+
+        try:
+            src_repo = AnsibleDistribution.objects.get(
+                base_path=self.kwargs['source_path']).repository
+            dest_repo = AnsibleDistribution.objects.get(
+                base_path=self.kwargs['dest_path']).repository
+        except ObjectDoesNotExist:
+            raise NotFound('Repo(s) for moving collection not found')
```
alikins, line 218: Same, make sure user errors or log messages indicate which collection caused the error

galaxy_ng/app/api/v3/viewsets/collection.py
```diff
@@ -180,3 +186,64 @@ def get(self, request, *args, **kwargs):
...
+
+        src_versions = CollectionVersion.objects.filter(pk__in=src_repo.latest_version().content)
+        if collection_version not in src_versions:
+            raise NotFound('Collection version not found in source repo')
```
alikins, line 222: ditto above

galaxy_ng/app/api/v3/viewsets/collection.py
```diff
@@ -180,3 +186,64 @@ def get(self, request, *args, **kwargs):
...
+
+        dest_versions = CollectionVersion.objects.filter(pk__in=dest_repo.latest_version().content)
+        if collection_version in dest_versions:
+            raise APIException('Collection version already found in destination repo')
```
alikins, line 226: ditto

galaxy_ng/app/api/v3/viewsets/collection.py
awcrosby: Ah, just saw this comment. The message now has the collection version added to it.

galaxy_ng/app/api/v3/viewsets/collection.py
awcrosby: incorporated

galaxy_ng/app/api/v3/viewsets/collection.py
awcrosby: incorporated

galaxy_ng/app/api/v3/viewsets/collection.py
awcrosby: incorporated

Issue #45, opened 2020-03-23
API: Update OpenAPI spec with content sync blacklist endpoints


chouseknecht:
Subtask of #17 

Design and document in the OpenAPI spec any endpoints needed to support content sync blacklist creation and management.

chouseknecht:
Blocked by #29. Need to know if we're extending v3 and creating v4.

Issue #17, opened 2020-03-17
Story: Content Sync


chouseknecht:
### Use Case

As an AH administrator: 
- From the universe of subscriber content, I want to construct a list of namespaces or namespace.collection(s) to exclude or hide from my private AH.
- On my private AH, I want to use the UI to configure the sync remote, API key, and any other parameters. 
- On my private AH, I want to initiate the sync from the UI.
- In the future, I may want to create multiple blacklists. For example, I might one per private AH instance. Or, perhaps I have multiple repositories in my one private AH and I want to control the content sync’d into each.

### Notes

- The list of what not to download to private AH (i.e., blacklist) will be created upstream, server side
- Build the feature for c.rh.com first. Defer the feature on community until the future of community is decided

### Design

- As an admin, log into AH and create a blacklist of collections. User will blacklist an entire collection, not specific versions. TBD what the actual user experience is and how it manifests in the UI.
- On the backend, create a repository containing all available collection versions minus any collections on the user’s blacklist, and create a distribution that contains the repository. The local AH synch will mirror this repository using the distribution URL.
- As an admin on private AH, manually configure the remote endpoint and credentials.
- As an admin on private AH, initiates a sync.

#### Workflow to support user selecting collections:

- User chooses to add a Collection to the blacklist.
- UI calls a /blacklist endpoint that accepts a Collection ID or a list of Collection IDs. The underlying viewset will offer list, post, and put operations to manipulate a Blacklist model.
- The API will:
    - Update new data model to track user’s request to exclude the collection, which is a many-to-many between user and collection. Think of collection as a ‘stream of versions’, thus we’re tracking the request at the collection level vs the collection version level.
    - Does the user already have a sync repository? If not, create one.
    - If creating a repository:
        - Create a repository (https://pulp-ansible.readthedocs.io/en/latest/restapi.html#operation/repositories_ansible_ansible_list)
       - Associate the repository with the user
       - Create a distribution (https://pulp-ansible.readthedocs.io/en/latest/restapi.html#operation/distributions_ansible_ansible_list)
       - Associate the distribution URL with the user. Need to be able to surface this URL in the UI, making it available for copy-paste to private AH.
      - Add collection versions to the repository using the Pulp modify API. Ideally we need a ‘Copy’ API in pulp_ansible. (Brian will create an issue and deliver this.)

Tasks
- [x] Model/design for upstream / automation-hub
[...]

PR #142, opened 2020-05-04
Add openapi defs for /_ui/synclists (#45)

alikins:
Issue: #45

### DIFF
--- a/openapi/ui_openapi.yaml
+++ b/openapi/ui_openapi.yaml
@@ -384,6 +384,115 @@ paths:
         'default':
           $ref: '#/components/responses/Errors'
 
+  #
+  # Sync lists
+  #
+
+  '/_ui/synclists/':
+    get:
+      summary: List SyncLists
+      operationId: listSyncLists
+      parameters:
+        - $ref: '#/components/parameters/PageOffset'
+        - $ref: '#/components/parameters/PageLimit'
+        # If we use pulp base viewsets, could potentially get include/exclude fields
+        # - $ref: '#/components/parameters/ResponseIncludeFields'
+        # - $ref: '#/components/parameters/ResponseExcludeFields'
+      tags:
+        - SyncLists
+      responses:
+        '200':
+          $ref: '#/components/responses/SyncListList'
+        'default':
+          $ref: '#/components/responses/Errors'
+
+    post:
+      summary: Create a SyncList
+      operationId: createSyncList
+      tags:
+        - SyncLists
+      requestBody:
+        $ref: "#/components/requestBodies/SyncList"
+      responses:
+        "201":
+          description: "The SyncList was successfully created"
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/SyncList"
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '409':
+          $ref: '#/components/responses/Conflict'
+        'default':
+          $ref: '#/components/responses/Errors'
+
+  '/_ui/synclists/{synclist_id}/':
+    parameters:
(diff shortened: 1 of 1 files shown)


openapi/ui_openapi.yaml
```diff
@@ -835,6 +944,125 @@ components:
...
+          example: 'sync_example_com_namespace'
+        policy:
+          $ref: '#/components/schemas/SyncListPolicy'
+        repository:
+          # TODO: may become a repository name or a RepositorySummary
+          title: Repository ID
+          description: 'The ID of the Repository to sync to'
+          type: string
+          format: uuid
+        collections:
```
cutwater, line 973: `collections` and `namespaces` list can be eventually large. I think it's better to return and modify them via separate API requests. (e.g. `(POST|DELETE) /api/v1/synclists/{synclist_id}/(collections|namespaces)`)
For example adding \ deleting single namespace to a list should not cause sending back and forth this whole large object.

newswangerd: The UI is going to have to load all of the collections in a sync list anyway, so I'm not sure we're saving ourselves any trouble by moving this to a separate API.

newswangerd: I would like to add a new parameter on the namespace and collection list APIs that allows for the UI to query collections/namespaces by synclist id

openapi/ui_openapi.yaml
```diff
@@ -835,6 +944,125 @@ components:
...
+          title: 'Groups'
+          description: 'Array of Groups associated with this synclist'
+          type: array
+          items:
+            # add Group/User schema
+            type: string
+          uniqueItems: true
+          example:
+            - 'example-identity-account:12345'
+        users:
```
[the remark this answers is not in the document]
alikins: Based on https://github.com/ansible/galaxy_ng/issues/17 and https://github.com/ansible/galaxy_ng/issues/17#issuecomment-599957146

> Update new data model to track user’s request to exclude the collection, which is a many-to-many between user and collection.

openapi/ui_openapi.yaml
alikins: @newswangerd That sounds reasonable at first glance. I'll consider that a different issue/pr though since it is change to namespace viewsets and collection viewsets (which may potentially need tweaks in pulpcore / pulp_ansible)

newswangerd: This might be a good opportunity to look into migrating the ui collection viewset away from proxying to pulp.

openapi/ui_openapi.yaml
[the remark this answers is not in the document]
alikins: @newswangerd But to be honest, I'm not sure why. At the moment, nothing really uses the names or groups, so it's kind of flexible I guess.

chouseknecht:
> The UI is designed such that it expects there to be only one synclist per Red Hat account (group). If there are multiple sync lists, how do we determine which one to display on the Collections page on the UI?

For this first iteration... On c.rh.c the backend should only ever create one synclist per organization. For on-premise AH, there should be no synclist creation, and these APIs should be disabled.

alikins:
@chouseknecht 

> For this first iteration... On c.rh.c the backend should only ever create one synclist per organization. For on-premise AH, there should be no synclist creation, and these APIs should be disabled.

That sounds like business logic to me that should be reflected in the generic API interface, but up to the API implementations to enforce. Doesn't seem like it should impact the proposed API much. One case where it might is adding some additional error responses for cases where the business logic prevents creation of a synclist.

alikins:
@cutwater @newswangerd rebased

PR #149, opened 2020-05-06
Moving CI from pulp-operator to pulp single container

fao89:
Use pulp single container instead of pulp-operator


.travis/Containerfile.ci_base
```diff
@@ -0,0 +1,82 @@
+FROM fedora:31
```
cutwater, line 1: I wonder why not pre-build a container and upload it to quay, to avoid building it each time from `fedora` image?

.travis/filter/repr.py
```diff
@@ -0,0 +1,25 @@
+from __future__ import absolute_import, division, print_function
```
cutwater, line 1: Is this ansible plugin executed by python 2?
Otherwise these future imports are not needed.

.travis/filter/repr.py
```diff
@@ -0,0 +1,25 @@
...
+
+__metaclass__ = type
```
cutwater, line 3: As well as this hack for new-style classes popular in Ansible ecosystem.

.travis/Containerfile.ci_base
mikedep333: That's on our to-do list.

mdellweg: I broke it out into this extra Containerfile to support just this. I wanted to postpone it until we have proof from a few plugins that we do not need to change it every day.
There are few more things to consider though:
- Where do we want to host this Containerfile?
- Can plugins (or branches) access older/alternative versions of that image?
- How can we incorporate testing against python 3.6?

.travis/filter/repr.py
mdellweg: This is common ansible boilerplate. And you never know if someone is trying to some of this with python2.
(You should be able to use the playbooks provided here in a local vm to debug those containers for example.)

Issue #77, opened 2020-04-07
API: Add authentication API for local Automation Hub


chouseknecht:
Subtask of #75 

When AH is running locally it needs to provide an authentication API. When on c.rh.c this API should be disabled. 

The authentication API needs to provide the following:

- For the UI it needs to support username/password authentication
- For the CLI it needs to offer a replacement to the OpenID endpoint that RH SSO provides

chouseknecht:
Our plan is to surface the Pulp login/logout endpoints through Galaxy.

PR #150, opened 2020-05-07
Implement authentication API endpoints for local AH

cutwater:
* Update OpenAPI specification.
  - Describe login, logout and token views.
* Implement new endpoints:
  - Login endpoint.
  - Logout endpoint.
  - Token get, create and revoke endpoints.
* Add unit tests.

Issue: #77


galaxy_ng/app/api/ui/views/auth.py
```diff
@@ -0,0 +1,69 @@
...
+class LogoutView(api_base.APIView):
+    permission_classes = (RestrictOnCloudDeployments,)
+
+    def post(self, request, *args, **kwargs):
+        django_auth.logout(request)
+        return Response(status=http_code.HTTP_204_NO_CONTENT)
+
+
+class TokenView(api_base.APIView):
+    permission_classes = (IsAuthenticated, RestrictOnCloudDeployments)
```
awcrosby, line 50: Since user needs to be authenticated to create a token, how is first token generated - after session auth… or is that out of scope of this pr?

galaxy_ng/app/settings.py
```diff
@@ -1,3 +1,16 @@
+import os
+
+DATABASES = {
```
awcrosby, line 3: Curious, is this needed specifically for auth?

galaxy_ng/app/api/ui/views/auth.py
cutwater: User has to be authenticated with Session auth (login view) to issue a token

galaxy_ng/app/settings.py
cutwater: Actually no, I will move it to a separate PR.
I'm trying to reduce number of custom configurations living in different places and put everything to `galaxy_ng/app/settings` to minimize custom settings size for different environments.

galaxy_ng/app/api/ui/views/auth.py
```diff
@@ -0,0 +1,69 @@
...
+class LoginView(api_base.GenericAPIView):
+    serializer_class = LoginSerializer
+    permission_classes = (RestrictOnCloudDeployments,)
+
+    def post(self, request, *args, **kwargs):
+        serializer = self.get_serializer(data=self.request.data)
+        serializer.is_valid(raise_exception=True)
+
+        username = serializer.validated_data['username']
+        password = serializer.validated_data['password']
```
newswangerd, line 31: This isn't a requirement, but it would be good to have a `remember me` so that users can choose to not remain logged in.

galaxy_ng/app/settings.py
```diff
@@ -28,13 +33,15 @@
...
-    "rest_framework.authentication.BasicAuthentication",
-    # "galaxy_ng.app.auth.auth.RHIdentityAuthentication",
+    "rest_framework.authentication.TokenAuthentication",
 ]
 GALAXY_PERMISSION_CLASSES = [
     'rest_framework.permissions.IsAuthenticated',
-    # 'galaxy_ng.app.auth.auth.RHEntitlementRequired',
 ]
+# Settings for insights mode
+# GALAXY_AUTHENTICATION_CLASSES = ["galaxy_ng.app.auth.auth.RHIdentityAuthentication"]
```
bmbouter, line 42: dynaconf does provide some capabilities to have different settings for different environments. Just an FYI https://dynaconf.readthedocs.io/en/latest/guides/usage.html#settings-file-formats

Issue #167, opened 2020-05-14
Collection POST responds with invalid task URL for status checks


ironfroggy:
## Bug Report
Affecting QA and CI. When I post a collection, the response that comes back with a task URL to poll for the import status provides the wrong URL. It looks like the URL we may expose for stand-alone mode, not the Insights API routed URLs we need to use here.

POST collection to here:
`https://ci.cloud.redhat.com/api/automation-hub/v3/artifacts/collections/`

Get a task URL here:
`https://ci.cloud.redhat.com/pulp_ansible/galaxy/automation-hub/api/v3/imports/collections/28e37da1-3aff-4a0b-9bca-c48b7c133215/`

alikins:
I think I see the problem (should probably return the get_absolute_url() instead).

it's returning the pulp_ansible task url instead of the collectionImport url
￼
Interesting... galaxy_ng CollectionImport model doesn't have a ref to a Repository or a Distribution to populate the <path> in it's get_absolute_url(). May need to add a field to CollectionImport for that. Which makes sense if CollectionImport is the record of the import which are "to" a particular Repository.

PR #169, opened 2020-05-14
Fix import record url returned from publish

alikins:
First pass at returning the CollectionImport url
include the repo <path> in the response to
a POST to:
        /api/automation-hub/content/automation-hub/v3/artifacts/collections/

Better fix will be either add method to CollectionImport to build
url with pass in <path>.

Or Better, to add a field to CollectionImport that references the
Repository the collection was imported to, so it can fill out the
<path> in the url pattern itself in get_absolute_url()

Issue: #167

### DIFF
--- a/galaxy_ng/app/api/v3/viewsets.py
+++ b/galaxy_ng/app/api/v3/viewsets.py
@@ -1,6 +1,7 @@
 import logging
 
 from django.core.exceptions import ValidationError
+from django.urls import reverse
 
 from rest_framework.response import Response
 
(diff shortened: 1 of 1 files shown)


alikins:
A slightly different approach is at https://github.com/ansible/galaxy_ng/compare/master...alikins:task_url_167_method

But really, galaxy_ng CollectionImport needs to be able to provide it's own get_absolute_url() for an instance, which means it needs to be able to get it's Distribution.base_path presumably via AnsibleRepository fk.

PR #173, opened 2020-05-16
Enforce CSRF protection for login view

cutwater:


cutwater:
@newswangerd You to GET request to the `.../auth/login` endpoint.

galaxy_ng/app/api/ui/views/auth.py
```diff
@@ -19,10 +22,24 @@
...
+        return user, None
+
+
 class LoginView(api_base.GenericAPIView):
     serializer_class = LoginSerializer
+    authentication_classes = (_CSRFSessionAuthentication,)
     permission_classes = (RestrictOnCloudDeployments,)
 
+    @method_decorator(ensure_csrf_cookie)
+    def get(self, request, *args, **kwargs):
```
alikins, line 40: To verify I understand correctly, the _CSRFSessionAuthentication Authentication class is used to enfore existence of a csrf token, even for unauthenticated or "safe" operations.

The 'get' method here requires user, auth, and csrf token and returns an empty response body. But the ensure_csrf_cookie() decorator will set the right response headers to include any needed fields for csrf
(either in the django session, or a csrf token in a cookie).

cutwater:
@alikins it isn't about requiring CSRF for "safe" methods. It's about requiring CSRF when no `sessionid` cookie is set.

galaxy_ng/app/api/ui/views/auth.py
cutwater: See my comment below. For unauthenticated and unsafe. Specifically for `POST` method to `.../auth/login`.

PR #176, opened 2020-05-18
[WIP] ModelAdmin objects for galaxy_ng for django-admin 

alikins:
Depends on pulp/pulpcore#630

### DIFF
--- a/galaxy_ng/app/admin.py
+++ b/galaxy_ng/app/admin.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 from django.contrib import admin
 from django.contrib.auth.models import Permission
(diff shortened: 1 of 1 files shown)


cutwater:
If it's not ready for review and depends on external PR that is not merged and it's merge time is unknown, please submit it as a DRAFT PR, to minimize noise in PRs list.

alikins:
It sounds like django-admin is going to be enabled in upcoming versions of pulpcore, so this will be even more useful then. (Since django-admin will be enabled by default, at least for standalone)

alikins:
Obsoleted by https://github.com/ansible/galaxy_ng/pull/369  which is based on changed needed for RBAC in pulpcore and pulp_galaxy.

Issue #170, opened 2020-05-15
Story: Auto-certify content


chouseknecht:
Add a configuration option for automatically certifying content. For local AH this will be enabled by default, and for c.rh.c it will be disabled. 

When enabled, a collection will automatically move into the 'certified' content repository upon successful import.

chouseknecht:
For 4.2 we're not implementing multiple repositories into the certification process. Instead, the importer will set the certification flag to certified.

PR #202, opened 2020-05-22
Auto-approve collection version on import

awcrosby:
* Add config setting for when to require approval of a collection version
* Add galaxy_ng task to call pulp_ansible `import_collection` task then auto-approve collection version
* In CollectionUploadViewSet dispatch pulp task from pulp_ansible or galaxy_ng based on config

Works #170

### DIFF
--- a/galaxy_ng/app/tasks/publishing.py
+++ b/galaxy_ng/app/tasks/publishing.py
@@ -1,0 +1,25 @@
+import logging
+
+from pulp_ansible.app.tasks.collections import import_collection
+from pulpcore.plugin.models import ContentArtifact
+
+log = logging.getLogger(__name__)
+
+VERSION_CERTIFIED = "certified"
+
+
+def import_and_auto_approve(artifact_pk, **kwargs):
+    """Import collection version and automatically approve.
+
+    Custom task to call pulp_ansible's import_collection() task
+    then automatically approve collection version so no
+    manual approval action needs to occur.
+
+    Approval currently is a collection version certification flag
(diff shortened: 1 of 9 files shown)


alikins:
@awcrosby Whats the minimal pulp_ansible version we need for this? Do the requirements in setup.py need to change?

galaxy_ng/app/tasks/publishing.py
```diff
@@ -0,0 +1,25 @@
...
+    """Import collection version and automatically approve.
+
+    Custom task to call pulp_ansible's import_collection() task
+    then automatically approve collection version so no
+    manual approval action needs to occur.
+
+    Approval currently is a collection version certification flag
+    Approval later will be moving from a staging to an approved repo
+    """
+    import_collection(artifact_pk=artifact_pk, repository_pk=kwargs.get('repository_pk', None))
```
alikins, line 21: Wondering if running a sub task like this needs to use enqueue? And/or task groups? I would kind of expect the outer task to end up enqueue'ing the inner task, but I don't know if that is required here. The only example I see of a task running a subtask is https://github.com/pulp/pulpcore/blob/master/pulpcore/app/tasks/importer.py#L194

cutwater: Using enqueue would mean that this task has to wait until the child task is finished.

awcrosby:
@alikins right now this needs pulp_ansible master, it relies on https://github.com/pulp/pulp_ansible/commit/6c733e342e8aa3b96abee5273fd4728ae0838bdb which has not been released

PR #208, opened 2020-05-28
Add support for running UI in compose.sh

newswangerd:
Requires: https://github.com/ansible/ansible-hub-ui/pull/143

Adds
- `ANSIBLE_HUB_UI_PATH` environment variables that can be used to specify the location of the UI repo for running the UI with `./compose`

#217


dev/insights/docker-compose-ui.yaml
```diff
@@ -0,0 +1,32 @@
+version: "3.7"
```
cutwater, line 1: Looks like insights and standalone docker-compose files are similar.
Can you extract parts that are common for both environments into common compose file?

dev/insights/docker-compose-ui.yaml
```diff
@@ -0,0 +1,32 @@
...
+      context: "${ANSIBLE_HUB_UI_PATH}"
+    environment:
+      - "DEPLOYMENT_MODE=${COMPOSE_PROFILE}"
+    ports:
+      - "8002:8002"
+    volumes:
+      - "${ANSIBLE_HUB_UI_PATH}:/hub/app/"
+      # Forces npm to ignore the node_node modules in the volume and look
+      # for it in ../node_modules instead
+      - "nodata:/hub/app/node_modules:ro"
```
cutwater, line 15: Since it's read only probably no need to give it a name.

.compose.env.example
```diff
@@ -4,3 +4,7 @@ COMPOSE_PROFILE=standalone
 
 # Add extra paths to run project dependencies from local filesystem in editable mode
 DEV_SOURCE_PATH='galaxy_ng'
+
+# If you want to run the UI, clone https://github.com/ansible/ansible-hub-ui
+# and set this to the path for the UI
+# ANSIBLE_HUB_UI_PATH='/absolute/path/to/ansible-hub-ui'
```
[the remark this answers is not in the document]
newswangerd: It wasn't super clear to me what the path was relative to. It looks like it's relative to `dev/docker-compose.yml`, which means you need to prepend `..` to the path, but that seemed kind of messy. I'll give it a try.

PR #234, opened 2020-06-10
Fix for bad multipart posts workaround got rm, readd

alikins:
In the move to use the pulp viewsets, the previous
workaround for ansible/galaxy-dev#246 didn't
get added back. That was added in 035371f819a168b
but got removed in 1245d37bf5.

So, add it back.

This manifests itself with older versions of ansible-galaxy
on 'publish'. If the server doesn't workaround the multipart
parser, the sha256 value in the body will be ignored, and
cause 400 errors about the sha256sum mismatch.

### DIFF
--- a/CHANGES/256.bugfix
+++ b/CHANGES/256.bugfix
@@ -1,0 +1,1 @@
+Add back workaround for multipart forms from ansible-galaxy.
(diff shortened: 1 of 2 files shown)


chouseknecht:
@alikins This needs an issue and change log snippet

PR #250, opened 2020-06-17
Versioned UI API

newswangerd:
- Adds current `v3/_ui/` api to `_ui/v1/`
- Adds `UIAPIVersions` constant which contains a list of active versions and the current version
- Adds `UIVersioning` class to all our UI viewsets. This provides the version information through the `self.request.version` variable. This variable can be used to change the bahaviour of the API base on the verision that's being accessed. An example of this can be found here: https://www.django-rest-framework.org/api-guide/versioning/#versioning-with-rest-framework


galaxy_ng/app/constants.py
```diff
@@ -10,3 +10,8 @@ class CertificationStatus(enum.Enum):
 class DeploymentMode(enum.Enum):
     STANDALONE = 'standalone'
     INSIGHTS = 'insights'
+
+
+class UIAPIVersions(enum.Enum):
```
cutwater, line 15: It doesn't seem like valid Enum usage. Here enum is not a type that represents static set of values, but rather a namespace for different constants. Please remove this enum and replace it with constants.

Issue #119, opened 2020-04-24
UI: Certification dashboard edits for content promotion


awcrosby:
Subtask of #20

Have certification-dashboard use new viewset and new api promotion endpoints for accept and reject

Issue #118, opened 2020-04-24
API: Add per-repo _ui viewset for certification dashboard


awcrosby:
Subtask of #20

Add a per-repo viewset for current endpoint `/api/automation-hub/v3/_ui/collection-versions/` to be used in approving and rejecting in the certification dashboard

PR #253, opened 2020-06-18
Support query by repo for _ui CollectionVersionViewSet

awcrosby:
Closes-Issue: #118
* Adds query by repo
* Removes the use of galaxy_pulp for list() and retrieve()

With this, the UI can make these changes to the Certification Dashboard (#119):
* Change query parameter used from `certification` to `repository`
* Change approval collection version api call from `certification/` to `move/` (#140)

This leaves the certification/ endpoint working until the UI stops using it

### DIFF
--- a/galaxy_ng/app/api/ui/viewsets/collection.py
+++ b/galaxy_ng/app/api/ui/viewsets/collection.py
@@ -1,9 +1,11 @@
 import galaxy_pulp
 from django.conf import settings
+from django.core.exceptions import ObjectDoesNotExist
 from django_filters import filters
 from django_filters.rest_framework import filterset, DjangoFilterBackend, OrderingFilter
 from drf_yasg.utils import swagger_auto_schema
 from rest_framework.decorators import action as drf_action
+from pulp_ansible.app.models import AnsibleDistribution, CollectionVersion, Collection
 from rest_framework.exceptions import NotFound
 from rest_framework.generics import get_object_or_404
 from rest_framework.response import Response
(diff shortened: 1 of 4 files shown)


galaxy_ng/app/api/ui/viewsets/collection.py
```diff
@@ -127,39 +129,36 @@ class CollectionVersionViewSet(api_base.GenericViewSet):
...
-        api = galaxy_pulp.PulpCollectionsApi(pulp.get_client())
-        response = api.list(namespace=namespace, name=name, version=version, limit=1)
-
-        if not response.results:
-            raise NotFound()
-
-        data = serializers.CollectionVersionDetailSerializer(response.results[0]).data
-        return Response(data)
-
+        try:
```
alikins, line 150: Could this use django.shortcut.get_object_or_404() ?

(Though, the error messages created by get_object_or_404() aren't great... maybe we should add a version of get_object_or_404 that generates better errors and use it where the try/except ObjectDoesNotExist are?)

galaxy_ng/app/api/ui/viewsets/collection.py
```diff
@@ -127,39 +129,36 @@ class CollectionVersionViewSet(api_base.GenericViewSet):
...
-            del params['sort']
+        queryset = CollectionVersion.objects.all()
+        repo = self.request.query_params.dict().get('repository')
+        if repo:
+            try:
+                distro = AnsibleDistribution.objects.get(base_path=repo)
+            except ObjectDoesNotExist:
+                raise NotFound(f'Repository not found: {repo}')
+            repository_version = distro.repository.latest_version()
+            queryset = queryset.filter(pk__in=repository_version.content)
```
alikins, line 140: Could / should this be a FilterSet ?

galaxy_ng/app/api/ui/viewsets/collection.py
```diff
@@ -127,39 +129,36 @@ class CollectionVersionViewSet(api_base.GenericViewSet):
...
-
-        # pulp uses ordering, but the UI is standardized around sort
-        if params.get('sort'):
-            params['ordering'] = params.get('sort')
-            del params['sort']
+        queryset = CollectionVersion.objects.all()
+        repo = self.request.query_params.dict().get('repository')
+        if repo:
+            try:
+                distro = AnsibleDistribution.objects.get(base_path=repo)
```
alikins, line 136: The pulp_ansible collection viewsets have some mixins and whatnot for doing some of this.
https://github.com/pulp/pulp_ansible/blob/master/pulp_ansible/app/galaxy/v3/views.py#L38

Not sure if the pulp mixin approach is worthwhile, but it's an option. Might make it a little easier to bring the ui collections views and the v3 / pulp_ansible collection viewsets closer together.

galaxy_ng/app/api/ui/viewsets/collection.py
awcrosby: Yeah that is why I didn't use it because I couldn't add details about the object in the error message, but since it is a GET on the url params maybe the 404 without object info in the error message is ok

galaxy_ng/app/api/ui/viewsets/collection.py
awcrosby: Yeah that probably is worth a try

galaxy_ng/app/api/ui/viewsets/collection.py
awcrosby: I'll look at that

galaxy_ng/app/api/ui/viewsets/collection.py
```diff
@@ -127,39 +129,36 @@ class CollectionVersionViewSet(api_base.GenericViewSet):
...
+            except ObjectDoesNotExist:
```
newswangerd, line 137: Don't return 404s on list endpoints. Just return an empty list.

galaxy_ng/app/api/ui/viewsets/collection.py
```diff
@@ -127,39 +129,36 @@ class CollectionVersionViewSet(api_base.GenericViewSet):
...
+                raise NotFound(f'Repository not found: {repo}')
```
newswangerd, line 138: Don't return a 404 on a list endpoint, just return an empty list.

galaxy_ng/app/api/ui/viewsets/collection.py
alikins: It just occurred to me that since _ui/collections/ and v3/collections/ are both views to the same Collection models, if we add any Collection view FilterSets we could potentially use them in both places.

alikins: (Not necessarily a requirement now though...)

galaxy_ng/app/api/ui/viewsets/collection.py
alikins: good catch.

galaxy_ng/app/api/ui/viewsets/collection.py
awcrosby: Added a FilterSet.

For sharing across `v3` and `_ui`... at least for CollectionVersion there are some differences: `v3` filters to just certified vs not in `_ui`, and `v3` uses url kwargs for namespace and name vs query params for those in `_ui`.

galaxy_ng/app/api/ui/viewsets/collection.py
awcrosby: This worked except we don't want the 404 when the repo is not found, so I scraped the mixin

galaxy_ng/app/api/ui/viewsets/collection.py
awcrosby: updated

awcrosby:
@newswangerd I added a filterset that included those query parameters to successfully filter, if there are other needed params we can add them

awcrosby:
@newswangerd `?sort` now works for those 4 params, I wasn't sure which ones were getting supported earlier via pulp api

Issue #247, opened 2020-06-16
Collection detail response contains invalid href


chouseknecht:
In our CI environment this endpoint https://ci.cloud.redhat.com/api/automation-hub/v3/collections/autohubtest2/collection_dep_a_hlrknkqw/ returns the following response object: 

```
{
    "href": "/pulp_ansible/galaxy/automation-hub/api/v3/collections/autohubtest2/collection_dep_a_hlrknkqw/",
    "created_at": "2020-02-21T23:01:52.727205Z",
    "updated_at": "2020-02-21T23:01:52.727232Z",
    "namespace": "autohubtest2",
    "name": "collection_dep_a_hlrknkqw",
    "deprecated": false,
    "versions_url": "/pulp_ansible/galaxy/automation-hub/api/v3/collections/autohubtest2/collection_dep_a_hlrknkqw/versions/",
    "highest_version": {
        "href": "/pulp_ansible/galaxy/automation-hub/api/v3/collections/autohubtest2/collection_dep_a_hlrknkqw/versions/1.0.0/",
        "version": "1.0.0"
    }
}
```

Notice the invalid `href` value in the above response. It should not contain `/pulp_ansible/galaxy`.

This issue is blocking work related to adding token support to content sync in the [pulp_installer](https://github.com/pulp/pulp_ansible) project. For details see: https://github.com/pulp/pulp_ansible/pull/326#issuecomment-644313611

alikins:
The serializers used by the pulp collection viewsets (and therefore galaxy_ng collection viewsets) base the values for 'href' on url reverse() lookups, and end up with the pulp paths.

May mean galaxy_ng viewsets need to provide their own serializer. 

But may be able to parameterize the reverse()'s to handle both cases. Possibly via providing a ViewSet.get_serializer_context()

chouseknecht:
@alikins IIRC, we decided on option 1 during standup on Tuesday (16 June).

alikins:
https://github.com/ansible/galaxy_ng/pull/254/ has a proposed fix that doesn't rely on any pulp_ansible changes.

It seems to work, but might be a bit weird. Could probably de-weirdify it a bit with some changes in pulp_ansible. The pulp_ansible changes would likely be similar (taking the url namespace into account, etc).

PR #254, opened 2020-06-18
Fix wrong href's in results from collection viewsets

alikins:
The pulp_ansible viewsets build hrefs by reverse() on the
pulp_ansible url name. But the galaxy_ng names are different
and include url namespaces ('galaxy:api:v3:collection' for ex).

Makes collection api endpoints serialize correctly, with the various
href's pointing to the right places. That includes:

  1. Not pointing to the pulp hrefs
  1. pointing to either the default-content endpoint or the per-repo
endpoints

Update the collection url 'names' so that suffix matches the
upstream pulp names exactly (sanes url namespace)

Add custom collection Serializer's that uses view_namespace for url reverse.

Add a serializers.collections.HrefNamespaceMixin to use
with the galaxy_ng subclasses of collection related
pulp_ansible.app.galaxy Serializers.

Update the collection related viewsets to use the new
viewsets.

Add a ViewNamespaceSerializerContextMixin to mixin
with the collection viewset. This provides a
get_serializer_context that adds 'view_namespace'
and 'view_route' info to the context, so the various
href serializer methods will generate the correct urls.

Issue: #247

### DIFF
--- a/CHANGES/247.bugfix
+++ b/CHANGES/247.bugfix
@@ -1,0 +1,1 @@
+Fix wrong href's in results from collection viewsets
--- a/galaxy_ng/app/api/v3/serializers/__init__.py
+++ b/galaxy_ng/app/api/v3/serializers/__init__.py
@@ -1,5 +1,7 @@
 from .collection import (
+    CollectionSerializer,
     CollectionVersionSerializer,
+    CollectionVersionListSerializer,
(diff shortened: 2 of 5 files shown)


chouseknecht:
LGTM. I really want to get this fixed, so happy to see this.

Issue #37, opened 2020-03-19
API: Manage namespace repositories


chouseknecht:
Subtask of #20 

- [x] When a namespace is created, the API will create a corresponding repository and distribution. The distribution should contain the namespace name.
- [x] When a namespace is deleted, remove the corresponding repository and distribution.

PR #272, opened 2020-07-02
Auto create and delete inbound pulp repo per namespace

awcrosby:
Closes-Issue: #37
* Create and delete repo on api namespace create and delete
* The repo is tied to the namespace by name
* This applies to pulp Repository and Distribution
* Convention uses same text for names and base_path of repo and distro

### DIFF
--- a/galaxy_ng/app/api/v3/viewsets/namespace.py
+++ b/galaxy_ng/app/api/v3/viewsets/namespace.py
@@ -1,7 +1,9 @@
 import logging
 
+from django.db import transaction
 from django_filters import filters
 from django_filters.rest_framework import filterset, DjangoFilterBackend
+from pulp_ansible.app.models import AnsibleRepository, AnsibleDistribution
 
 from galaxy_ng.app import models
 from galaxy_ng.app.api import permissions
@@ -9,6 +11,8 @@ from galaxy_ng.app.api import base as api_base
 from galaxy_ng.app.api.v3 import serializers
 
 log = logging.getLogger(__name__)
+
+INBOUND_REPO_NAME_FORMAT = "inbound-{namespace_name}"
 
 
 class NamespaceFilter(filterset.FilterSet):
@@ -52,3 +56,27 @@ class NamespaceViewSet(api_base.ModelViewSet):
             permission_list.append(permissions.IsNamespaceOwnerOrPartnerEngineer())
 
         return permission_list
+
+    @transaction.atomic
+    def create(self, request, *args, **kwargs):
+        """Override to also create inbound pulp repository and distribution."""
+        name = INBOUND_REPO_NAME_FORMAT.format(namespace_name=request.data['name'])
+        repo = AnsibleRepository.objects.create(name=name)
+        AnsibleDistribution.objects.create(name=name, base_path=name, repository=repo)
+        return super().create(request, *args, **kwargs)
+
+    @transaction.atomic
+    def destroy(self, request, *args, **kwargs):
+        """Override to also delete inbound pulp repository and distribution.
+        RepositoryVersion objects get deleted on delete of AnsibleRepository.
+        """
+        name = INBOUND_REPO_NAME_FORMAT.format(namespace_name=kwargs['name'])
+        try:
+            AnsibleRepository.objects.get(name=name).delete()
+        except AnsibleRepository.DoesNotExist:
+            pass
+        try:
+            AnsibleDistribution.objects.get(name=name).delete()
+        except AnsibleDistribution.DoesNotExist:
+            pass
+        return super().destroy(request, *args, **kwargs)
--- a/galaxy_ng/app/migrations/0003_inbound_repo_per_namespace.py
+++ b/galaxy_ng/app/migrations/0003_inbound_repo_per_namespace.py
@@ -1,0 +1,31 @@
+from django.db import migrations
+
+def create_inbound_repo_per_namespace(apps, schema_editor):
+    AnsibleRepository = apps.get_model('ansible', 'AnsibleRepository')
+    AnsibleDistribution = apps.get_model('ansible', 'AnsibleDistribution')
+    Namespace = apps.get_model('galaxy', 'Namespace')
+    db_alias = schema_editor.connection.alias
+
+    for namespace in Namespace.objects.using(db_alias).all():
+        name = f'inbound-{namespace.name}'
+        repo = AnsibleRepository.objects.using(db_alias).create(
+            name=name,
+            pulp_type='ansible.ansible',
+        )
+        distro = AnsibleDistribution.objects.using(db_alias).create(
+            name=name,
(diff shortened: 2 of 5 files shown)


galaxy_ng/app/api/v3/viewsets/namespace.py
```diff
@@ -52,3 +55,21 @@ def get_permissions(self):
             permission_list.append(permissions.IsNamespaceOwnerOrPartnerEngineer())
 
         return permission_list
+
+    @transaction.atomic
+    def create(self, request, *args, **kwargs):
```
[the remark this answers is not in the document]
awcrosby: I'm not sure there's a need for a db releationship between AnsibleRepository and Namespace.

As far as using a django manager versus viewset for creation of the repository, no strong opinion either way.

galaxy_ng/app/api/v3/viewsets/namespace.py
```diff
@@ -52,3 +55,21 @@ def get_permissions(self):
...
+        """Override to also create inbound pulp repository and distribution."""
+        name = INBOUND_REPO.format(namespace_name=request.data['name'])
+        repo = AnsibleRepository.objects.create(name=name)
+        AnsibleDistribution.objects.create(name=name, base_path=name, repository=repo)
```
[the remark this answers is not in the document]
awcrosby: Agreed, I will add those to this PR

galaxy_ng/app/migrations/0003_inbound_repo_per_namespace.py
```diff
@@ -0,0 +1,30 @@
+from django.db import migrations
+
+def create_inbound_repo_per_namespace(apps, schema_editor):
+    AnsibleRepository = apps.get_model('ansible', 'AnsibleRepository')
+    AnsibleDistribution = apps.get_model('ansible', 'AnsibleDistribution')
+    Namespace = apps.get_model('galaxy', 'Namespace')
+
+    for namespace in Namespace.objects.all():
```
cutwater, line 8: Better specify db_alias explicitly as it's recommended in Django [docs](https://docs.djangoproject.com/en/3.0/ref/migration-operations/#runpython).

galaxy_ng/app/api/v3/viewsets/namespace.py
```diff
@@ -52,3 +55,21 @@ def get_permissions(self):
...
+        return super().create(request, *args, **kwargs)
+
+    @transaction.atomic
+    def destroy(self, request, *args, **kwargs):
+        """Override to also delete inbound pulp repository and distribution.
+        RepositoryVersion objects get deleted on delete of AnsibleRepository.
+        """
+        name = INBOUND_REPO.format(namespace_name=kwargs['name'])
+        AnsibleRepository.objects.get(name=name).delete()
```
cutwater, line 73: This code doesn't handle NotFound exception. IIRC it will end up with 500 response.

galaxy_ng/app/migrations/0003_inbound_repo_per_namespace.py
alikins: Nice tip!

galaxy_ng/app/migrations/0003_inbound_repo_per_namespace.py
```diff
@@ -0,0 +1,30 @@
...
+        name = f'inbound-{namespace.name}'
```
alikins, line 9: Can migrations make use of django.settings?  (I assume so but have never tried it)
If so, I wonder if it would be worthwhile to make the repo naming format a settings option? 

I pulled the synclist naming format out to a config option, mostly just as a way to define it one place. I don't expect it to change, though hard to tell what standalone users will want.

alikins:
Migrations look good to me.

galaxy_ng/app/migrations/0003_inbound_repo_per_namespace.py
awcrosby: Added db_alias

galaxy_ng/app/api/v3/viewsets/namespace.py
awcrosby: Updated

galaxy_ng/app/migrations/0003_inbound_repo_per_namespace.py
awcrosby: I think defining the format in one place is good, I added it to settings like the synclist format. … I was thinking this migration might not need to last forever, as we are making the data “catch-up” to the code, I added `elidable=True` so it will be removed in any future squashing of migrations https://docs.djangoproject.com/en/3.0/ref/migration-operations/#django.db.migrations.operations.RunPython @alikins @cutwater - does that ^ sound right?

galaxy_ng/app/settings.py
```diff
@@ -30,6 +30,10 @@
 # created by the initial-data.json data fixture
 GALAXY_API_DEFAULT_DISTRIBUTION_BASE_PATH = "automation-hub"
 
+# The format for the name of the per namespace inbound repository,
+# and the associated distribution, and distribution base_path
+GALAXY_API_INBOUND_REPO_NAME_FORMAT = "inbound-{namespace_name}"
```
cutwater, line 35: I think it is better to be a constant, rather than settings.

awcrosby: PR #289 has a proposed `GALAXY_API_SYNCLIST_NAME_FORMAT = "{account_number}-synclist"` in settings, im ok either way but probably should be consistent

alikins: synclist and account_number is kind of inherently automation-hub / cloud.redhat.com specific, so lends itself to settings. And if any future versions of standalone would support synclists, the default synclist naming would need to change, so also more suitable for settings.

In theory, namespace probably won't change for c.r.c or standalone. Although I could see a case for standalone where folks deploying standalone may want to change the default inbound repos. For ex, maybe having a single inbound repo, or multiple inbound repos one for each team, or even one inbound repo for each collection. Or not having any local namespaces at all and having inbound repos.  But maybe that means inbound-repo-per-namespace is really a c.r.c specific feature which suggests driving it's behavior from settings make sense.

awcrosby: Given some of those scenarios such as multiple inbound repos cannot be handled with a settings change, and since this probably won't change, I moved this to a local module constant.

galaxy_ng/app/migrations/0003_inbound_repo_per_namespace.py
alikins: If we are relying on the inbound-repo naming to be consistent  (ie, if there is code that assumes Namespace.objects.get(name='inbound-mynamespace')), at some point we may need a join table to make that connection more explicit and robust. But probably not something to worry about now, I hope.

PR #280, opened 2020-07-08
Build sdist and wheel packages including static assets

cutwater:


setup.py
```diff
@@ -6,32 +6,48 @@
...
 
     def run(self):
-        target_dir = os.path.join(self.build_lib, "galaxy_ng/app/static/galaxy_ng")
+        if os.path.exists(self.TARGET_DIR):
+            log.warn(f"Static directory {self.TARGET_DIR} already exists, skipping. ")
+            return
 
         with tempfile.NamedTemporaryFile() as download_file:
-            log.info(f"Downloading UI distribution {download_file.name}")
+            log.info(f"Downloading UI distribution to temporary file: {download_file.name}")
```
awcrosby, line 37: I noticed when running `python3 setup.py sdist bdist_wheel` if static files are not present, after this download/extract step happens, the source tarball does not have a `static/` directory (but logs show the wheel getting it). Running the command again shows `creating galaxy-ng-4.2.0a8/galaxy_ng/app/static` as I would expect.

cutwater: Yes, you're right. It doesn't work as expected.

awcrosby: After the update, this looks like it is fixed.

PR #284, opened 2020-07-08
Add repository_list to data returned in _ui collection-version endpoint

awcrosby:
For the certification dashboard, add a `repository_list` field to the _ui collection-version endpoint.

This field includes a `Repository` if the `CollectionVersion` is included in the latest `RepositoryVersion`

An example:
```
"repository_list": [
    "automation-hub",
    "org-my_org"
]
```

### DIFF
--- a/CHANGES/288.misc
+++ b/CHANGES/288.misc
@@ -1,0 +1,1 @@
+Add repository_list to ui collection version endpoint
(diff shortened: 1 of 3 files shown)


chouseknecht:
@newswangerd @ZitaNemeckova 

The Javascript cannot hardcode `automation-hub` as the name for the golden repository. When in standalone mode, the name will be `published` and on SaaS it will be `automation-hub`. We may want to go one step further and future proof it, such that the golden repository (the repository containing certified collections) can have an arbitrary name.

chouseknecht:
Follow-up... We decide in chat that we're not going to future proof this now. For standalone mode the name will be `publshed`, and for SaaS the name will be 'automation-hub'.

Issue #285, opened 2020-07-09
API: Update org repositories when new collection version published


chouseknecht:
Subtask of #20 

For c.rh.c, when a collection version is promoted from the *staging* repository to the *published* repository, the subscriber org repositories must be updated with the new artifact. 

The promotion event has to:
 - Kick-off n number of tasks, where n is the number of org repos
 - Each task will:
     - run a db query that includes all collection versions from the *certified* repo, subtracting any denylist colletions for the org
     - produce an org repository version containing the result of the query

chouseknecht:
Notes from Galaxy/Pulp check-in meeting regarding task development: 
-  https://docs.pulpproject.org/plugins/plugin-writer/concepts/index.html#tasks
    - Group task report - reports progress on a group of tasks
    - Could be used to surface promotion status in UI

alikins:
https://github.com/ansible/galaxy_ng/pull/315 is the WIP for this.

alikins:
This will be in https://github.com/ansible/galaxy_ng/pull/315/. 

This subtask is where most of the ~8-10 hours of work needed for https://github.com/ansible/galaxy_ng/issues/265#issuecomment-671365304 comes from.

alikins:
https://github.com/ansible/galaxy_ng/pull/427  adds an api endpoint for triggering a curate on all synclist repos associated with golden repo at  /api/automation-hub/_ui/v1/repo/str:repo_name/curate/

https://github.com/ansible/galaxy_ng/pull/432  also provides a mechanism for also triggering the golden repo -> synclist-repos curration when content is 'moved' in or out of the golden repo

alikins:
See also https://github.com/ansible/galaxy_ng/issues/428

alikins:
One implementation is in #432

PR #287, opened 2020-07-09
Content Sync: On content change in the golden repo, replicate in all org repos

awcrosby:
Works #285 

In Insights mode, when a `CollectionVersion` is added to or removed from the golden repo, the same action should be replicated in the appropriate org repos.

A `CollectionVersion` promoted to the golden repo via the `CollectionVersion` `move/` endpoint will kick off N tasks to add the `CollectionVersion` to N org repos. Where N is the number of org repos where the parent `Collection` is not on the org's denylist.

A `CollectionVersion` removed from the golden repo via the `CollectionVersion` `move/` endpoint will kick off N tasks to remove the `CollectionVersion` from N org repos. Where N is the total number of org repos.

The Certification Dashboard uses the CollectionVersion `move/` endpoint

### DIFF
--- a/galaxy_ng/app/api/v3/viewsets/collection.py
+++ b/galaxy_ng/app/api/v3/viewsets/collection.py
@@ -285,7 +285,13 @@ class CollectionVersionMoveViewSet(ViewSet):
             settings.GALAXY_DEPLOYMENT_MODE == DeploymentMode.INSIGHTS.value and
             dest_repo == AnsibleRepository.objects.get(name='automation-hub')  # use constant or param
         ):
-            dispatch_copy_to_org_repos(collection_version)
+            dispatch_tasks_to_org_repos(collection_version, action='add')
+
+        if (
+            settings.GALAXY_DEPLOYMENT_MODE == DeploymentMode.INSIGHTS.value and
+            src_repo == AnsibleRepository.objects.get(name='automation-hub')  # use constant or param
+        ):
+            dispatch_tasks_to_org_repos(collection_version, action='remove')
 
         return Response(
             data={
--- a/galaxy_ng/app/api/v3/viewsets/collection.py
+++ b/galaxy_ng/app/api/v3/viewsets/collection.py
@@ -281,6 +281,12 @@ class CollectionVersionMoveViewSet(ViewSet):
         add_task = self._add_content(collection_version, dest_repo)
         remove_task = self._remove_content(collection_version, src_repo)
 
+        if (
+            settings.GALAXY_DEPLOYMENT_MODE == DeploymentMode.INSIGHTS.value and
+            dest_repo == AnsibleRepository.objects.get(name='automation-hub')  # use constant or param
+        ):
+            dispatch_copy_to_org_repos(collection_version)
+
         return Response(
             data={
                 'add_task_id': add_task.id,
--- a/galaxy_ng/app/tasks/promotion.py
+++ b/galaxy_ng/app/tasks/promotion.py
@@ -28,8 +28,8 @@ def remove_content_from_repository(collection_version_pk, repository_pk):
         new_version.remove_content(qs)
 
 
-def dispatch_copy_to_org_repos(collection_version_pk):
-    """Starts tasks to copy CollectionVersion to org repos.
+def dispatch_tasks_to_org_repos(collection_version_pk, action):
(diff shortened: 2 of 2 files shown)


galaxy_ng/app/tasks/promotion.py
```diff
@@ -25,3 +26,29 @@ def remove_content_from_repository(collection_version_pk, repository_pk):
...
         new_version.remove_content(qs)
+
+
+def dispatch_copy_to_org_repos(collection_version_pk):
+    """Starts tasks to copy CollectionVersion to org repos.
+    Method to start tasks, not a task itself."""
+
+    # get all org distros
+    # TODO: replace with correct convention, regex in constants?
+    org_distros = AnsibleDistribution.objects.filter(name__startswith='org-')
```
chouseknecht, line 37: I guess looking for `org-` in the name is OK, since this is confined to c.rh.c

galaxy_ng/app/api/v3/viewsets/collection.py
```diff
@@ -281,6 +281,18 @@ def move_content(self, request, *args, **kwargs):
         add_task = self._add_content(collection_version, dest_repo)
         remove_task = self._remove_content(collection_version, src_repo)
 
+        if (
+            settings.GALAXY_DEPLOYMENT_MODE == DeploymentMode.INSIGHTS.value and
+            dest_repo == AnsibleRepository.objects.get(name='automation-hub')  # use constant or param
```
alikins, line 286: I've been wondering if there is a good way to have a common "get_the_default_repo()" since it comes up in lots of synclist stuff.

galaxy_ng/app/api/v3/viewsets/collection.py
```diff
@@ -281,6 +281,18 @@ def move_content(self, request, *args, **kwargs):
         add_task = self._add_content(collection_version, dest_repo)
         remove_task = self._remove_content(collection_version, src_repo)
 
+        if (
+            settings.GALAXY_DEPLOYMENT_MODE == DeploymentMode.INSIGHTS.value and
```
alikins, line 285: Could this use galaxy_ng.app.api.permissions.RestrictOnCloudDeployments  or it's friends?

galaxy_ng/app/tasks/promotion.py
alikins: In theory, looking up 'base_path'  would be more specific, but:

1) For now, Distro name and base_path will be the same.
2) I'm not 100% sure 'base_path' get's a db index though probably does, but name definitely should.

galaxy_ng/app/tasks/promotion.py
```diff
@@ -25,3 +26,35 @@ def remove_content_from_repository(collection_version_pk, repository_pk):
...
+
+    # filter out distros which point to golden repo
+    golden = AnsibleRepository.objects.get(name='automation-hub')  # use constant or param
+    org_repos = set([d.repository for d in org_distros])
+    try:
+        org_repos.remove(golden)
+    except KeyError:
+        pass
+
```
alikins, line 46: This may be a good place to use the pulpcore 'TaskGroup' stuff  
 https://docs.pulpproject.org/plugins/plugin-writer/concepts/index.html#tasks
https://github.com/pulp/pulpcore/blob/master/pulpcore/app/models/task.py#L423
https://docs.pulpproject.org/restapi.html#tag/task-groups

alikins: ie, so their is a TaskGroup for all the tasks that are updating repos, and that TaskGroup has progress status showing how many of the tasks are done.

galaxy_ng/app/tasks/promotion.py
```diff
@@ -25,3 +26,35 @@ def remove_content_from_repository(collection_version_pk, repository_pk):
...
+            enqueue_with_reservation(remove_content_from_repository, locks, args=task_args)
+    elif action == 'add':
+        # filter out repos with this Collection in their denylist
+        # TODO ^
+
+        # dispatch copy tasks
+        for repo in org_repos:
+            locks = [repo]
+            task_args = (collection_version_pk, repo.pk)
+            enqueue_with_reservation(add_content_to_repository, locks, args=task_args)
```
alikins, line 60: What all ends up in the db transactions?  Feels like all the changes to a org repo should be in a single transaction.

Issue #269, opened 2020-07-01
API: PATCH broken on my-synclists


newswangerd:
PATCH is broken on the my-synclists endpoint. Right now it throws a 500

```
api_1               | Traceback (most recent call last):
api_1               |   File "/venv/lib64/python3.6/site-packages/django/core/handlers/exception.py", line 34, in inner
api_1               |     response = get_response(request)
api_1               |   File "/venv/lib64/python3.6/site-packages/django/core/handlers/base.py", line 115, in _get_response
api_1               |     response = self.process_exception_by_middleware(e, request)
api_1               |   File "/venv/lib64/python3.6/site-packages/django/core/handlers/base.py", line 113, in _get_response
api_1               |     response = wrapped_callback(request, *callback_args, **callback_kwargs)
api_1               |   File "/venv/lib64/python3.6/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
api_1               |     return view_func(*args, **kwargs)
api_1               |   File "/venv/lib64/python3.6/site-packages/rest_framework/viewsets.py", line 114, in view
api_1               |     return self.dispatch(request, *args, **kwargs)
api_1               |   File "/venv/lib64/python3.6/site-packages/rest_framework/views.py", line 505, in dispatch
api_1               |     response = self.handle_exception(exc)
api_1               |   File "/venv/lib64/python3.6/site-packages/rest_framework/views.py", line 465, in handle_exception
api_1               |     self.raise_uncaught_exception(exc)
api_1               |   File "/venv/lib64/python3.6/site-packages/rest_framework/views.py", line 476, in raise_uncaught_exception
api_1               |     raise exc
api_1               |   File "/venv/lib64/python3.6/site-packages/rest_framework/views.py", line 502, in dispatch
api_1               |     response = handler(request, *args, **kwargs)
api_1               |   File "/venv/lib64/python3.6/site-packages/rest_framework/mixins.py", line 82, in partial_update
api_1               |     return self.update(request, *args, **kwargs)
api_1               |   File "/venv/lib64/python3.6/site-packages/rest_framework/mixins.py", line 67, in update
api_1               |     serializer.is_valid(raise_exception=True)
api_1               |   File "/venv/lib64/python3.6/site-packages/rest_framework/serializers.py", line 235, in is_valid
api_1               |     self._validated_data = self.run_validation(self.initial_data)
[...]

Issue #267, opened 2020-06-30
Content Sync: On synclist creation, create the repository


alikins:
When creating a synclist, ensure that the curated repo and distribution exists, and create them if needed.

This is related to https://github.com/ansible/galaxy_ng/issues/264 and https://github.com/ansible/galaxy_ng/issues/265

Current synclist API doesn't do this, so this is an add on for synclist API.

The repository name, and Distribution name and base_path should be unique and likely need to be based on account_number (as mentioned in  https://github.com/ansible/galaxy_ng/issues/264).

(There is WIP code for this, so adding issue for the pr)

chouseknecht:
We decide in the Pulp/Galaxy Check-in meeting on 7 July:

- on login create a unique distribution for the org, if it does not exist, and associate it with the published repo
- the first time a user adds a collection to the org's denylist, create a repo and associate it with the distribution, and then start a task that copies collection version from the published repo into the new repo.

alikins:
https://github.com/ansible/galaxy_ng/pull/315 is the WIP for this.

alikins:
Some work needed here, mostly tweaks related to pr review. Probably ~6 hours of work.

alikins:
https://github.com/ansible/galaxy_ng/pull/315

Issue #264, opened 2020-06-30
Content Sync: Create default synclist and associated repository/distribution on login


chouseknecht:
Subtask of #17 

- This is for c.rh.c only. It does not apply when running a local Galaxy.
- Whenever a subscriber logs in, check if the default synclist for their organiztion exists, and if not, create it.
- Default the synclist name, repository name, and distribution name to the group number + '-synclist'.

alikins:
@chouseknecht

>  Default the synclist name, repository name, and distribution name to the group number + '-synclist'.

If I understand correctly, for a custom in account "6297347"  that is in the group "rh-identity-account:6297347", the default synlists names would be something roughly like:

- 6297347-synclist

One small issue is that don't store the plain account number "6297347" on it's own anywhere in the data. It only exists as a substring in the group name "rh-identity-account:6297347".

Very easy to parse out the account number in that case of course, but it feels like something that would very likely break in the future.  if for ex:

- we stop using the rh-identity-account prefix for group name
- we allow standalone systems to use arbitrary group names  (not likely an issue for synclists, but if there are other features that rely on account number in the future)

Nothing that blocks anything now, I'll add a 'record account number in the db' issue to keep track of it.

chouseknecht:
> @chouseknecht
> 
> > Default the synclist name, repository name, and distribution name to the group number + '-synclist'.
> 
> If I understand correctly, for a custom in account "6297347" that is in the group "rh-identity-account:6297347", the default synlists names would be something roughly like:
> 
> * 6297347-synclist
> 

Yes. That seems right.

> One small issue is that don't store the plain account number "6297347" on it's own anywhere in the data. It only exists as a substring in the group name "rh-identity-account:6297347".
> 
> Very easy to parse out the account number in that case of course, but it feels like something that would very likely break in the future. if for ex:
> 
> * we stop using the rh-identity-account prefix for group name
> * we allow standalone systems to use arbitrary group names  (not likely an issue for synclists, but if there are other features that rely on account number in the future)
> 
> Nothing that blocks anything now, I'll add a 'record account number in the db' issue to keep track of it.

That seems look idea. +1

chouseknecht:
We decide in the Pulp/Galaxy Check-in meeting on 7 July:

- on login create a unique distribution for the org, if it does not exist, and associate it with the *published* repo
- the first time a user adds a collection to the org's denylist, create a repo and associate it with the distribution, and then start a task that copies collection version from the *published* repo into the new repo

[...]

PR #289, opened 2020-07-09
Create distro and synclist repo on login, toggle org distro between "golden"/synclist_repo as synclist becomes "active"

alikins:
Create a Repository/Distribution if needed when creating synclist

Fixes issues with PATCH and empty or non existent 'groups'

Add account_number() method to Group model, to extract
the account number info from the group name.

Add 'upstream_repository' to Synclist

Detect when a SyncList is a noop and point the related
Distribution to the default repository.

On synclist creation or modification, update the per-account
name format f"{account_number}-synclist" AnsibleDistribution
to point to either:

    - If the synclist is no-op-ish (ie, 'collections' is empty)
    then point the distro to the default AnsibleRepository
    (ie, "automation-hub")

    - If the synclist has collections to include/exclude that would make
    the synclist repo different than the default repo, then point
    the accounts f"{account_number}-synclist" AnsibleDistribution to their
    per-account synclist AnsibleRepository

The synclist name formation is set in
settings.GALAXY_API_SYNCLIST_NAME_FORMAT

Update the synclist 'policy' choice names to 'include', 'exclude'

Issue: #267
Issue: #269
Issue: #264


galaxy_ng/app/api/ui/serializers/synclist.py
```diff
@@ -45,23 +74,108 @@ def _sanitize_groups(self, group_names):
...
         data['groups'] = self._sanitize_groups(groups)
+
+        users = data.get('users', [])
+        data['users'] = self._sanitize_users(users)
+
+        repository_data = data.get('repository', None)
+        if repository_data:
+            data['repository'] = self._get_repository(repository_data)
+
+        upstream_repository_data = data.get('upstrean_repository', None)
```
awcrosby, line 99: s/upstrean_repository/upstream_repository

galaxy_ng/app/api/ui/serializers/synclist.py
```diff
@@ -45,23 +74,108 @@ def _sanitize_groups(self, group_names):
...
+        # lookup groups, to get a group name to use in naming the repository/distro
+        try:
+            description = f"Synclist repository for account {account_name}"
+
+            repository, created = AnsibleRepository.objects.get_or_create(name=name_slug)
+            repository.description = description
+            repository.save()
+            return repository
+        except IntegrityError as exc:
+            raise ValidationError("AnsibleRepository %s already exists: %s" % (name_slug, exc))
```
awcrosby, line 135: think this `IntegrityError` won't ever get triggered since using `get_or_create` with one parameter

galaxy_ng/app/api/ui/serializers/synclist.py
```diff
@@ -45,23 +74,108 @@ def _sanitize_groups(self, group_names):
...
+
+    def _update_distribution(self, name_slug, repository):
+        # Now the distro
+        try:
+            distribution, create = AnsibleDistribution.objects.get_or_create(
+                name=name_slug,
+                base_path=name_slug,
+            )
+        except IntegrityError as exc:
+            raise ValidationError("AnsibleDistribution %s already exists: %s" % (name_slug, exc))
```
awcrosby, line 148: i guess this would only trigger if name and base_path are different

galaxy_ng/app/settings.py
```diff
@@ -30,6 +30,10 @@
 # created by the initial-data.json data fixture
 GALAXY_API_DEFAULT_DISTRIBUTION_BASE_PATH = "automation-hub"
 
+# The format for the name of the per account synclist, and the
+# associated repository, distribution, and distribution base_paths
+GALAXY_API_SYNCLIST_NAME_FORMAT = "{account_name}-synclist"
```
awcrosby, line 35: Since this will also be the repo name... should we align the "type" and the "identifier" format for different repo types? ie either
1. `inbound-{namespace}` and `synclist-{account_name}` ... or ...
2. `{namespace}-inbound` and `{account_name}-synclist`

I'm kinda partial to the first one

galaxy_ng/app/api/ui/serializers/synclist.py
cutwater: +1

galaxy_ng/app/api/ui/serializers/synclist.py
```diff
@@ -45,23 +74,108 @@ def _sanitize_groups(self, group_names):
...
+        try:
+            description = f"Synclist repository for account {account_name}"
+
+            repository, created = AnsibleRepository.objects.get_or_create(name=name_slug)
+            repository.description = description
+            repository.save()
+            return repository
+        except IntegrityError as exc:
+            raise ValidationError("AnsibleRepository %s already exists: %s" % (name_slug, exc))
+        except Exception as exc:
```
cutwater, line 136: Why are you handling generic exception here for nothing than logging?
This exception must be handled by the top-level exception handler.

galaxy_ng/app/api/ui/serializers/synclist.py
```diff
@@ -45,23 +74,108 @@ def _sanitize_groups(self, group_names):
...
+            raise
+
+    def _update_distribution(self, name_slug, repository):
+        # Now the distro
+        try:
+            distribution, create = AnsibleDistribution.objects.get_or_create(
+                name=name_slug,
+                base_path=name_slug,
+            )
+        except IntegrityError as exc:
```
cutwater, line 147: IntegrityError will not be raised in case of duplicate records, because of `get_or_create`.

galaxy_ng/app/api/ui/serializers/synclist.py
```diff
@@ -45,23 +74,108 @@ def _sanitize_groups(self, group_names):
...
+            raise ValidationError("AnsibleDistribution %s already exists: %s" % (name_slug, exc))
+        except Exception as exc:
```
cutwater, line 149: ditto.

galaxy_ng/app/migrations/0003_synclist_upstream_and_migrate.py
```diff
@@ -0,0 +1,60 @@
...
+    # Nothing to update
+    if not synclists:
+        return
+
+    try:
+        default_repo = AnsibleRepository.objects.using(db_alias).get(name="automation-hub")
+    except AnsibleRepository.DoesNotExist:
+        # If there are synclists, but no default repo, then let the exception raise
+        if synclists:
+            raise
```
cutwater, line 24: Probably makes sense to raise more descriptive error rather than `DoesNotExist`, because with this check the error starts having different meaning.

galaxy_ng/app/models/auth.py
```diff
@@ -23,7 +30,45 @@ def create_identity(self, scope, name):
...
+
+    @staticmethod
+    def _create_repo(group, repository_name):
+        # TODO: the lifetime of the default repo is long, so shouldn't need to look it
+        #       up on every auth attempt. ie, cache it.
+        # TODO: This may be a reasonable place to hook up a django signal
+        try:
+            default_repo = AnsibleRepository.objects.get(name=repository_name)
+        except AnsibleRepository.DoesNotExist as exc:
+            log.exception(exc)
```
cutwater, line 54: Why are you catching an exception only to log it here?

galaxy_ng/app/models/auth.py
```diff
@@ -23,7 +30,45 @@ def create_identity(self, scope, name):
...
+
+    @staticmethod
+    def _create_distribution(group, repository):
+        distro_name = settings.GALAXY_API_SYNCLIST_NAME_FORMAT.format(
+            account_number=group.account_number())
+
+        try:
+            distro = AnsibleDistribution.objects.get(name=distro_name, base_path=distro_name)
+        except AnsibleDistribution.DoesNotExist as exc:
+            log.debug('distro_name: %s exc: %s', distro_name, exc)
```
cutwater, line 67: Probably makes sense to add operation name that is executed in addition to some local data.

galaxy_ng/tests/unit/api/test_api_ui_synclists.py
```diff
@@ -71,43 +40,43 @@ def test_synclist_create_as_user(self):
             response = self.client.post(self.synclists_url, post_data, format='json')
             self.assertEqual(response.status_code, http_code.HTTP_403_FORBIDDEN)
 
-    def test_synclist_create_as_pe_group(self):
-        repo = self._create_repository('test_post_repo')
-        repo.save()
+    # def test_synclist_create_as_pe_group(self):
```
cutwater, line 43: Please don't leave commented code in PRs.

chouseknecht:
This got rolled into #315. Closing, as it seems to be no longer needed.

Issue #316, opened 2020-07-23
Tasking: Auto-certify with pulp repos


awcrosby:
When `GALAXY_REQUIRE_CONTENT_APPROVAL=False`, change the auto-approval to use pulp repos instead of certification flag.

Related to #170

PR #317, opened 2020-07-23
Update import_and_auto_approve task to use pulp repos

awcrosby:
Closes-Issue: #316

* On GALAXY_REQUIRE_CONTENT_APPROVAL=False edit the auto approve task to
  use pulp repos to automatically promote content to golden repo

### DIFF
--- a/galaxy_ng/app/tasks/publishing.py
+++ b/galaxy_ng/app/tasks/publishing.py
@@ -12,8 +12,8 @@ log = logging.getLogger(__name__)
 
 VERSION_CERTIFIED = "certified"
 
-golden_name = settings.GALAXY_API_DEFAULT_DISTRIBUTION_BASE_PATH
-staging_name = settings.GALAXY_API_STAGING_DISTRIBUTION_BASE_PATH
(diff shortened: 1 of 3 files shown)


galaxy_ng/app/tasks/publishing.py
```diff
@@ -52,12 +52,27 @@ def import_and_auto_approve(artifact_pk, **kwargs):
     Custom task to call pulp_ansible's import_collection() task
     then automatically approve collection version so no
     manual approval action needs to occur.
-
-    Approval currently is a collection version certification flag
-    Approval later will be moving from a staging to an approved repo
     """
-    import_collection(artifact_pk=artifact_pk, repository_pk=kwargs.get('repository_pk', None))
+    inbound_repository_pk = kwargs.get('repository_pk')
```
alikins, line 56: I think the two tasks are pretty close to be able to be parameterized, especially in combo with https://github.com/ansible/galaxy_ng/pull/336/files

I think those two tasks could be one task with the inbound_repo as a kwarg if we wanted to.
At that point it is getting pretty close to pulpcore's add_or_remove task: https://github.com/pulp/pulpcore/blob/master/pulpcore/app/tasks/repository.py#L121

But let's get this pr and #336 merged and then refactor the tasks if needed.

Issue #330, opened 2020-07-28
Infra: Point source checkouts to Pulp_Ansible and Pulpcore Master branches


chouseknecht:
In order to merge in RBAC and Content Sync features, the latest changes found in [Pulpcore](https://github.com/pulp/pulpcore) and [Pulp_Ansible](https://github.com/pulp/pulp_ansible) are required. To solve this, we'll need to update our dev dependency requirements to point the master branch of each project. This will also require sorting out some bugs related to collection imports.

alikins:
Implemented in  https://github.com/ansible/galaxy_ng/pull/334 and merged to master at https://github.com/ansible/galaxy_ng/commit/325266ad06c04c14ed51b5cc5dfd19054ebe054d

PR #334, opened 2020-07-28
Set up galaxy to work with pulp@master.

alikins:
Issue: #330

### DIFF
--- a/requirements/requirements.common.txt
+++ b/requirements/requirements.common.txt
@@ -12,26 +12,23 @@ ansible==2.9.11           # via ansible-lint, galaxy-importer
 async-timeout==3.0.1      # via aiohttp
 attrs==19.3.0             # via aiohttp, galaxy-importer, jsonschema
 backoff==1.10.0           # via pulpcore
-bleach-whitelist==0.0.10  # via galaxy-importer
+bleach-whitelist==0.0.11  # via galaxy-importer
 bleach==3.1.5             # via galaxy-importer
 certifi==2020.6.20        # via galaxy-ng (setup.py), requests
 cffi==1.14.1              # via cryptography, pycares
 chardet==3.0.4            # via aiohttp, requests
 click==7.1.2              # via rq
-coreapi==2.3.3            # via drf-yasg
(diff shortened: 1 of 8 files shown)


requirements/requirements.common.txt
```diff
@@ -4,88 +4,89 @@
 #
 #    pip-compile --output-file=requirements/requirements.common.txt setup.py
 #
+git+https://github.com/pulp/pulp_ansible.git@master#egg=pulp_ansible  # via galaxy-ng (setup.py)
```
cutwater, line 7: This is wrong. `requirements.txt` is used to lock versions installed in docker containers. Installing from SCM must include a fixed reference (e.g. tag or commit SHA).

alikins: pip-compile really seems to want there to be a tag, but at the moment, there is no pulpcore 3.6.* tag. 
Spec'ing the latest hash from master gives warnings:

WARNING: Requested pulpcore>=3.0 from git+https://github.com/pulp/pulpcore.git@64b9fb994704691173258cd3f0fecfca70a2bff5#egg=pulpcore (from galaxy-ng (setup.py)), but installing version 3.6.0.dev0

setup.py
```diff
@@ -60,9 +60,11 @@ def run(self):
 
 requirements = galaxy_pulp_requirements + [
     "Django~=2.2.3",
-    "pulpcore>=3.4",
-    "pulp-ansible>=0.2.0b13",
+    "pulpcore",
```
newswangerd, line 63: For the RBAC work I'm doing I had to use this for my requirements

```
    "pulpcore @ git+https://github.com/pulp/pulpcore.git@master#egg=pulpcore",
    "pulp_ansible @ git+https://github.com/pulp/pulp_ansible.git@master#egg=pulp_ansible",
    "pulp_container @ git+https://github.com/pulp/pulp_container.git@master#egg=pulp_container",
```

requirements/requirements.common.txt
cutwater: That's likely because `pulpcore>=3.0` constraint does not include pre-release versions.

Issue #333, opened 2020-07-28
API/Importer: Update uses of pulp_ansible import_collection tasks to use PulpTemporaryFile


alikins:
Update galaxy_ng use the the pulp_ansible import_collection task to work with the pulpcore/pulp_ansible changes
related to:

https://pulp.plan.io/issues/6718
https://github.com/pulp/pulp_ansible/pull/335/files
https://github.com/pulp/pulpcore/pull/793
https://hackmd.io/@galaxyng/check-in-notes#Agenda1

Subtask of https://github.com/ansible/galaxy_ng/issues/330

alikins:
https://github.com/ansible/galaxy_ng/pull/336 is ready for review.

#336 updates the galaxy_ng tasks that use import_collection. Also updates the local dev container builds to use shared storage for /var/lib/pulp to enable PulpTemporaryFile

PR #336, opened 2020-07-29
Use PulpTemporaryFile

alikins:
Changes to use PulpTemporaryFile instead of Artifact as per:

https://pulp.plan.io/issues/6718
https://github.com/pulp/pulp_ansible/pull/335/files
https://github.com/pulp/pulpcore/pull/793/files

Issue: #330
Issue: #333

### DIFF
--- a/dev/docker-compose.yml
+++ b/dev/docker-compose.yml
@@ -29,9 +29,8 @@ services:
     volumes:
       - "./common/settings.py:/etc/pulp/settings.py:z"
       - "${COMPOSE_CONTEXT}/..:/src:z"
-      - "pulp_artifact:/var/lib/pulp/artifact"
+      - "pulp:/var/lib/pulp"
     tmpfs:
-      - "/var/lib/pulp/tmp:uid=1000,gid=1000"
       - "/tmp/ansible:uid=1000,gid=1000"
     depends_on:
       - _base
@@ -49,9 +48,8 @@ services:
     volumes:
       - "./common/settings.py:/etc/pulp/settings.py:z"
       - "${COMPOSE_CONTEXT}/..:/src:z"
-      - "pulp_artifact:/var/lib/pulp/artifact"
+      - "pulp:/var/lib/pulp"
     tmpfs:
-      - "/var/lib/pulp/tmp:uid=1000,gid=1000"
       - "/tmp/ansible:uid=1000,gid=1000"
     depends_on:
       - _base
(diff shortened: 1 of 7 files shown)


dev/docker-compose.yml
```diff
@@ -116,6 +112,7 @@ services:
 
 
 volumes:
+  pulp: {}
   pulp_artifact: {}
```
awcrosby, line 116: Is `pulp_artifact` still used / mapped above?

galaxy_ng/app/tasks/publishing.py
```diff
@@ -25,54 +42,67 @@ def import_and_move_to_staging(artifact_pk, **kwargs):
...
-    enqueue_with_reservation(add_content_to_repository, locks, args=task_args)
 
-    # enqueue task to remove collection_verion from inbound repo
     inbound_repo = AnsibleRepository.objects.get(pk=inbound_repository_pk)
-    locks = [inbound_repo]
-    task_args = (collection_version.pk, inbound_repository_pk)
-    enqueue_with_reservation(remove_content_from_repository, locks, args=task_args)
 
+    remove_locks = [staging_repo]
+    add_locks = [inbound_repo]
```
awcrosby, line 56: I think these two values need to be switched

galaxy_ng/app/api/v3/viewsets/collection.py
```diff
@@ -140,10 +140,11 @@ class CollectionUploadViewSet(LocalSettingsMixin, pulp_ansible_views.CollectionU
     ]
     parser_classes = [AnsibleGalaxy29MultiPartParser]
 
-    def _dispatch_import_collection_task(self, artifact_pk, repository=None, **kwargs):
+    def _dispatch_import_collection_task(self, temp_file_pk, repository=None, **kwargs):
         """Dispatch a pulp task started on upload of collection version."""
-        locks = [str(artifact_pk)]
-        kwargs["artifact_pk"] = artifact_pk
+        locks = [str(temp_file_pk)]
```
fao89, line 145: I believe temp_file does not need a lock, @bmbouter wdyt?

dev/docker-compose.yml
alikins: It is not used now, removing.

galaxy_ng/app/tasks/publishing.py
alikins: It certainly looks wrong now. Testing it again now to see if I switch it for a reason.

galaxy_ng/app/api/v3/viewsets/collection.py
alikins: pulp_ansible's impl of _dispatch_import_collection_task() doesn't seem to use.

That may be a left over from before I figured out the share /var/lib/pulp/ issues.
Seeing if I can remove it.

alikins: Doesn't seem to need a lock. Updated.

Issue #282, opened 2020-07-08
API: Content Sync Management


chouseknecht:
Subtask of #22 

Create API endpoints needed to configure and manage remotes for the `community` and `rh-certified` repositories.

We need to provide programmatic access for creating remotes. For this we can evaluate and potentially proxy pulp_ansible endpoints.

Since the UI is expecting a single endpoint for managing all repositories, this endpoint will be part of `_ui`.

chouseknecht:
For programmatic access, we can evaluate and potentially proxy pulp_ansible endpoints.

Since the UI needs a single endpoint for managing all repositories, it's OK to place this in `_ui`.

Issue #22, opened 2020-03-17
Story: Configure and manage content sync


chouseknecht:
As an admin of a private AH, I want to:

- Configure the remote source of content, including: URL, API token
- Initiate the content sync task
- Know the status of the sync task and view any logs
- ~~Set the provenance for the remote (e.g., local, Red Hat, Community)~~

chouseknecht:
Related to #17

chouseknecht:
We need to decide in which repository sync'd content will land. Our original assumption is that it would land in the 'golden' or default repository, and that's probably not what we want. Ideally we want the user to provide a repository name, and as part of the sync that repository gets created if it does not already exist.

chouseknecht:
Based on the discussion this morning (7 July) with John:

- [ ] only allow the sync'ing of content from community and c.rh.c. (Guessing we'll expand this in the future, but for now we agreed to limit it.)
- [ ] during installation, the installer will create the following default repositories:

    | Repository | Description |
    | ----------- | ----------- |
    | staging       | staging for private collection versions |
    | rejected      | rejected private collection versions |
    | published.  | certified or accepted private collection versions |
    | community | collections sync'd from galaxy.ansible.com
    | rh-certified | collections sync'd from c.rh.c |

- [ ] The UI will enable an admin user to configure the remote for the *community* repository and the remote for the *rh-certified* repository.
- [ ] For the *community* remote, the user will be required to provide a URL, token, and a string representing a requirements.yml file. 
- [ ] For the *rh-certified* remote, the user will be require to provide a URL and token. (Question: is an `auth_url` required? Believe the CLI requires this when authentication with c.rh.c)

alikins:
For the rh-certified remote, it may make sense to provide a cloud.redhat.com galaxy_ng api for the remote config info in one place. (for ex, possibly as part of the synclists api. Or on a my-repositories endpoint if we add that). 

Goal being only thing needed to setup sync of rh-certified from cloud.redhat.com would be login info, and c.r.c galaxy_ng could provide a list of the possible remotes for that user/org/account/

chouseknecht:
@rochacbruno @ZitaNemeckova 

Notes from the PM Sync meeting on 3 August:

- For the community remote, the `requirements.yml` should be an upload, not a copy-paste field
- In our documentation we can tell users how to syntax check their YAML before uploading it
- The backend is not expected to provide YAML syntax checking

PR #356, opened 2020-08-03
[WIP] Add endpoints and models for collection sync config and task

rochacbruno:
Related to #22
Closes-issue #282


dev_requirements.txt
```diff
@@ -8,3 +8,7 @@ flake8-docstrings
 
 flake8-tuple
 flake8-quotes
+
+django-extensions
```
newswangerd, line 12: What are these new requirements for?

galaxy_ng/app/api/ui/views/remotes.py
```diff
@@ -0,0 +1,45 @@
...
+
+TMP_MOCK = {
+    "meta": {"count": 2},
+    "links": {},
+    "data": [
+        {
+            "name": "rh-certified",
+            "url": "FQDN/automation-hub/v3/",
+            "auth_url": "FQDN/auth/realms/redhat-external/protocol/openid-connect/token",
+            "token": "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9...",
```
newswangerd, line 31: Lets make token write only when you get to the serializers.

dev_requirements.txt
rochacbruno: This is a dev only extension to provide the "django-admin shell_plus" an Ipython based django shell for debugging.

galaxy_ng/app/api/ui/serializers/collection.py
```diff
@@ -43,6 +43,24 @@ class CollectionVersionSummarySerializer(Serializer):
     created = serializers.CharField()
 
 
+class CollectionRemoteSerializer(Serializer):
+    name = serializers.CharField()
```
newswangerd, line 47: You should use a ModelSerializer. That way it will automatically handle some CRUD operations for you and you don't have to define the field types for each field. You should be able to import the [CollectionRemote](https://github.com/pulp/pulp_ansible/blob/master/pulp_ansible/app/models.py#L211) model from pulp ansible

Here's an example of a model serializer: https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/api/ui/serializers/namespace.py#L19

rochacbruno:
Replaced by #398

PR #375, opened 2020-08-11
Expose /repositories/ and /distributions/ api endpoints

alikins:


chouseknecht:
We took a different approach and no longer require these endpoints.

Issue #380, opened 2020-08-19
Infra: Point dev environment to latest Pulp packages


chouseknecht:
Now that the latest changes we need from Pulp are released, we can reverse the work done in #339

PR #381, opened 2020-08-19
Pinning to latest pulpcore, pulp-ansible and pulp-container

chouseknecht:
Now that all the Pulp features needed for 4.2.0b are released, we can pin to the latest packages. Reverses the work done in #330.

Closes-Issue: #380

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -60,9 +60,8 @@ galaxy_pulp_requirements = ["urllib3 >= 1.15", "six >= 1.10", "certifi", "python
 
 requirements = galaxy_pulp_requirements + [
     "Django~=2.2.3",
-    "pulpcore @ git+https://github.com/pulp/pulpcore.git@master#egg=pulpcore",
-    "pulp_ansible @ git+https://github.com/pulp/pulp_ansible.git@master#egg=pulp_ansible",
-    "pulp_container @ git+https://github.com/pulp/pulp_container.git@master#egg=pulp_container",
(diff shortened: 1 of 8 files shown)


setup.py
```diff
@@ -60,9 +60,9 @@ def run(self):
 
 requirements = galaxy_pulp_requirements + [
     "Django~=2.2.3",
-    "pulpcore @ git+https://github.com/pulp/pulpcore.git@master#egg=pulpcore",
-    "pulp_ansible @ git+https://github.com/pulp/pulp_ansible.git@master#egg=pulp_ansible",
-    "pulp_container @ git+https://github.com/pulp/pulp_container.git@master#egg=pulp_container",
+    "pulpcore==3.6.0",
```
cutwater, line 63: Please use loose version restrictions in `setup.py`. Pulp uses semantic versioning, so it must be safe to use Compatible Release Specifiers, like: 

```
pulpcore~=3.6.0
pulp-ansible~=0.2.0
pulp-container~=2.0.0
```

cutwater (changes requested):

setup.py
cutwater: pulp-container is not a direct dependency of `galaxy_ng`. It must not be a part of `setup.py`.

chouseknecht: Fixed.

cutwater (approved):

PR #399, opened 2020-08-25
Pin pulpcore and update travis

fao89:
No-Issue

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -63,6 +63,7 @@ requirements = galaxy_pulp_requirements + [
     "pulpcore~=3.5.0",
     "pulp-ansible==0.2.0b15",
     "django-prometheus>=2.0.0",
+    "galaxy-importer",
     "drf-spectacular",
 ]
 
--- a/template_config.yml
+++ b/template_config.yml
@@ -24,7 +24,7 @@ plugin_name: galaxy_ng
 plugin_snake: galaxy_ng
 pulp_settings: null
 pulpcore_branch: master
-pulpcore_pip_version_specifier: null
+pulpcore_pip_version_specifier: ~=3.5.0
 pydocstyle: true
 pypi_username: null
 redmine_project: null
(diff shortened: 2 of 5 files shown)


template_config.yml
```diff
@@ -24,7 +24,7 @@ plugin_name: galaxy_ng
 plugin_snake: galaxy_ng
 pulp_settings: null
 pulpcore_branch: master
-pulpcore_pip_version_specifier: null
+pulpcore_pip_version_specifier: ~=3.5.0
```
fao89, line 27: I specified the pulpcore requirement here, then:
```
cd plugin_template
./plugin-template --travis galaxy_ng
```

setup.py
```diff
@@ -63,6 +63,7 @@ def run(self):
     "pulpcore~=3.5.0",
     "pulp-ansible==0.2.0b15",
     "django-prometheus>=2.0.0",
+    "galaxy-importer",
```
fao89, line 66:
```
- jsonschema [required: >=2.6.0, installed: 3.2.0]
            - attrs [required: >=17.4.0, installed: 20.1.0]
- aiohttp [required: Any, installed: 3.6.2]
              - attrs [required: >=17.3.0, installed: 20.1.0]
- galaxy-importer==0.2.7
       - attrs [required: >=19.3.0,<20, installed: 20.1.0]
```
https://travis-ci.com/github/fao89/galaxy_ng/builds/181352373#L822
I don't know why, but `attrs-20.1.0` was being installed for `aiohttp`, and it was not installing a `galaxy-importer` compatible version

awcrosby: @fao89 Looks like CI is pulling a different version of `galaxy-importer` (`0.2.7` release) than we have specified in requirements (`0.2.8rc12` pre-release) - should that be expected when we have a pre-release?

fao89: @awcrosby as far as I know the CI is getting only the dependencies listed on `setup.py`. I see multiple requirements files, but I believe the CI is ignoring them. I think this is the real issue here.

Issue #372, opened 2020-08-10
API: RBAC permissions for publishing content


chouseknecht:
- [x] Disallow publishing collections to the golden
- [x] Extend namespace owner permissions to the inbound repository via name association

chouseknecht:
Subtask of #20

PR #406, opened 2020-08-26
Add RBAC permisisons for publishing content

awcrosby:
Closes-Issue: #372

### DIFF
--- a/CHANGES/372.feature
+++ b/CHANGES/372.feature
@@ -1,0 +1,1 @@
+On publish check if inbound repo allows publishing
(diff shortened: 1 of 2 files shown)


galaxy_ng/app/access_control/access_policy.py
```diff
@@ -40,6 +44,21 @@ def can_update_collection(self, request, view, permission):
         namespace = models.Namespace.objects.get(name=collection.namespace)
         return request.user.has_perm('galaxy.upload_to_namespace', namespace)
 
+    def check_inbound_repo_format_and_match_namespace(self, request, view, permission):
+        """Reject if inbound repo is not the expected format
+           or inbound repo does not match namespace in filename."""
+        path = request.parser_context['kwargs']['path']
+        distro = get_object_or_404(AnsibleDistribution, base_path=path)
+        repo_name = distro.repository.name
+        try:
```
newswangerd, line 53: Is there a way to move this to the viewset? We shouldn't be validating data in the permission checks.

awcrosby: @newswangerd maybe to address this issue, no rbac edits are needed at all... I pushed a commit with all checks in the view, see how that looks.

PR #407, opened 2020-08-27
Replace pulpcore branch from master to 3.5.0

fao89:
@cutwater I noticed we already had a config for the branch.


newswangerd:
Shouldn't this be puplcore 3.6.0?

fao89:
> Shouldn't this be puplcore 3.6.0?

I'm following `setup.py`:
https://github.com/ansible/galaxy_ng/blob/stable/setup.py#L63

PR #409, opened 2020-08-28
Add api endpoint for starting collection syncs.

newswangerd:
Add a `/api/automation-hub/content/<distribution>/v3/sync/` endpoint that allows users to start syncs on the collection remote attached to the specified distribution.

Syncs are kicked off with a `POST` request and return the ID of the task.

Response
```json
{
    "task": "3024deed-4df2-44c7-a3f1-a38124c6ccf7"
}
```

PR #410, opened 2020-08-28
Add endpoints for sync/config management

rochacbruno:
- endpoint: `/content/<distribution>/v3/sync/config/`
- methods: GET/PUT

- endpoint: `/content/<distribution>/v3/sync/`  (taken from #409) 
- methods: POST

Issue: #282

- [x] Add endpoints for `sync/config` GET and PUT
- [x] Add endpoint for `sync/` POST
- [x] Validate PUT data assuring `rh-certified` repository has a `null` `requirement-file` and the opposite for `community` repository
- [x] Validate PUT data assuring that `name` is a `VALID_REMOTE_NAME`
- [x] Add access policies and permissions
- [x] Write unit tests

### DIFF
--- a/galaxy_ng/app/api/ui/viewsets/collection.py
+++ b/galaxy_ng/app/api/ui/viewsets/collection.py
@@ -14,6 +14,7 @@ from rest_framework.exceptions import NotFound
 from rest_framework.generics import get_object_or_404
 from rest_framework.response import Response
 
+from galaxy_ng.app.constants import VALID_DISTRO_NAMES
 from galaxy_ng.app import models
 from galaxy_ng.app.api import base as api_base
 from galaxy_ng.app.access_control import access_policy
@@ -314,7 +315,7 @@ class CollectionImportViewSet(api_base.GenericViewSet):
 
 
 class CollectionRemoteViewSet(api_base.ModelViewSet):
-    queryset = CollectionRemote.objects.filter(name__in=['rh-certified', 'community'])
+    queryset = CollectionRemote.objects.filter(name__in=VALID_DISTRO_NAMES)
     serializer_class = serializers.CollectionRemoteSerializer
 
     permission_classes = [access_policy.CollectionRemoteAccessPolicy]
--- a/galaxy_ng/app/api/ui/viewsets/collection.py
+++ b/galaxy_ng/app/api/ui/viewsets/collection.py
@@ -14,7 +14,6 @@ from rest_framework.exceptions import NotFound
 from rest_framework.generics import get_object_or_404
 from rest_framework.response import Response
 
-from galaxy_ng.app.constants import VALID_DISTRO_NAMES
 from galaxy_ng.app import models
 from galaxy_ng.app.api import base as api_base
 from galaxy_ng.app.access_control import access_policy
@@ -315,7 +314,7 @@ class CollectionImportViewSet(api_base.GenericViewSet):
 
 
 class CollectionRemoteViewSet(api_base.ModelViewSet):
-    queryset = CollectionRemote.objects.filter(name__in=VALID_DISTRO_NAMES)
+    queryset = CollectionRemote.objects.all()
     serializer_class = serializers.CollectionRemoteSerializer
 
     permission_classes = [access_policy.CollectionRemoteAccessPolicy]
--- a/galaxy_ng/app/api/v3/serializers/sync.py
+++ b/galaxy_ng/app/api/v3/serializers/sync.py
@@ -1,12 +1,18 @@
 from rest_framework import serializers
 from pulp_ansible.app.models import CollectionRemote
 from pulp_ansible.app.viewsets import CollectionRemoteSerializer
-from galaxy_ng.app.constants import COMMUNITY_DOMAINS, VALID_DISTRO_NAMES
+from galaxy_ng.app.constants import COMMUNITY_DOMAINS
 
 
 class SyncConfigSerializer(CollectionRemoteSerializer):
     created_at = serializers.DateTimeField(source='pulp_created', required=False)
     updated_at = serializers.DateTimeField(source='pulp_last_updated', required=False)
+    policy = serializers.ChoiceField(
+        choices=CollectionRemote.POLICY_CHOICES,
+        default=CollectionRemote.ON_DEMAND
+    )
+    token = serializers.CharField(allow_null=True, required=False, max_length=2000, write_only=True)
+    name = serializers.CharField(read_only=True)
 
     class Meta:
         model = CollectionRemote
@@ -14,6 +20,7 @@ class SyncConfigSerializer(CollectionRemoteSerializer):
             'name',
             'url',
             'auth_url',
+            'token',
             'policy',
             'requirements_file',
             'created_at',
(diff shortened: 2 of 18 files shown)


galaxy_ng/app/api/v3/viewsets/sync.py
```diff
@@ -0,0 +1,22 @@
...
+    mixins.UpdateModelMixin,
+    api_base.GenericViewSet,
+):
+    serializer_class = SyncConfigSerializer
+    permission_classes = [access_policy.CollectionRemoteAccessPolicy]
+
+    def get_object(self):
+        return get_object_or_404(
+            CollectionRemote,
+            basedistribution__name=self.kwargs['path']
```
rochacbruno, line 21: @newswangerd I am not 100% sure about this query, what do you think? Is it possible to have multiple CollectionRemote pointing to same basedistribution? if so it will raise a `Multiple Results Error`

newswangerd: there's a one to many relation ship between distributions and repositories and another one to many relationship between repos and remotes, so that means that for every distrbution there should be no more than one repository associated with it and no more than one remote associated with it, so this should work

galaxy_ng/app/api/v3/serializers/sync.py
```diff
@@ -0,0 +1,45 @@
...
+        errors = {}
+        name = data['name']
+        requirements_file = data['requirements_file']
+
+        if name not in VALID_REMOTE_NAMES:
+            errors['name'] = (
+                f'{name} is not a valid remote name, possible values are {VALID_REMOTE_NAMES}'
+            )
+
+        if name == 'rh-certified' and requirements_file is not None:
```
newswangerd, line 32: I don't think we need to enforce no requirements file for red hat certified. We won't offer it as an option in the UI, but if someone wants to back door the API and put one in, we don't have to stop them

galaxy_ng/app/api/v3/serializers/sync.py
```diff
@@ -0,0 +1,45 @@
...
+            errors['requirements_file'] = (
+                'The field requirements_file must be null for rh-certified repository'
+            )
+
+        if name == 'community' and requirements_file is None:
```
newswangerd, line 37: Instead of checking the name of the remote, lets check that the URL contains `api/v2`, since that's the version that's supported on Galaxy.

galaxy_ng/app/api/v3/urls.py
```diff
@@ -16,6 +16,18 @@
     path("auth/token/", views.TokenView.as_view(), name="auth-token"),
 ]
 
+sync_urls = [
```
newswangerd, line 19: please add a note saying that these are included in the base router so that they only appear under `content/<distro>/v3` and not `v3/`

galaxy_ng/app/api/v3/serializers/sync.py
rochacbruno: OK, I will change that. 

This is what I read on doc: 
```
requirements_file (string) – Explicit list of requirements to synchronize (Must be null for rh-certified repository and non-empty for community repository).
```

galaxy_ng/app/api/ui/viewsets/collection.py
```diff
@@ -314,7 +315,7 @@ def retrieve(self, request, *args, **kwargs):
 
 
 class CollectionRemoteViewSet(api_base.ModelViewSet):
-    queryset = CollectionRemote.objects.filter(name__in=['rh-certified', 'community'])
```
newswangerd, line 317: I think it might be a better idea to remove the filter here. We don't allow people to add new remotes, so this hopefully won't be needed.

galaxy_ng/app/api/v3/serializers/sync.py
```diff
@@ -0,0 +1,48 @@
...
+    class Meta:
+        model = CollectionRemote
+        fields = (
+            'name',
+            'url',
+            'auth_url',
+            'policy',
+            'requirements_file',
+            'created_at',
+            'updated_at',
```
newswangerd, line 24: We still need to return the token, but as a write only field.

Here's an example showing how to do this: https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/api/ui/serializers/user.py#L35

galaxy_ng/app/api/v3/serializers/sync.py
```diff
@@ -0,0 +1,48 @@
...
+        )
+
+    def validate(self, data):
+        data = super().validate(data)
+        errors = {}
+        name = data['name']
+        requirements_file = data['requirements_file']
+        url = data['url']
+
+        if name not in VALID_DISTRO_NAMES:
```
newswangerd, line 34: remove this check and just make the name read only.

https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/api/v3/serializers/namespace.py#L101

galaxy_ng/app/api/v3/serializers/sync.py
```diff
@@ -0,0 +1,48 @@
...
+        if not requirements_file and any([domain in url for domain in COMMUNITY_DOMAINS]):
+            errors['requirements_file'] = (
+                'Syncing content from community domains without specifying a '
+                'requirements file is not allowed.'
+            )
+
+        if errors:
+            raise serializers.ValidationError(detail=errors)
+
+        return data
```
newswangerd, line 48: `return super().validate(data)`

newswangerd (changes requested):

galaxy_ng/app/api/v3/serializers/sync.py
rochacbruno: I had to re-define the fields adding `write_only=True` because `extra_vars` doesn't work when  we are overriding the `.validate()` method.

newswangerd (approved):

galaxy_ng/app/api/v3/serializers/sync.py
```diff
@@ -0,0 +1,41 @@
...
+from pulp_ansible.app.models import CollectionRemote
+from pulp_ansible.app.viewsets import CollectionRemoteSerializer
+from galaxy_ng.app.constants import COMMUNITY_DOMAINS
+
+
+class SyncConfigSerializer(CollectionRemoteSerializer):
+    created_at = serializers.DateTimeField(source='pulp_created', required=False)
+    updated_at = serializers.DateTimeField(source='pulp_last_updated', required=False)
+    policy = serializers.ChoiceField(
+        choices=CollectionRemote.POLICY_CHOICES,
```
bmbouter, line 11: This will allow policies other than immediate to be used, but since our convo @rochacbruno I've learned that pulp-ansible cannot support this. We will need to resolve this issue first https://pulp.plan.io/issues/7470

As such I think this specific aspect needs to be reverted (and just use the version defined on the parent class `CollectionRemoteSerializer`.

@newswangerd fyi also ^

rochacbruno: Ok  @bmbouter thanks for the info, I will remove that by now and keep with the super class implementation.

Issue #428, opened 2020-09-10
Initiate curate task on collection move


chouseknecht:
The task for [curating content](https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/tasks/synclist.py) needs to be initiated whenever a new collection lands in the golden repository.

The place where this task should be initiated has been identified here: 

https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/api/v3/viewsets/collection.py#L268

chouseknecht:
https://gist.github.com/awcrosby/f412811feaca512c67586a12f9eefd3b

chouseknecht:
Example implementation from @awcrosby: https://gist.github.com/awcrosby/f412811feaca512c67586a12f9eefd3b

alikins:
@chouse @newswangerd And we want this on a /v3/ endpoint, not a /ui/ endpoint correct?

alikins:
@chouse @newswangerd Assuming v3 endpoint, are there any additional RBAC requirements to account for?

For example, would it need to support letting user A move content to golden-repo, but not allow user A to trigger a curation of all the synclist repos? More importantly, does it need to do that _now_?. 

Can I assume that level of RBAC granularity isn't needed this sprint?

chouseknecht:
I think based on the discussion in slack that we don't need to add RBAC to this. If a user has permission to invoke the move_content method, then they automatically have permission to invoke the curate task.

alikins:
See also https://github.com/ansible/galaxy_ng/issues/285

alikins:
One implementation is in https://github.com/ansible/galaxy_ng/pull/432

hendersonreed:
What's the external representation of this task being kicked off? I see that in the response to the move-content response there's a `curate_task.id` returned. Is that going to be the primary way of determining that this task is running, or is there something else the tests should be evaluating as well?

newswangerd:
Moved to issues.redhat.com/

PR #432, opened 2020-09-10
On changes to golden repo, trigger all synclist tasks

alikins:
In the CollectionMoveViewSet, if either of the repositories
involved is the 'golden' repo, then enqueue a
'curate_all_synclist_repository' task that start
tasks that curate all synclist repositories based on
the synclist associated with it.

ie, when the main repo changes, kick off tasks to update
all the repos based on it.

Issue: #428

### DIFF
--- a/galaxy_ng/app/api/v3/viewsets/collection.py
+++ b/galaxy_ng/app/api/v3/viewsets/collection.py
@@ -19,12 +19,13 @@ from pulpcore.plugin.models import ContentArtifact, Task
 from pulpcore.plugin.tasking import enqueue_with_reservation
 from pulp_ansible.app.galaxy.v3 import views as pulp_ansible_views
 from pulp_ansible.app.models import CollectionVersion, AnsibleDistribution
-
 from galaxy_ng.app.api.base import (
     APIView,
     ViewSet,
     LocalSettingsMixin,
 )
+
+from galaxy_ng.app.constants import DeploymentMode
 
 from galaxy_ng.app.api.v3.viewsets.namespace import INBOUND_REPO_NAME_FORMAT
 from galaxy_ng.app import models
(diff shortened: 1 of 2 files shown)


galaxy_ng/app/api/v3/viewsets/collection.py
```diff
@@ -301,10 +303,26 @@ def move_content(self, request, *args, **kwargs):
...
         remove_task = self._remove_content(collection_version, src_repo)
 
+        if settings.GALAXY_DEPLOYMENT_MODE == DeploymentMode.INSIGHTS.value:
+            golden_repo = AnsibleDistribution.objects.get(
+                base_path=settings.GALAXY_API_DEFAULT_DISTRIBUTION_BASE_PATH
+            ).repository
+
+            if dest_repo == golden_repo or src_repo == golden_repo:
+                repo_name = golden_repo.name
+                locks = [repo_name]
```
awcrosby, line 313: The other `locks` in this module use the repository object rather than name... following this pulp_ansible pattern: https://github.com/pulp/pulp_ansible/blob/master/pulp_ansible/app/galaxy/mixins.py#L18

On a side note I see https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/tasks/synclist.py#L72 uses str of an object id... I had assumed `locks` to require an object, or if id was accepted it would need to be a pulp id since a pulp id is resolvable across models, but I'm not sure.

alikins: I agree, it's odd. I'm not sure really which is correct. I've switched it back and fourth a few times. 

At one point the advice given to me was that locks only needed to be strings uniq to an object (ie, more or less a repr),
and that using objects may cause deadlocks.

But looking at more pulpcore code, a lot of it uses objects.  And the task results 'reserved resources' stuff certainly makes more sense when the locks are objects.

alikins: Just updated it to use the Repository itself as the 'lock' / reserved resource. Seems to work fine.

awcrosby (approved):

PR #436, opened 2020-09-11
Exclude synclist and inbound repos from list of repos on collection versions API.

newswangerd:


alikins (approved):
Seems okay for now. 

In the future can probably be more explicit about the filtering later (ie, filter by rbac/guardian/etc based query filters to exclude incoming and synclist for everyone except current user)

awcrosby (approved):

PR #462, opened 2020-09-17
TLS: Allow one to disable certificate verification from settings file

Spredzy:
This might be me not knowing where to look for it. But I was looking to disable cert validation by enabling a toggle on the settings file and I couldn't find it. This commit allows me to do that.

If there is another way to currently do it please let me know and feel free to close this PR.

Issue: #464


cutwater:
I am not sure about this one. It modifies the auto-generated code which we don't update frequently, but it still can be overridden by openapi schema generator.

Spredzy:
@cutwater where would be the best place in the code source to implement such a behavior ?

cutwater (approved):
LGTM

Issue #476, opened 2020-09-22
Upgrade to Pulp 3.7.0


chouseknecht:
- Upgrade to pulpcore 3.7.0 and allow for 3.8.0 (see https://docs.pulpproject.org/pulpcore/restapi.html#tag/Status)
- Upgrade pulp_container to compatibility release for pulpcore 3.7.0
- Upgrade pulp_ansible to compatibility release for pulpcore 3.7.0
- Regenerate Travis CI template after the above are completed

newswangerd:
Related: https://github.com/ansible/galaxy_ng/issues/474

alikins:
Pulp api status docs: 
https://docs.pulpproject.org/pulpcore/plugins/plugin-writer/concepts/index.html#plugin-api-stability-and-deprecation-policy
(source https://github.com/pulp/pulpcore/blob/master/docs/plugins/plugin-writer/concepts/index.rst#plugin-api-stability-and-deprecation-policy)

chouseknecht:
@alikins 

From the #pulp-dev irc channel:  pulpcore 3.7.0, pulp_ansible 0.4.0 and pulp_container

bmbouter:
pulp_container compatibility version will be 2.1.0

alikins:
Fix committed via pr #483

bmbouter:
pulpcore 3.7.0, pulp_ansible 0.4.0 and pulp_container 2.1.0 are all available on pypi

bpereto:
pulpcore 3.9.0 is out.

newswangerd:
Moved to issues.redhat.com/

PR #483, opened 2020-09-23
Update pulpcore to >=3.7,<3.9, pulp_ansible~=0.4.0  #476

alikins:
The ANSIBLE_SKIP_CONFLICT_CHECK additions to
the pip-compile lines are because otherwise
pip-compile fails with error:

  Upgrading directly from ansible-2.9 or less to ansible-2.10 or greater with pip is
  known to cause problems.  Please uninstall the old version found at:

  /home/adrian/venvs/galaxy_ng_2/lib64/python3.6/site-packages/ansible/__init__.py

  and install the new version:

      pip uninstall ansible
      pip install ansible

  If you have a broken installation, perhaps because ansible-base was installed before
  ansible was upgraded, try this to resolve it:

      pip install --force-reinstall ansible ansible-base

  If ansible is installed in a different location than you will be installing it now
  (for example, if the old version is installed by a system package manager to
  /usr/lib/python3.8/site-packages/ansible but you are installing the new version into
  ~/.local/lib/python3.8/site-packages/ansible with `pip install --user ansible`)
  or you want to install anyways and cleanup any breakage afterwards, then you may set
  the ANSIBLE_SKIP_CONFLICT_CHECK environment variable to ignore this check:

      ANSIBLE_SKIP_CONFLICT_CHECK=1 pip install --user ansible

Issue: #476

### DIFF
--- a/.travis/before_install.sh
+++ b/.travis/before_install.sh
@@ -101,7 +101,7 @@ if [ -z "$TRAVIS_TAG" ]; then
 fi
 
 
-git clone --depth=1 https://github.com/pulp/pulp_ansible.git --branch 0.3.0
+git clone --depth=1 https://github.com/pulp/pulp_ansible.git --branch 0.4.0
 if [ -n "$PULP_ANSIBLE_PR_NUMBER" ]; then
   cd pulp_ansible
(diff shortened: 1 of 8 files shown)


chouseknecht (approved):
Tested the dev environment build. LGTM.

Issue #474, opened 2020-09-21
FIPS compliance settings


newswangerd:
There's a new setting in pulp 3.7.0 that allows users to set a list of allowed checksums: https://github.com/pulp/pulpcore/blob/master/docs/settings.rst#allowed_content_checksums

To be FIPS compliant we need to set the allowed checksums to

```
{"sha224", "sha256", "sha384", "sha512"}
```

and test galaxy_ng on a FIPS compliant os to ensure that it works.

chouseknecht:
Need pulp 3.7.0 release and a release of pulp_ansible that supports 3.7.0

bmbouter:
For pulpcore 3.7.0 FIPS environments, I recommend configuring the [ALLOWED_CONTENT_CHECKSUMS](https://docs.pulpproject.org/pulpcore/en/3.7/settings.html#allowed-content-checksums) setting with these values:

`ALLOWED_CONTENT_CHECKSUMS = ["sha224", "sha256", "sha384", "sha512"]`

If you use the pulp_ansible installer version 3.7.0 it will also ensure all of your dependencies are also FIPS compatible at install time. For rpm builds, we've sent over the only necessary patch (for Django) to the build team building your RPMs here: https://github.com/theforeman/pulpcore-packaging/pull/21/files

For an environment to test FIPS in, pulplift provides both an EL7-fips and EL8-fips environments. If you want to enable it in another environment, e.g. EL7 [here are some docs](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/security_guide/chap-federal_standards_and_regulations) on how to use the kernel param.

daviddavis:
@bmbouter I had the thought yesterday that galaxy_ng could just always set `ALLOWED_CONTENT_CHECKSUMS = ["sha224", "sha256", "sha384", "sha512"]` regardless of whether it's in FIPS mode or not. Would that work?

bmbouter:
> @bmbouter I had the thought yesterday that galaxy_ng could just always set `ALLOWED_CONTENT_CHECKSUMS = ["sha224", "sha256", "sha384", "sha512"]` regardless of whether it's in FIPS mode or not. Would that work?

Yes I believe this setting would be safe to apply in all environments.

newswangerd:
Moved to issues.redhat.com/

PR #486, opened 2020-09-24
Add ALLOWED_CONTENT_CHECKSUMS to support FIPS compliance.

newswangerd:
Issue: #474


bmbouter:
I don't understand how pulpcore is shipping a settings.py and galaxy_ng is also shipping a settings.py. The line I see here though looks right.

cutwater (approved):

bmbouter:
> I don't understand how pulpcore is shipping a settings.py and galaxy_ng is also shipping a settings.py. The line I see here though looks right.

Actually this is perfectly correct. I didn't recognize this as a "plugin shipped settings file", which it is.

PR #491, opened 2020-09-25
Fix OpenAPI spec auth url

hendersonreed:
* Fix authentication url
* Change server url to match the dev environment.

## Details

The OpenAPI spec had an incorrect url for authentication, as it hadn't been updated since the URL was changed from `/auth/token` to `/v3/auth/token`.

~~I've also changed the server url so that it properly matches the dev environment, which starts the server on port 8002 rather than 5001.~~ EDIT 3: This change resulted due to my misunderstanding of the containerization setup, and I've revised the PR to omit it.

EDIT: you can view my changes at https://petstore.swagger.io/?url=https://raw.githubusercontent.com/hendersonreed/galaxy_ng/fix-openapi-spec-auth-url/openapi/openapi.yaml#/

EDIT 2: pasted the wrong url in my first edit, so I fixed that.

### DIFF
--- a/openapi/openapi.yaml
+++ b/openapi/openapi.yaml
@@ -91,7 +91,7 @@ paths:
   # Authentication
   # -------------------------------------
 
-  /auth/token/:
+  /v3/auth/token/:
     post:
       summary: "Issue Auth Token"
       operationId: createAuthToken


alikins:
Looks correct to me. 

But not sure about

> "I've also changed the server url so that it properly matches the dev environment, which starts the server on port 8002 rather than 5001."

At least with the `./compose` stuff, the api stuff is always available at 5001, but only available at 8002 if the ansible-hub-ui server is running as well.

hendersonreed:
ah, that's a good point! I wasn't aware of that detail - I think I agree with you, so I'll cut that out.

hendersonreed:
What about adding an additional server? One for when you are running the UI, and one for when you aren't.

newswangerd (approved):

Issue #496, opened 2020-09-28
Original V3 Upload endpoint broke backwards compat when inbound endpoints added


ironfroggy:
##### SUMMARY
Basically, that original upload endpoint we used prior to the repository-based certifications and adding our inbound upload endpoints is now completely broken. It is now trying to upload to the "default" repository which is never allowed because you can only upload now to the namespace found in the artifact filename and metadata.

##### STEPS TO REPRODUCE
Try to upload any artifact using a previously valid single URL configuration.

##### EXPECTED RESULTS
I expect an API endpoint that was previously implemented and published and used by released versions of `ansible-galaxy` to continue to work. I do not expect that internal changes to our database (repositories vs booleans for tracking certifications) to change the external endpoints I, as a user, interact with.

##### ACTUAL RESULTS

The upload fails with an unfriendly 500 error and the following response body:

```
{"errors":[{"status":"500","code":"error","title":"A server error occurred.","detail":"Path does not match: \\"inbound-autohubtest2\\""}]}
```

Internally, this is happening because the URL for the original upload endpoint now checks that the artifact is being uploaded to the correct repository matching the namespace in the artifact filename, but that URL without an explicit path defaults to the default repo (usually `"automation-hub"`) instead of directing that artifact to the correct repository based on said namespace extracted from the filename.

thedoubl3j:
Over the past few weeks, we (PE) have encountered multiple instances where some of the basic things for developing collections were missed. I know that collection development doesn't have a direct tie to this issue but it shows that the path people get to collections and certification are not strictly through the docs. 

I agree with what the expected results would be and hope we can continue that functionality. Introducing a change that could potentially break release pipelines and change upload understanding might not be the best for our maintainers at the current time. For the long run, the easiest path (read least changed) would probably be the best as it provides consistency for every user's experience.

chouseknecht:
As part of the RC release we will add logic that maintains the backward compatibility for publishing to the existing endpoint. This applies to both 'standalone' and 'insights' mode. When an artifact is published to the existing endpoint without a namespace, the underlying view will inspect the artifact to determine the namespace, check user permissions, and publish the artifact to the correct inbound repository.

ironfroggy:
Works great 👍

newswangerd:
Moved to issues.redhat.com/

PR #500, opened 2020-09-29
Use filename namespace on upload when no distro specified

awcrosby:
Closes-Issue: #496

Uploads with no distro specified will use `inbound-<namespace>` distro based on filename namespace


alikins (approved):
Looks good to me. Looks like it does a reasonable thing when the
old implicit API is used.

PR #502, opened 2020-09-30
Better error for invalid filename

### LINKED ISSUE #31: Artifact Upload: Invalid filename errors do not include an error code
```json
{
  "status": "400",
  "code": null,
  "title": "Invalid input.",
  "detail": "Invalid version string name from filename a-wrong-name.tar.gz. Expected semantic version format.",
  "source": {"parameter": "filename"}
}
```

When given an invalid filename, the error response body does not include a `code` as dictated by the standard error format. The same error on Galaxy gives the error code `invalid` so it probably makes sense to be consistent here.

alikins:
The ValidationError was getting raised with a 'details'
dict that didn't include a 'code', so it was unset.

Pass in a code to use _get_error_details() to generate
the 'detail' dict correctly with a valid 'invalid'.

Fixes #31

### DIFF
--- a/CHANGES/31.bugfix
+++ b/CHANGES/31.bugfix
@@ -1,0 +1,1 @@
+Make error return for upload filename parsing errors provides an error code 'invalid'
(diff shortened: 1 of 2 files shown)


alikins:
@ironfroggy @hendersonreed  this is a mostly untested fix for #31   YMMV

awcrosby (approved):

PR #514, opened 2020-10-12
Remove ui api non-repo collections endpoint

awcrosby:
This endpoint is no longer used by the UI and was replaced with per-repo collections logic.

### DIFF
--- a/galaxy_ng/app/api/ui/viewsets/collection.py
+++ b/galaxy_ng/app/api/ui/viewsets/collection.py
@@ -23,15 +23,15 @@ from galaxy_ng.app.api.ui import serializers, versioning
 from galaxy_ng.app.common import pulp
 
 
-class RepositoryCollectionFilter(CollectionVersionFilter):
+class CollectionFilter(CollectionVersionFilter):
     versioning_class = versioning.UIVersioning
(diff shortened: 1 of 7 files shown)


newswangerd (changes requested):
This looks good, but it shouldn't be merged until after the next cloud release happens.

awcrosby:
@newswangerd I added back in the _ui/collections endpoint as deprecated and changed it to use the `published` repo so it can be removed after the UI deployment.

galaxy_ng/app/api/ui/viewsets/collection.py
```diff
@@ -168,9 +54,19 @@ def get_queryset(self):
...
     def get_serializer_class(self):
         if self.action == 'list':
-            return serializers.RepositoryCollectionListSerializer
+            return serializers.CollectionListSerializer
         else:
-            return serializers.RepositoryCollectionDetailSerializer
+            return serializers.CollectionDetailSerializer
+
+
+# TODO: Remove when ui is updated and no longer uses this endpoint
```
alikins, line 62: This is related to the cloud deploy stuff discussed yesterday?  ie, needing to have both versions of this endpoint running until there is no ui running that needs it?

newswangerd (approved):

PR #522, opened 2020-10-22
Add an endpoint to return a CollectionVersion's resolved dependencies

bmclaughlin:
Issue: AAH-8

### DIFF
--- a/galaxy_ng/app/api/v3/serializers/collection.py
+++ b/galaxy_ng/app/api/v3/serializers/collection.py
@@ -17,6 +17,8 @@ from pulp_ansible.app.galaxy.v3.serializers import (
     CollectionVersionListSerializer as _CollectionVersionListSerializer,
 )
 
+from pulp_ansible.app.models import CollectionVersion
+
 log = logging.getLogger(__name__)
 
 
@@ -116,6 +118,27 @@ class CollectionVersionSerializer(_CollectionVersionSerializer, HrefNamespaceMix
         kwargs = {"path": self.context["path"], "namespace": obj.namespace,
                   "name": obj.name, "version": obj.version}
         return self._get_href("collection-versions-detail", **kwargs)
+
+
+class CollectionVersionDependencySerializer(_CollectionSerializer):
+    dependencies = serializers.JSONField()
+
+    class Meta:
+        fields = (
+            "dependencies",
+        )
+        model = CollectionVersion
+
+    def get_href(self, obj):
+        """Get href."""
+        kwargs = {
+            "path": self.context["path"],
+            "namespace": obj.namespace,
+            "name": obj.name,
+            "version": obj.version,
+        }
+
+        return self._get_href("collection-version-dependency", **kwargs)
 
 
 class CollectionUploadSerializer(Serializer):
--- a/galaxy_ng/app/api/v3/serializers/collection.py
+++ b/galaxy_ng/app/api/v3/serializers/collection.py
@@ -125,6 +125,7 @@ class CollectionVersionDependencySerializer(_CollectionSerializer):
 
     class Meta:
         fields = (
+            "version",
             "dependencies",
         )
         model = CollectionVersion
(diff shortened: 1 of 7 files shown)


bmbouter:
Generally the entire Galaxy API V3 is put into pulp_ansible and then subclassed back to galaxy_ng. Have we considered where the right place to contribute this is?

galaxy_ng/app/api/v3/serializers/collection.py
```diff
@@ -17,6 +17,8 @@
     CollectionVersionListSerializer as _CollectionVersionListSerializer,
 )
 
+from pulp_ansible.app.models import CollectionVersion
```
cutwater, line 20: PEP8: 
```
Imports should be grouped in the following order:

Standard library imports.
Related third party imports.
Local application/library specific imports.
```

galaxy_ng/app/api/v3/viewsets/collection.py
```diff
@@ -8,10 +8,14 @@
 
 
 from django.conf import settings
+from django.db.models import Q
 from django.http import HttpResponseRedirect, StreamingHttpResponse
 from django.shortcuts import get_object_or_404
 from django.urls import reverse
 
+from functools import reduce
+from operator import __and__ as AND
```
cutwater, line 17: Please use `operator.and_`

https://docs.python.org/3.9/library/operator.html

```
For backward compatibility, many of these have a variant with the double underscores kept. The variants without the double underscores are preferred for clarity.
```

galaxy_ng/app/api/v3/viewsets/collection.py
```diff
@@ -353,3 +358,94 @@ def _remove_content(collection_version, repo):
...
+                filters = []
+                for d in split_deps:
+                    offset = 0
+                    if '>=' in d or '<=' in d:
+                        offset = 2
+                    if '=' in d or '>' in d or '<' in d:
+                        offset = 1
+                    operand = d[:offset]
+                    version = d[offset:]
+                    if '>=' == operand:
```
cutwater, line 416: You can use the "dispatch dictionary" pattern to replace this if-elif-else construction.

galaxy_ng/app/api/v3/viewsets/collection.py
```diff
@@ -353,3 +358,94 @@ def _remove_content(collection_version, repo):
...
+                    resolved_version_number = results.data['version']
+                    new_dependencies = results.data['dependencies']
+                else:
+                    resolved_version_number = False
+                    new_dependencies = []
+
+            if resolved_version_number:
+                resolved_deps[key] = resolved_version_number
+            else:
+                resolved_deps[key] = f"Conflict! {deps}"
```
cutwater, line 446: In case of dependency conflict it does not produce an error, but returns formatted string. Could you please give examples of API response format?

cutwater (changes requested):

galaxy_ng/app/api/v3/viewsets/collection.py
bmclaughlin: With conflict:
```
{
    "bmclaughlin.conflict2": "0.0.3",
    "bmclaughlin.deps2": "Conflict! {'bmclaughlin.deps2': '>0.1.0,<0.1.0'}"
}
```
Without conflict:
```
{
    "bmclaughlin.deps2": "0.0.3",
    "bmclaughlin.deps1": "0.0.4",
    "bmclaughlin.execution_env_test": "0.1.2",
    "ansible.netcommon": "1.3.0"
}
```

cutwater: Response format in case of conflict is vague and requires parsing on a client.

Can you please give a link to this API format specification?

cutwater (changes requested):

bmclaughlin:
@cutwater There isn't a formal specification for this endpoint.  Would raising a `ValidationError` make more sense in the case of a conflict?  Do you have a recommendation?

chouseknecht:
We're going to migrate this to pulp_ansible.

PR #528, opened 2020-10-23
Add migration to move collections by certification field

cutwater:
Issue: AAH-26

### DIFF
--- a/CHANGES/26.misc
+++ b/CHANGES/26.misc
@@ -1,0 +1,2 @@
+Added migration to distribute collection versions into appropriate repositories
+depending on the `certification` attribute.
--- a/galaxy_ng/app/migrations/0012_move_collections_by_certification.py
+++ b/galaxy_ng/app/migrations/0012_move_collections_by_certification.py
@@ -1,0 +1,95 @@
+from django.db import migrations
+
+MIGRATE_COLLECTIONS_QUERY = """
+CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
+
+CREATE FUNCTION _new_repository_version(repository_id UUID) RETURNS UUID AS $$
+DECLARE
+    result UUID;
+BEGIN
+     INSERT INTO core_repositoryversion (
+        pulp_id, pulp_created, pulp_last_updated, number,
+        complete, base_version_id, repository_id
+    ) VALUES (
+        uuid_generate_v4(), now(), now(),
+        (SELECT next_version FROM core_repository WHERE pulp_id = repository_id),
+        TRUE, NULL, repository_id
+    ) RETURNING pulp_id INTO result;
+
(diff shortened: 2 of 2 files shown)


newswangerd:
I don't really know enough about postgres or the pulp data model to weigh in on this.

@daviddavis @bmbouter could either of you take a look at this?

newswangerd:
The issue number appears to be incorrect. It should be https://issues.redhat.com/browse/AAH-26

galaxy_ng/app/migrations/0011_move_collections_by_certification.py
```diff
@@ -0,0 +1,87 @@
...
+    );
+    from_version UUID;
+    to_version   UUID;
+BEGIN
+    INSERT INTO core_repositoryversion (
+        pulp_id, pulp_created, pulp_last_updated, number, 
+        complete, base_version_id, repository_id
+    ) VALUES (
+        uuid_generate_v4(), now(), now(), 
+        (SELECT max(number) + 1 AS number FROM core_repositoryversion
```
daviddavis, line 23: You should use [next_version on core_repository](https://github.com/pulp/pulpcore/blob/4063c346b4b75ffe74d2b0a95d650732dfb5ed8e/pulpcore/app/models/repository.py#L48). You'll need to increment next_version after this INSERT.

daviddavis:
It may not matter here but in the pulp code, we don't create a new repo version if there's nothing that changes. I don't think it'll harm anything if you do though.

cutwater:
@daviddavis Thank you for review. This is very helpful.

awcrosby (commented):
other than pulp team comment above this seems to in good shape to test

cutwater:
@daviddavis I addressed your comment and refactored it a little to remove code duplication. Could you give another round of review please?

daviddavis (approved):
Looks good to me.

newswangerd (approved):

awcrosby:
@cutwater I tested locally with a database that had 3 collections and 13 collection versions ([pre-migration dump](https://drive.google.com/file/d/1wHpPwER-Esam8LPBxgYbt2lGlx4gfPEW/view?usp=sharing)).

After migration, one of the collection versions had [no repo](https://gist.github.com/awcrosby/473ebf81516a2d6b519c817fb7b4c78f#file-post-cert-migration-results-json-L387), and another had [two repos](https://gist.github.com/awcrosby/473ebf81516a2d6b519c817fb7b4c78f#file-post-cert-migration-results-json-L536-L539). All other collection versions had one repo as expected.

cutwater:
@awcrosby In your migration 3 collections were deleted from the repository. As we discussed this migration is based on assumption (according to the actual state in our environments and the fact that our code don't delete content from pulp repositories) that repositories don't have content that has been deleted.

PR #529, opened 2020-10-26
Fix version query param on ui api collection detail endpoint

awcrosby:
Supports version dropdown on collection detail page.

https://issues.redhat.com/browse/AAH-90

Issue: AAH-90

### DIFF
--- a/galaxy_ng/app/api/ui/serializers/collection.py
+++ b/galaxy_ng/app/api/ui/serializers/collection.py
@@ -121,31 +121,21 @@ class _CollectionSerializer(Serializer):
     def get_deprecated(self, obj):
         return obj.collection.deprecated
 
-    # TODO(awcrosby): refactor once pulp_ansible is_highest param filters by repo
-    # https://pulp.plan.io/issues/7428
     def _get_versions_in_repo(self, obj):
-        repo_param = self.context['path']
-        distro = AnsibleDistribution.objects.get(base_path=repo_param)
+        distro = AnsibleDistribution.objects.get(base_path=self.context['path'])
         repository_version = distro.repository.latest_version()
-        collection_versions = CollectionVersion.objects.filter(collection=obj.collection)
-
-        versions_in_repo = collection_versions.filter(pk__in=repository_version.content)
-        versions_in_repo = sorted(
+        versions_in_repo = CollectionVersion.objects.filter(
+            pk__in=repository_version.content,
+            collection=obj.collection,
+        )
+        return sorted(
             versions_in_repo, key=lambda obj: semantic_version.Version(obj.version), reverse=True
         )
-        return versions_in_repo
-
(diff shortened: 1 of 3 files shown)


galaxy_ng/app/api/ui/viewsets/collection.py
```diff
@@ -24,35 +25,41 @@
...
-    # https://pulp.plan.io/issues/7428
-    def get_queryset(self):
-        """
-        Overrides to not use is_highest param and
-        ensures one collection version per collection is returned.
-        """
-        distro_content = self.get_distro_content(self.kwargs["path"])
+    def get_object(self):
+        """Return CollectionVersion object, latest or via query param 'version'."""
+        version = self.request.query_params.get('version', None)
```
alikins, line 47: Would this be a reasonable place to set a 'lookup_field' on the viewset (ala, https://www.django-rest-framework.org/api-guide/generic-views/#attributes)

alikins: hmm, nm. Given all the other cases to handle, 'lookup_field' isn't super useful

galaxy_ng/app/api/ui/serializers/collection.py
```diff
@@ -121,31 +121,21 @@ def get_namespace(self, obj):
     def get_deprecated(self, obj):
         return obj.collection.deprecated
 
-    # TODO(awcrosby): refactor once pulp_ansible is_highest param filters by repo
-    # https://pulp.plan.io/issues/7428
     def _get_versions_in_repo(self, obj):
```
alikins, line 124: Feels little odd to define _get_versions_in_repo in this class, but only use it in CollectionDetailSerializer.
But shrug, _CollectionSerializer is always subclass so doesn't matter a ton.

galaxy_ng/app/api/ui/viewsets/collection.py
alikins, line 47: Wondering if some of the logic here would fit better in CollectionViewSet.retrieve() ?

alikins: Though, I guess this needs to apply to retrieve() and list(), so get_object() is probably correct.

alikins (approved):
Haven't tested, but LGTM

newswangerd (approved):

galaxy_ng/app/api/ui/serializers/collection.py
```diff
@@ -155,6 +145,7 @@ def get_all_versions(self, obj):
         versions_in_repo = self._get_versions_in_repo(obj)
         return CollectionVersionSummarySerializer(versions_in_repo, many=True).data
 
+    # TODO(awcrosby): rename field to "version_details" since with
+    # "version" query param this won't always be the latest version
```
alikins, line 149: Or possibly in the future, just always return a list for 'version' and let clients assume the first item is the highest or best version.

PR #530, opened 2020-10-26
Add explict IsAuthenticated permission to root endpoints

awcrosby:
Adds explicit permission that was previously inherited

Related to #524, #527

No-Issue


newswangerd (approved):

hendersonreed:
:tada:  This does indeed fix our failing integration test! Thanks for the quick work.

PR #536, opened 2020-10-29
Fix repo content count 

newswangerd:
Previously repo content counts were showing the sum of all content across all repo versions. This update will only show the content in the latest version.

Issue: AAH-94


alikins (approved):
LGTM

PR #537, opened 2020-10-29
Fix import task errors about ansible.CollectionImport

alikins:
The pkey of galaxy's CollectionImport (galaxy_collectionimport table)
is task_id. But the 'task_id' is a OneToOne to a
pulp_ansible.app.models.CollectionImport now instead of just
a plain UUID.

So, get the right pulp CollectionImport and use it for
task_id when creating a galaxy_ng.app.models.CollectionImport.

### DIFF
--- a/CHANGES/99.bugfix
+++ b/CHANGES/99.bugfix
@@ -1,0 +1,1 @@
+Fix `"CollectionImport.task_id" must be a "CollectionImport" instance.` errors on import task.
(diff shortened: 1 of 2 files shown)


Spredzy (approved):
Works for me. Solves the issue.

awcrosby (approved):

chouseknecht (approved):

PR #544, opened 2020-11-02
Add account enhanced API metrics.

chambridge:
* Add account label to django API request metrics
* Update settings to optionally apply metrics to avoid impacting standalone
* Update dev docker-compose.yml with environment variable

Uses the following capability: https://github.com/korfuri/django-prometheus#adding-custom-labels-to-middleware-requestresponse-metrics

# Local Testing
Add `ACCOUNT_ENHANCED_METRICS=True` to `.compose.env`
./compose up

Launch `insights-proxy` and `automation-hub-ui`, login with QA account and view UI

Next check `/metrics` endpoint for data with added *account* label.

```
# HELP django_http_requests_latency_seconds_by_view_method Histogram of request processing time labelled by view.
# TYPE django_http_requests_latency_seconds_by_view_method histogram
django_http_requests_latency_seconds_by_view_method_bucket{account="6089xyz",le="0.01",method="GET",view="galaxy:api:ui:v1:me"} 0.0
django_http_requests_latency_seconds_by_view_method_bucket{account="6089xyz",le="0.025",method="GET",view="galaxy:api:ui:v1:me"} 0.0
django_http_requests_latency_seconds_by_view_method_bucket{account="6089xyz",le="0.05",method="GET",view="galaxy:api:ui:v1:me"} 0.0
django_http_requests_latency_seconds_by_view_method_bucket{account="6089xyz",le="0.075",method="GET",view="galaxy:api:ui:v1:me"} 0.0
django_http_requests_latency_seconds_by_view_method_bucket{account="6089xyz",le="0.1",method="GET",view="galaxy:api:ui:v1:me"} 1.0
django_http_requests_latency_seconds_by_view_method_bucket{account="6089xyz",le="0.25",method="GET",view="galaxy:api:ui:v1:me"} 1.0
django_http_requests_latency_seconds_by_view_method_bucket{account="6089xyz",le="0.5",method="GET",view="galaxy:api:ui:v1:me"} 1.0
django_http_requests_latency_seconds_by_view_method_bucket{account="6089xyz",le="0.75",method="GET",view="galaxy:api:ui:v1:me"} 1.0
django_http_requests_latency_seconds_by_view_method_bucket{account="6089xyz",le="1.0",method="GET",view="galaxy:api:ui:v1:me"} 1.0
django_http_requests_latency_seconds_by_view_method_bucket{account="6089xyz",le="2.5",method="GET",view="galaxy:api:ui:v1:me"} 1.0
django_http_requests_latency_seconds_by_view_method_bucket{account="6089xyz",le="5.0",method="GET",view="galaxy:api:ui:v1:me"} 1.0
django_http_requests_latency_seconds_by_view_method_bucket{account="6089xyz",le="7.5",method="GET",view="galaxy:api:ui:v1:me"} 1.0
django_http_requests_latency_seconds_by_view_method_bucket{account="6089xyz",le="10.0",method="GET",view="galaxy:api:ui:v1:me"} 1.0
django_http_requests_latency_seconds_by_view_method_bucket{account="6089xyz",le="25.0",method="GET",view="galaxy:api:ui:v1:me"} 1.0
django_http_requests_latency_seconds_by_view_method_bucket{account="6089xyz",le="50.0",method="GET",view="galaxy:api:ui:v1:me"} 1.0
django_http_requests_latency_seconds_by_view_method_bucket{account="6089xyz",le="75.0",method="GET",view="galaxy:api:ui:v1:me"} 1.0
django_http_requests_latency_seconds_by_view_method_bucket{account="6089xyz",le="+Inf",method="GET",view="galaxy:api:ui:v1:me"} 1.0
django_http_requests_latency_seconds_by_view_method_count{account="6089xyz",method="GET",view="galaxy:api:ui:v1:me"} 1.0
django_http_requests_latency_seconds_by_view_method_sum{account="6089xyz",method="GET",view="galaxy:api:ui:v1:me"} 0.0929204998537898
```

```
# HELP django_http_requests_total_by_view_transport_method_total Count of requests by view, transport, method.
# TYPE django_http_requests_total_by_view_transport_method_total counter
django_http_requests_total_by_view_transport_method_total{account="6089xyz",method="GET",transport="http",view="galaxy:api:ui:v1:me"} 1.0
django_http_requests_total_by_view_transport_method_total{account="unknown",method="GET",transport="http",view="prometheus-django-metrics"} 2.0
```

### DIFF
--- a/galaxy_ng/app/settings.py
+++ b/galaxy_ng/app/settings.py
@@ -1,5 +1,17 @@
+import os
+
+
+PROMETHEUS_BEFORE_MIDDLEWARE = 'django_prometheus.middleware.PrometheusBeforeMiddleware'
+PROMETHEUS_AFTER_MIDDLEWARE = 'django_prometheus.middleware.PrometheusAfterMiddleware'
+PROM_ACCT_BEFORE = 'galaxy_ng.contrib.metrics_middleware.AccountEnhancedMetricsBeforeMiddleware'
+PROM_ACCT_AFTER = 'galaxy_ng.contrib.metrics_middleware.AccountEnhancedMetricsAfterMiddleware'
+ACCOUNT_ENHANCED_METRICS = os.environ.get('ACCOUNT_ENHANCED_METRICS', False) in (True, 'True')
+if ACCOUNT_ENHANCED_METRICS:
+    PROMETHEUS_BEFORE_MIDDLEWARE = PROM_ACCT_BEFORE
+    PROMETHEUS_AFTER_MIDDLEWARE = PROM_ACCT_AFTER
+
 MIDDLEWARE = [
-    'django_prometheus.middleware.PrometheusBeforeMiddleware',
+    PROMETHEUS_BEFORE_MIDDLEWARE,
     # BEGIN: Pulp standard middleware
     'django.middleware.security.SecurityMiddleware',
     'whitenoise.middleware.WhiteNoiseMiddleware',
@@ -10,7 +22,7 @@ MIDDLEWARE = [
     'django.contrib.messages.middleware.MessageMiddleware',
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
     # END: Pulp standard middleware
-    'django_prometheus.middleware.PrometheusAfterMiddleware',
+    PROMETHEUS_AFTER_MIDDLEWARE,
     'django_currentuser.middleware.ThreadLocalUserMiddleware',
 ]
 
--- a/galaxy_ng/contrib/metrics_middleware.py
+++ b/galaxy_ng/contrib/metrics_middleware.py
@@ -1,0 +1,50 @@
+import base64
+import json
+
+from django_prometheus.middleware import (
+    Metrics,
+    PrometheusAfterMiddleware,
+    PrometheusBeforeMiddleware,
+)
+
+from galaxy_ng.app.auth.auth import RHIdentityAuthentication
+
+
+EXTENDED_METRICS = [
+    "django_http_requests_latency_seconds_by_view_method",
+    "django_http_responses_total_by_status_view_method",
+    "django_http_requests_total_by_view_transport_method",
+]
+
+
+class AccountEnhancedMetrics(Metrics):
+    def register_metric(self, metric_cls, name, documentation, labelnames=(), **kwargs):
+        if name in EXTENDED_METRICS:
+            labelnames += ("account",)
+        return super().register_metric(
+            metric_cls, name, documentation, labelnames=labelnames, **kwargs
+        )
+
+
+class AccountEnhancedMetricsBeforeMiddleware(PrometheusBeforeMiddleware):
+    metrics_cls = AccountEnhancedMetrics
+
+
+class AccountEnhancedMetricsAfterMiddleware(PrometheusAfterMiddleware):
+    metrics_cls = AccountEnhancedMetrics
+
+    def label_metric(self, metric, request, response=None, **labels):
+        new_labels = labels
+        if metric._name in EXTENDED_METRICS:
+            account = "unknown"
+            encoded_identity = request.META.get(RHIdentityAuthentication.header)
+            if encoded_identity is not None:
+                try:
+                    json_string = base64.b64decode(encoded_identity)
+                    identity_header = json.loads(json_string)
+                    account = identity_header.get("identity", {}).get("account_number", "unknown")
+                except ValueError:
+                    pass
+            new_labels = {"account": account}
+            new_labels.update(labels)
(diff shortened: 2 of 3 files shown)


newswangerd:
I think this looks fine, but I'd like to get sign off from @cutwater

galaxy_ng/contrib/metrics_middleware.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+class AccountEnhancedMetricsAfterMiddleware(PrometheusAfterMiddleware):
+    metrics_cls = AccountEnhancedMetrics
+
+    def label_metric(self, metric, request, response=None, **labels):
+        new_labels = labels
+        if metric._name in EXTENDED_METRICS:
+            account = "unknown"
+            encoded_identity = request.META.get(RHIdentityAuthentication.header)
```
alikins, line 40: Seems like this will handle standalone cases where RHIdentityAuthentication will never be available by setting account="unknown".

So wondering if this should be toggled on/off based on DEPLOYMENT_MODE. But I also like using
ACCOUNT_ENHANCED_METRICS as the flag since it's not coupled to the deployment mode.
So as is is fine I think

galaxy_ng/contrib/metrics_middleware.py
```diff
@@ -0,0 +1,50 @@
...
+            if encoded_identity is not None:
+                try:
+                    json_string = base64.b64decode(encoded_identity)
+                    identity_header = json.loads(json_string)
+                    account = identity_header.get("identity", {}).get("account_number", "unknown")
+                except ValueError:
+                    pass
+            new_labels = {"account": account}
+            new_labels.update(labels)
```
alikins, line 49: Is adding the account number to the metric labels going to cause any issues?  Account number will be fairly high cardinality and wondering if that could become a problem.

https://prometheus.io/docs/practices/instrumentation/#do-not-overuse-labels

> As a general guideline, try to keep the cardinality of your metrics below 10, and for metrics that exceed that, aim to limit them to a handful across your whole system. The vast majority of your metrics should have no labels.
> 
> If you have a metric that has a cardinality over 100 or the potential to grow that large, investigate alternate solutions such as reducing the number of dimensions or moving the analysis away from monitoring and to a general-purpose processing system.

galaxy_ng/contrib/metrics_middleware.py
alikins: A new metric will be created/updated for every request?  

Instead of extracting it from the RHIdentity header each time, could set any of that info needed on the request once and then reference it here?  For ex, I think the request has the 'user' and the 'user' groups are essential the account number.

That would also provide an alternate 'account' value for standalone deployments.

alikins (changes requested):
Mostly worried about the metric cardinality stuff.

galaxy_ng/contrib/metrics_middleware.py
chambridge: In a standalone case you wouldn't enable ACCOUNT_ENHANCED_METRICS so the `account="unknown"` wouldn't be there, the unknown would only be hit for routes like `/metrics` or a status endpoint that aren't hit by 3scale.

galaxy_ng/app/settings.py
```diff
@@ -1,5 +1,17 @@
+import os
+
+
+PROMETHEUS_BEFORE_MIDDLEWARE = 'django_prometheus.middleware.PrometheusBeforeMiddleware'
+PROMETHEUS_AFTER_MIDDLEWARE = 'django_prometheus.middleware.PrometheusAfterMiddleware'
+PROM_ACCT_BEFORE = 'galaxy_ng.contrib.metrics_middleware.AccountEnhancedMetricsBeforeMiddleware'
+PROM_ACCT_AFTER = 'galaxy_ng.contrib.metrics_middleware.AccountEnhancedMetricsAfterMiddleware'
+ACCOUNT_ENHANCED_METRICS = os.environ.get('ACCOUNT_ENHANCED_METRICS', False) in (True, 'True')
```
cutwater, line 8: We use `1 \ 0` for booleans outside python and `true \ false` because of dynaconf. Please use one of these styles for consistency (or both)

For example:

```suggestion
ACCOUNT_ENHANCED_METRICS = os.environ.get('ACCOUNT_ENHANCED_METRICS') in ('1', 'true')
```

cutwater (changes requested):

galaxy_ng/contrib/metrics_middleware.py
cutwater: Agreed, considering number of accounts may be large and grow over time and that this code adds labels to all metrics, I'm concerned it may cause performance issues. Also this particular use case seems to be more appropriate for the logging system rather then metrics collection.

PR #558, opened 2020-11-06
add sdb support to environment

chrismeyersfsu:
https://asciinema.org/a/c9K6ydZxCK869PHcl5sZiekb8

Not sure which environment I should put the sdb package in. Threw it in standalone.

### DIFF
--- a/galaxy_ng/app/settings.py
+++ b/galaxy_ng/app/settings.py
@@ -1,3 +1,5 @@
+import os
+
 MIDDLEWARE = [
     'django_prometheus.middleware.PrometheusBeforeMiddleware',
     # BEGIN: Pulp standard middleware
@@ -91,3 +93,5 @@ X_PULP_API_PREFIX = "pulp_ansible/galaxy/automation-hub/api"
 
 X_PULP_CONTENT_HOST = "localhost"
 X_PULP_CONTENT_PORT = 24816
+
+os.environ['SDB_NOTIFY_HOST'] = os.popen('ip route').read().split(' ')[2]
--- a/dev/docker-compose.yml
+++ b/dev/docker-compose.yml
@@ -21,9 +21,12 @@ services:
     command: ['run', 'api-reload']
     ports:
       - "5001:8000"
+      - "7899-7999:7899-7999" # default sdb ports
     environment:
       - "WITH_DEV_INSTALL=1"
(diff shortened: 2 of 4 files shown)


galaxy_ng/app/settings.py
```diff
@@ -91,3 +93,5 @@
 
 X_PULP_CONTENT_HOST = "localhost"
 X_PULP_CONTENT_PORT = 24816
+
+os.environ['SDB_NOTIFY_HOST'] = os.popen('ip route').read().split(' ')[2]
```
ryanpetrello, line 97: Worth pointing out, I think this only works if you're running Docker on Linux hosts, not "Docker for Mac".

chouseknecht: Is there a workaround? I think we have some folks developing on Mac OS.

cutwater (changes requested):
@chrismeyersfsu What is the advantage of the `sdb`, over `rpdb` or similar solutions, that exposes a listen port to connect to?

This PR adds dependency on an external tool, which is called only for debugging purposes. At least it should be conditional. In general I would prefer to avoid calling external environment-specific tools if possible.

galaxy_ng/app/settings.py
chrismeyersfsu: Yep @chouseknecht there is. I can include that. I _think_ it's only needed for the OLDDDD docker for mac that ran in a VM and not the newer docker for mac that runs in a hypervisor.

chrismeyersfsu:
@cutwater Looks like there is a `dev_requirements.txt` but I couldn't find any reference to the usage of that file. Maybe I can resurrect that conditional behavior or just create a new conditional behavior. Was hoping to get a pointer on where to put this dependency from folks more familiar with the code base.

> @chrismeyersfsu What is the advantage of the sdb, over rpdb or similar solutions, that exposes a listen port to connect to?
I'm not actually sure. Let me ask @ryanpetrello

chrismeyersfsu:
@cutwater not much different in `sdb` vs. `rpdb`. I say, just that the Tower team uses `sdb` (written before `rpdb` existed).

rochacbruno:
I wrote the steps I am using to debug with sdb here: https://github.com/ansible/galaxy_ng/wiki/Debugging-with-PDB

It would be still great if we can add the dependencies to the base image so we save some time.

PR #546, opened 2020-11-03
Fix the passing along of filename data

awcrosby:
https://issues.redhat.com/browse/AAH-118

dependent on https://github.com/pulp/pulp_ansible/pull/407

PR #561, opened 2020-11-10
Upgrade pulp_ansible to 0.5.1

awcrosby:
Updates included in 0.5.1:
* allows for galaxy_ng fix: Fix the passing along of filename data #546
* track when versions for a collection have been updated https://pulp.plan.io/issues/7775, includes field updated_at to use latest instead of highest version pulp/pulp_ansible#416
* resolve token refreshes every request to cloud.redhat.com

No-Issue

### DIFF
--- a/.travis/before_install.sh
+++ b/.travis/before_install.sh
@@ -101,7 +101,7 @@ if [ -z "$TRAVIS_TAG" ]; then
 fi
 
 
-git clone --depth=1 https://github.com/pulp/pulp_ansible.git --branch 0.5.0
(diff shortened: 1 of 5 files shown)


alikins (changes requested):
Add references to the "why" for the upgrade.

Namely, pulp issue https://pulp.plan.io/issues/7775  and the https://github.com/pulp/pulp_ansible/pull/416  change in particular.

(At least I think that is the main reason for needing 0.5.1, if there are others mention that as well)

alikins (approved):
LGTM

awcrosby:
When this is merged, we will need to re-apply the diff in #546

PR #562, opened 2020-11-10
Fix for 2 issues reported on AHH-132

rochacbruno:
Related isssue: https://issues.redhat.com/browse/AAH-132

1. Namespace creation fails when links are associated.
  solution: link_data was missing relationship with namespace instance

2. Error 500 was raised when trying to create a namespace with existing name.
  solution: validate for existence before the creation of dependent instances on ViewSet.create method.

```
❯ http post :8002/api/automation-hub/_ui/v1/namespaces/ "Authorization: Token b1a99acc28ebfd33fc2ce4684408079456efe6de" <<<'{"name":"testing15","company":"Redhat","email":"user@example.com","avatar_url":"https://www.redhat.com/cms/managed-files/styles/wysiwyg_full_width/s3/Logo-RedHat-Hat-Color-CMYK%20%281%29.jpg?itok=Mf0Ff9jq","description":"string","links":[{"name": "homepage","url":"http://www.google.com"}],"groups":[{"name":"system:partner-engineers","object_permissions":["change_namespace","upload_to_namespace"]}]}'
HTTP/1.1 409 Conflict
X-Powered-By: Express
allow: GET, POST, HEAD, OPTIONS
connection: close
content-length: 210
content-type: application/json
date: Tue, 10 Nov 2020 18:28:12 GMT
server: gunicorn/20.0.4
vary: Accept, Cookie
x-frame-options: SAMEORIGIN

{
    "errors": [
        {
            "code": "conflict",
            "detail": "There is an existing namespace with the same name, choose a different name",
            "source": {
                "parameter": "name"
            },
            "status": "409",
            "title": "Data conflicts with existing entity."
        }
    ]
}
```

### DIFF
--- a/galaxy_ng/app/api/v3/viewsets/namespace.py
+++ b/galaxy_ng/app/api/v3/viewsets/namespace.py
@@ -1,6 +1,7 @@
 import logging
 
 from django.db import transaction
+from django.db.utils import IntegrityError
 from django_filters import filters
 from django_filters.rest_framework import filterset, DjangoFilterBackend
 from pulp_ansible.app.models import AnsibleRepository, AnsibleDistribution
@@ -60,15 +61,15 @@ class NamespaceViewSet(api_base.ModelViewSet):
     def create(self, request, *args, **kwargs):
         """Override to also create inbound pulp repository and distribution."""
 
-        if models.Namespace.objects.filter(name=request.data['name']).exists():
+        try:
+            name = INBOUND_REPO_NAME_FORMAT.format(namespace_name=request.data['name'])
+            repo = AnsibleRepository.objects.create(name=name)
+            AnsibleDistribution.objects.create(name=name, base_path=name, repository=repo)
+            return super().create(request, *args, **kwargs)
+        except IntegrityError as e:
             raise ConflictError(detail={
-                'name': 'There is an existing namespace with the same name, choose a different name'
+                'name': 'There is a conflict error: {}'.format(str(e))
             })
-
-        name = INBOUND_REPO_NAME_FORMAT.format(namespace_name=request.data['name'])
-        repo = AnsibleRepository.objects.create(name=name)
-        AnsibleDistribution.objects.create(name=name, base_path=name, repository=repo)
-        return super().create(request, *args, **kwargs)
 
     @transaction.atomic
     def destroy(self, request, *args, **kwargs):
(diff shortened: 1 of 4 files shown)


galaxy_ng/app/api/v3/viewsets/namespace.py
```diff
@@ -58,6 +59,12 @@ def get_serializer_class(self):
     @transaction.atomic
     def create(self, request, *args, **kwargs):
         """Override to also create inbound pulp repository and distribution."""
+
+        if models.Namespace.objects.filter(name=request.data['name']).exists():
+            raise ValidationError(detail={
+                'name': 'There is an existing namespace with the same name, choose a different name'
+            })
```
rochacbruno, line 66: This validation has to be done here because this `create` method is called before the validators defined on the serializer.

another option if we want to keep validation on the serializer will be moving the creation of Repo and Distro to 
there so there will be validation happening in a single place.

cutwater: IMO here 409 conflict response code is more appropriate.

alikins: @rochacbruno 

> another option if we want to keep validation on the serializer will be moving the creation of Repo and Distro to there so there will be validation happening in a single place.

Yeah, possibly. That was debated originally iirc. I think the overall thinking was to try to keep db stuff out of the serializers as far as possible (and in the viewsets). Another common option seems to be to move the "create three different model instances at the same time" stuff into one (or more) of the model objectManagers. 

So this seems reasonable for now, with a note to revisit it.

alikins (approved):
Seems reasonable to me.

Though we should revisit how the per namespace Repository/Distribution are created later I think.

galaxy_ng/app/api/v3/viewsets/namespace.py
alikins: Would this be a place to apply ScopedErrorListSerializer https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/api/v3/serializers/namespace.py#L17 ?

alikins: @cutwater If 409 is more appropriate here (and I agree that it is), it feels like letting this hit a db error and catching the exception somewhere appropriate would be my expectation.

chouseknecht (approved):

galaxy_ng/app/api/v3/viewsets/namespace.py
rochacbruno: Resolved as a 409 conflict error

rochacbruno: @alikins changed to use try/except as you suggested so it goes EAFP and hit the database

rochacbruno:
Issue AAH-139 https://issues.redhat.com/browse/AAH-139 filled as a follow up for this PR.

newswangerd (approved):

PR #564, opened 2020-11-10
Revert "Fix the passing along of filename data (#546)"

awcrosby:
This reverts commit 0ab1e0f which requires pulp_ansible 0.5.1

No-Issue


alikins (approved):
Revert is fine. But I'm pretty confused about which pulp_ansible version we want/need to require and when we need to do it.  ie, I'm not following if this obsoletes https://github.com/ansible/galaxy_ng/pull/561

PR #566, opened 2020-11-10
Add token param to unit tests for syncconfig

awcrosby:
Required by pulp_ansible https://github.com/pulp/pulp_ansible/pull/418

No-Issue


chouseknecht (approved):

alikins (approved):
Seems OKAY.

I'm not sure if the sync config should always require a 'token' and a 'auth_url' (not sure that other scenarios the test cover), so there may be other cases to test  (mainly, just 'token' I think)

PR #593, opened 2020-12-04
Handle drf auth/perm exc to keep the better error msgs

alikins:
All of the drf auth and perm related exceptions are
subclasses of ApiException, so check the auth related
ones first, to avoid raising a PermissionDenied() with
no info

Pass drf exception context to our exception handler

The exc context includes the request and view objects
and the request args/kwargs

Add AccessPolicy and deployment mode to api exc context

### DIFF
--- a/CHANGES/206.feature
+++ b/CHANGES/206.feature
@@ -1,0 +1,1 @@
+Provide more context to access_policy permission errors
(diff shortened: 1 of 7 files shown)


chouseknecht:
Closing due to inactivity.

chouseknecht:
Closing due to inactivity.

PR #594, opened 2020-12-04
Don't allow user to modify other users groups aah148

alikins:
Fix standalone UserAccessPolicy to require is_current_user

Add has_groups_param_obj_perms condition to UserAccessPolicy

### DIFF
--- a/galaxy_ng/app/api/ui/serializers/user.py
+++ b/galaxy_ng/app/api/ui/serializers/user.py
@@ -44,13 +44,20 @@ class UserSerializer(serializers.ModelSerializer):
     def validate_groups(self, groups):
         request_user = self.context['request'].user
 
+        group_set = set(groups)
+        instance_group_set = set()
+        if self.instance:
+            instance_group_set = set(list(self.instance.groups.all()))
+
+        group_difference = instance_group_set.symmetric_difference(group_set)
+
         if not request_user.has_perm('galaxy.change_group'):
             authed_user_groups = request_user.groups.all()
-            for g in groups:
+            for g in group_difference:
                 if not authed_user_groups.filter(pk=g.id).exists():
                     raise ValidationError(detail={
-                        "groups": "'galaxy.change_group' permission is required to add"
-                                  " users to a group that the current user is not in."
+                        "groups": "'galaxy.change_group' permission is required to change"
+                                  " a users group that the requesting user is not in."
                     })
 
         return groups


galaxy_ng/app/access_control/statements/standalone.py
```diff
@@ -94,9 +94,15 @@
         },
         {
             "action": ["update", "partial_update"],
-            "principal": "authenticated",
+            "principal": ["authenticated"],
+            "effect": "allow",
+            "condition": "is_current_user"
```
newswangerd, line 99: We don't want to allow users to update themselves on the `UserViewSet`. The `UserViewSet` contains the ability to update a user's groups, which means that if user's can update themselves from this API endpoint, they can change their groups and elevate their permissions. That's why we have a separate access policy for `MyUserViewSet`

galaxy_ng/app/access_control/statements/standalone.py
```diff
@@ -94,9 +94,15 @@
...
+        },
+        {
+            "action": ["update", "partial_update"],
+            "principal": ["group:org_admin"],
```
newswangerd, line 103: Don't use group principals. This is means that nobody can update any users unless they are `org_admin` group, which is the same as just hardcoding `system:partner-engineers` like we used to do.

galaxy_ng/app/access_control/access_policy.py
```diff
@@ -62,13 +63,34 @@ class CollectionRemoteAccessPolicy(AccessPolicyBase):
...
+
+        """
+
+        serializer = UserSerializer(data=request.data, context={"request": request})
+
+        try:
+            serializer.is_valid(raise_exception=True)
+        except Exception as exc:
+            log.exception(exc)
+            raise
```
newswangerd, line 84: Something about this seems broken. I get a `A user with that username already exists` message when I try to edit users and when I try to add a new group I get this 500

```
api_1               | pulp: django.request:ERROR: Internal Server Error: /api/automation-hub/_ui/v1/users/3/
api_1               | Traceback (most recent call last):
api_1               |   File "/venv/lib64/python3.6/site-packages/django/core/handlers/exception.py", line 34, in inner
api_1               |     response = get_response(request)
api_1               |   File "/venv/lib64/python3.6/site-packages/django/core/handlers/base.py", line 115, in _get_response
api_1               |     response = self.process_exception_by_middleware(e, request)
api_1               |   File "/venv/lib64/python3.6/site-packages/django/core/handlers/base.py", line 113, in _get_response
api_1               |     response = wrapped_callback(request, *callback_args, **callback_kwargs)
api_1               |   File "/venv/lib64/python3.6/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
api_1               |     return view_func(*args, **kwargs)
api_1               |   File "/venv/lib64/python3.6/site-packages/rest_framework/viewsets.py", line 114, in view
api_1               |     return self.dispatch(request, *args, **kwargs)
api_1               |   File "/venv/lib64/python3.6/site-packages/rest_framework/views.py", line 505, in dispatch
api_1               |     response = self.handle_exception(exc)
api_1               |   File "/venv/lib64/python3.6/site-packages/rest_framework/views.py", line 465, in handle_exception
api_1               |     self.raise_uncaught_exception(exc)
api_1               |   File "/venv/lib64/python3.6/site-packages/rest_framework/views.py", line 476, in raise_uncaught_exception
api_1               |     raise exc
api_1               |   File "/venv/lib64/python3.6/site-packages/rest_framework/views.py", line 502, in dispatch
api_1               |     response = handler(request, *args, **kwargs)
api_1               |   File "/venv/lib64/python3.6/site-packages/rest_framework/mixins.py", line 67, in update
api_1               |     serializer.is_valid(raise_exception=True)
api_1               |   File "/venv/lib64/python3.6/site-packages/rest_framework/serializers.py", line 234, in is_valid
api_1               |     self._validated_data = self.run_validation(self.initial_data)
api_1               |   File "/venv/lib64/python3.6/site-packages/rest_framework/serializers.py", line 433, in run_validation
api_1               |     value = self.to_internal_value(data)
api_1               |   File "/src/galaxy_ng/galaxy_ng/app/api/ui/serializers/user.py", line 88, in to_internal_value
api_1               |     for field in group:
api_1               | TypeError: 'int' object is not iterable
```

galaxy_ng/app/access_control/access_policy.py
```diff
@@ -62,13 +63,34 @@ class CollectionRemoteAccessPolicy(AccessPolicyBase):
...
+
+        groups = serializer.validated_data.get("groups")
+
+        return request.user.has_perm(permission, groups)
```
newswangerd, line 88: I don't think this will work. This is is just checking to see if the any of the groups that are being added to the user have the `galaxy.change_user` permission, which doesn't seem right to me.

newswangerd (changes requested):

galaxy_ng/app/access_control/statements/standalone.py
alikins: Good catch. That is a left over from an earlier approach that needs removing.

galaxy_ng/app/access_control/access_policy.py
alikins: What user is the request being made as?   Did the request end up being a POST or a PATCH?

alikins: Hmm, I'm seeing variation of the user validation failing now.  I didn't seem to hit that before so I'm not sure what is going on.

ie, something like 

```
 galaxyapi PATCH /api/automation-hub/_ui/v1/users/18/ groups:='[{"id":10, "name":"rh-identity-account:99996"}]'
```

Was working before. 

I suspect something is preventing the "partial" users from being valid now. I don't think I changed how it was called.
Investigating...

alikins: Yeah, I lost a 'partial=True' for that serializers somewhere. Updated.

alikins:
@newswangerd Updated with a new approach. Just making sure the UserSerializer.validate_groups() checks any groups that have been changed, not just added. All the access_policy changes are removed.

newswangerd (approved):

alikins:
wasn't handling the case of self.instance=None, updated to handle that.

PR #611, opened 2020-12-16
Open API Updates AAH-57

alikins:
WIP for getting a openapi spec we can use with c.r.c

Issue: AAH-57

### DIFF
--- a/galaxy_ng/app/common/openapi.py
+++ b/galaxy_ng/app/common/openapi.py
@@ -1,0 +1,25 @@
+import logging
+
+log = logging.getLogger(__name__)
+log.debug('openapi autoschema')
+
+
+def preprocess_debug_logger(endpoints, **kwargs):
+    """Return an iterable of (path, path_regex, method, callback) tuples"""
+
+    log.debug("kwargs: %s", repr(kwargs))
+
+    return [
+        (path, path_regex, method, callback) for path, path_regex, method, callback in endpoints
+        if not log.debug('path=%s, path_regex=%s, methos=%s, callback=%s',
+                         path, path_regex, method, callback)
+    ]
+
+
+def preprocess_exclude_pulp_endpoints(endpoints, **kwargs):
+    """Return an iterable of (path, path_regex, method, callback) with /pulp* endpoints removed"""
+
+    return [
+        (path, path_regex, method, callback) for path, path_regex, method, callback in endpoints
+        if not path.startswith('/pulp')
+    ]
--- a/galaxy_ng/app/common/openapi.py
+++ b/galaxy_ng/app/common/openapi.py
@@ -5,21 +5,28 @@ log.debug('openapi autoschema')
 
 
 def preprocess_debug_logger(endpoints, **kwargs):
-    """Return an iterable of (path, path_regex, method, callback) tuples"""
+    """Log and then return an iterable of (path, path_regex, method, callback) tuples"""
 
     log.debug("kwargs: %s", repr(kwargs))
 
-    return [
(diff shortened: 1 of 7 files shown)


galaxy_ng/app/settings.py
```diff
@@ -82,3 +82,20 @@
 # ----------------------
 X_PULP_CONTENT_HOST = "localhost"
 X_PULP_CONTENT_PORT = 24816
+
+# works
+# SPECTACULAR_SETTINGS__TITLE = "Automation Hub API __TITLE"
```
awcrosby, line 87: are these lines something that can be removed since title is below?

galaxy_ng/app/settings.py
```diff
@@ -82,3 +82,20 @@
...
+
+SPECTACULAR_SETTINGS = {
+    "TITLE": "Automation Hub API",
+    "DESCRIPTION": "Fetch, Upload, Organize, and Distribute Ansible Collections",
+    "VERSION": "v3",
+    "LICENSE": {
+        "name": "GPLv2+",
+        "url": "https://raw.githubusercontent.com/pulp/pulpcore/master/LICENSE",
```
awcrosby, line 95: should we link to galaxy_ng license?

galaxy_ng/app/common/openapi.py
```diff
@@ -0,0 +1,26 @@
...
+
+def preprocess_debug_logger(endpoints, **kwargs):
+    """Return an iterable of (path, path_regex, method, callback) tuples"""
+
+    log.debug("kwargs: %s", repr(kwargs))
+
+    return [
+        (path, path_regex, method, callback) for path, path_regex, method, callback in endpoints
+        if not log.debug('path=%s, path_regex=%s, methos=%s, callback=%s',
+                         path, path_regex, method, callback)
```
awcrosby, line 15: curious what `preprocess_debug_logger()` is used for... and does `log.debug(...)` sometimes evaluate to a boolean false by not returning any result?

awcrosby (approved):
nothing blocking merging from my pov, but did add a couple clarifying questions.

galaxy_ng/app/settings.py
alikins: Yeah, that is meant as an example of how to set drf_spectular settings vis env variables / dynaconf. I'll rephrase to make that clearer.

galaxy_ng/app/settings.py
alikins: Ah, yeah. Good catch.

galaxy_ng/app/common/openapi.py
alikins: nah, log.debug() always returns nothing. That is leftover from some c&p from the other preprocess function

alikins: I take it back, forgot it was in the 'if' clause of the generator, and the if `not log.debug()` ends up if True, so all the elements are returned.

alikins: I've refactored that a bit to look less weird. I'll update after a './compose build' cycle to catch up some deps to test.

hendersonreed:
A lot of the api's endpoints take specific parameters when POSTing- is there a way to generate request samples for those? Currently it looks like there's only request samples for the PATCH and PUT listings.

alikins:
@hendersonreed Yeah, lots of stuff is incomplete at the moment. It can be corrected, either by tweaks to the spec generator, or direct markup on the viewsets and serializers.

hendersonreed:
Seems to me that ideally we do those QoL improvements before merging this? I'm happy to contribute code if I can, since I'll also be a consumer of the final result.

PR #612, opened 2021-01-04
Fix how CI checks for Jira issue existence

rochacbruno:
Check for 404 status code instead of `ok` so it works
even for unlogged sessions.

Issue: AAH-44

### DIFF
--- a/.travis/custom_check_pull_request.py
+++ b/.travis/custom_check_pull_request.py
@@ -66,7 +66,10 @@ def check_changelog_record(issue):
 
 def check_issue_exists(issue):
     response = requests.head(JIRA_URL.format(issue=issue))
-    if not response.ok:
+    if response.status_code == 404:
+        # 200 is returned for logged in sessions
+        # 401 is for not authorized access on existing issue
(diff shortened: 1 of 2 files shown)


.travis/custom_check_pull_request.py
```diff
@@ -66,7 +66,10 @@ def check_changelog_record(issue):
 
 def check_issue_exists(issue):
     response = requests.head(JIRA_URL.format(issue=issue))
-    if not response.ok:
+    if response.status_code == 404:
```
newswangerd, line 69: Does jira ever respond with any other 4xx codes?

rochacbruno: @newswangerd if issue doesn't exist it responds with 404, as we only want to check if the issue exists any other code is valid, like 401 (unauthorized) or 200 which will return only if user is authenticated.

Jira can respond with server error (50x) but in that case do we want to block the CI?

alikins (approved):
I like it, simple solution. Nice!

newswangerd (approved):

PR #614, opened 2021-01-14
Sync toggle 222

alikins:
Add access_policy so synclist need org admin to see
    
    Add a is_org_admin access_condition to the base access policy.
    This checks the x_rh_if if available to see if use is an org admin
    
    Apply is_org_admin access condition to synclists on insights
    
    Issue: AAH-222

(also currently includes)
    Add AccessPolicyVerboseMixin
    
    blip
    Add some debug logging about access policy for troubleshooting
    
    Issue: AAH-205


alikins:
I might split the AccessPolicyVerboseMixin out to it's own .py file. Or should I leave it as is?

galaxy_ng/app/access_control/statements/insights.py
```diff
@@ -90,31 +90,36 @@
             "action": ["list"],
             "principal": "authenticated",
             "effect": "allow",
-            "condition": ["has_model_perms:galaxy.view_synclist", "has_rh_entitlements"]
+            "condition": ["has_model_perms:galaxy.view_synclist",
+                          "has_rh_entitlements",
+                          "is_org_admin"]
```
awcrosby, line 95: any reason to have `has_rh_entitlements` in these statements when `is_org_admin` already calls underlying entitlement check?

galaxy_ng/tests/unit/api/rh_auth.py
```diff
@@ -15,7 +15,8 @@ def user_x_rh_identity(username, account_number=None):
                 {"username": "%(username)s",
                 "email": "%(username)s",
                 "first_name": "%(title_username)s",
-                "last_name": "%(title_username)sington"
+                "last_name": "%(title_username)sington",
+                "is_org_admin": true
```
awcrosby, line 19: do the unit tests in `test_api_ui_synclists.py` check for `is_org_admin` being true and/or false?, seems most test logic changes are log outputs rather than asserts

galaxy_ng/app/access_control/statements/insights.py
alikins: is_org_admin requiring has_rh_entitlements (essentially...) is kind of an implementation detail of is_org_admin. 
is_org_admin and has_rh_entitlement are slightly abstract so in theory, is_org_admin may not required has_rh_entitlement.

But mainly just to be consistent with the other access policies that require has_rh_entitlements as a separate and sometimes only condition. It's a little bit more explicit.

galaxy_ng/tests/unit/api/rh_auth.py
alikins: If the unit tests are run in 'insights' mode, the viewset permissions use the insights access_policy which will enforce is_org_admin. 

In 'standalone' mode, most of those tests are skipped.  And by default the tests run in 'standalone' mode (in CI as well), so most of those test don't run. 

(There is a wider issue related to how/when/when settings and access_policy get loaded that some of the deployment mode specific code can't be fully mocked or overridden. Running full test suite in 'insights' mode causes auth related tests to fail for the same reason except vice versa).

So at the moment, most of test_api_ui_synclist tests are only hit when running only the synclists tests using 'insights' deployment mode. Fixing the deployment mode dependent tests (in synclists and elsewhere) needs a deeper fix.

awcrosby (approved):
LGTM

PR #617, opened 2021-01-19
Some fiddling with allowing dev to use pulpcore pulp_ansible master etc

alikins:
The main downsides here are:

- The python modules installed in the 'dev' containers will not match that of 'hosted' containers or of 'standalone' releases
- The python modules installed will potentially be different every time it's started up. [1]
- The python modules installed will likely be slightly different for every developer at any given time.
- Since pip installs of various things happen later in the process, and with fewer constraints, each startup will likely take noticeably longer as it more or less re-installs all the modules.

ie, the issues the current pip-compile based tools tries to avoid.

[1] And that could include changing between each container startup on a deploy. ie, 'api' and 'pulp-worker' could get a different version of 'requests' or whatever


setup.py
```diff
@@ -58,16 +56,13 @@ def run(self):
...
-    "django-prometheus>=2.0.0",
-    "drf-spectacular",
+    "pulp-ansible",
+    "django-prometheus",
 ]
 
 package_name = os.environ.get("GALAXY_NG_ALTERNATE_NAME", "galaxy-ng")
-version = os.environ.get("ALTERNATE_VERSION", __version__)
+# version = os.environ.get("ALTERNATE_VERSION", __version__)
+version = "4.3.0.dev"
```
rochacbruno, line 65: version is defined on `__init__.py` and then repeated here again, I think we should have a single place where version is defined, in this case here we can get back to the approach of reading it `statically` from `__init__.py`

chouseknecht:
Closing due to inactivity.

PR #626, opened 2021-02-01
Add field detailing if a write only value is set or not.

newswangerd:
Issue: AAH-120

Adds the following field to the sync endpoints:

```
    "write_only_fields": [
        {
            "name": "client_key",
            "is_set": false
        },
        {
            "name": "token",
            "is_set": false
        },
        {
            "name": "password",
            "is_set": false
        },
        {
            "name": "proxy_password",
            "is_set": true
        }
    ]
 ```

### DIFF
--- a/galaxy_ng/app/api/v3/serializers/sync.py
+++ b/galaxy_ng/app/api/v3/serializers/sync.py
@@ -245,11 +245,12 @@ def get_write_only_fields(serializer, obj, extra_data={}):
     """
     fields = []
 
+    # returns false if field is "" or None
     def _is_set(field_name):
         if (field_name in extra_data):
-            return extra_data[field_name] is not None
+            return bool(extra_data[field_name])
         else:
-            return getattr(obj, field_name) is not None
+            return bool(getattr(obj, field_name))
 
     # There are two ways to set write_only. This checks both.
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: galaxy_ng/app/api/v3/serializers/sync.py [hunk 1]
```diff
@@ -129,6 +131,15 @@ def validate(self, data):
             )
         return super().validate(data)
 
+    def get_write_only_fields(self, obj):
+        fields = []
+
+        for e in self.Meta.extra_kwargs:
+            if self.Meta.extra_kwargs[e].get('write_only', False):
+                fields.append({"name": e, "is_set": getattr(obj, e) is None})
```
rochacbruno, line 139: `"is_set": getattr(obj, e) is None`

What dows it return if field is not set? like the first time it is requested?

galaxy_ng/app/api/v3/serializers/sync.py
```diff
@@ -225,3 +232,36 @@ def get_last_sync_task_queryset(self, obj):
...
+    serializer: Serializer instance
+    obj: model object being serialized
+    extra_data: extra fields that might not be on obj. This is used when a write
+        only field is not one of the fields in the underlying data model.
+    """
+    fields = []
+
+    def _is_set(field_name):
+        if (field_name in extra_data):
+            return extra_data[field_name] is not None
```
rochacbruno, line 250: If you use `bool` you address the `""` and `None`
```suggestion
            return bool(extra_data[field_name])
```

rochacbruno (approved):

newswangerd:
Removing backport-4.3 label since this was merged before we started backporting.

PR #627, opened 2021-02-02
Users shouldn't be able to delete themselves (AAH-265)

alikins:
Add and use is_current_user to the user 'destroy' condition.

Any requests for a user to delete themselves is denied, causing
a 403 response.

Fix name of TestUiUserViewSet
Add a unit test for user delete

Issue: AAH-265

This is one potential approach to AAH-265. Using access_policy to raise a 403 error on attempts for a user to delete themselves.

### DIFF
--- a/CHANGES/265.bufix
+++ b/CHANGES/265.bufix
@@ -1,0 +1,3 @@
+Users should not be able to delete themselves.
+
+Even if they have 'delete-user' perms.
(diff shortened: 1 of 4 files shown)


newswangerd:
Closing in favor of https://github.com/ansible/galaxy_ng/pull/628

PR #628, opened 2021-02-02
Users or admins cant delete admins 266

alikins:
Prevent admin users from deleting admin users

Don't allow users that are 'superuser' or are
called 'admin' to be deleted.

Add conditionals to UserViewSet policy for
'user_is_superuser' and 'user_is_admin'.

Issue: AAH-266

(Currently includes changes for AAH-265 as well, since they are needed)

### DIFF
--- a/galaxy_ng/app/access_control/access_policy.py
+++ b/galaxy_ng/app/access_control/access_policy.py
@@ -238,6 +238,9 @@ class CollectionRemoteAccessPolicy(AccessPolicyBase):
 class UserAccessPolicy(AccessPolicyBase):
     NAME = 'UserViewSet'
 
+    def is_current_user(self, request, view, action):
+        return request.user == view.get_object()
+
 
 class MyUserAccessPolicy(AccessPolicyBase):
     NAME = 'MyUserViewSet'
--- a/galaxy_ng/app/access_control/access_policy.py
+++ b/galaxy_ng/app/access_control/access_policy.py
@@ -238,6 +238,10 @@ class CollectionRemoteAccessPolicy(AccessPolicyBase):
 class UserAccessPolicy(AccessPolicyBase):
     NAME = 'UserViewSet'
 
+    def user_is_superuser(self, request, view, action):
+        user = view.get_object()
+        return user.is_superuser
+
     def is_current_user(self, request, view, action):
         return request.user == view.get_object()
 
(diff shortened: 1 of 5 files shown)


galaxy_ng/app/access_control/access_policy.py
```diff
@@ -238,6 +238,17 @@ class CollectionRemoteAccessPolicy(AccessPolicyBase):
 class UserAccessPolicy(AccessPolicyBase):
     NAME = 'UserViewSet'
 
+    def user_is_superuser(self, request, view, action):
+        user = view.get_object()
+        return user.is_superuser
+
+    def user_is_admin(self, request, view, action, ):
```
newswangerd, line 245: Being called "admin" doesn't give the user any extra privileges, so this shouldn't be here.

newswangerd (changes requested):
This looks good aside from the `user.get_username() == 'admin'` check

galaxy_ng/app/access_control/access_policy.py
alikins: I tend to agreed.

Though issue https://issues.redhat.com/browse/AAH-266 was kind of vague about what exactly an 'admin' was.

The comment at 
https://issues.redhat.com/browse/AAH-266?focusedCommentId=15727810&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-15727810

tends to suggest specifically targeting 'admin'.

But I don't think that is necessary with the typical install, since the main root style user is a django superuser. It also happens to be called 'admin' by default. The app ui doesn't provide a way to create another 'superuser' but it is easy enough to do it direct db access.

newswangerd: To me an admin is any user that has `is_super_user`.

alikins: Oh, and even if we want to specifically prevent deleting a user with username='admin', that could be done with a parameterized `is_user`  with a condition like `is_user:admin`

hendersonreed: I think what's most important is that you shouldn't be able delete a django super-user, since currently it isn't trivial to add another one. Currently I think there's a 'is_superuser' field somewhere, and we could just check that when deciding if it's possible to delete a user.

alikins: Cool. Will update to remove the is_admin condition.

alikins: Updated.

newswangerd (approved):

PR #630, opened 2021-02-03
Split proxy password field

rochacbruno:
the pulpcore database field proxy_url holds
information in form: `http://user:pass@host:port`

## Representation.

This PR splits that field in 3 different fields
only for API retrieval (GET)

proxy_address - `http://host:port`
proxy_username - `user`
proxy_password - `pass`

## Saving.

As those 3 fields doesn;t exist (yet) in pulpcore database model
in order to save we need to rejoin those 3 fields so
it get back to it regular form.

## Why?

we want to omit the `:pass` fragment from our retrieval API, but still expose user, host and port fragments

Issue: AAH-291

### DIFF
--- a/galaxy_ng/app/api/v3/serializers/sync.py
+++ b/galaxy_ng/app/api/v3/serializers/sync.py
@@ -9,6 +9,7 @@ from pulp_ansible.app.models import (
 
 from galaxy_ng.app.constants import COMMUNITY_DOMAINS
 from galaxy_ng.app.models.collectionsync import CollectionSyncTask
+from galaxy_ng.app.common.proxy_url import strip_auth_from_url, join_proxy_url
 
 
 class AnsibleDistributionSerializer(serializers.ModelSerializer):
@@ -78,6 +79,21 @@ class CollectionRemoteSerializer(LastSyncTaskMixin, pulp_viewsets.CollectionRemo
     last_sync_task = serializers.SerializerMethodField()
     created_at = serializers.DateTimeField(source='pulp_created', required=False)
     updated_at = serializers.DateTimeField(source='pulp_last_updated', required=False)
+    proxy_url = serializers.URLField(
+        help_text="The proxy url e.g: http://IP:PORT",
+        allow_null=True,
+        required=False,
+    )
+    proxy_username = serializers.CharField(
+        allow_null=True,
+        required=False
+    )
+    proxy_password = serializers.CharField(
+        allow_null=True,
+        required=False,
+        style={'input_type': 'password'},
+        write_only=True
+    )
     token = serializers.CharField(
         allow_null=True,
         required=False,
@@ -118,6 +134,8 @@ class CollectionRemoteSerializer(LastSyncTaskMixin, pulp_viewsets.CollectionRemo
             'pulp_href',
             'download_concurrency',
             'proxy_url',
+            'proxy_username',
+            'proxy_password',
         )
         extra_kwargs = {
             'name': {'read_only': True},
@@ -126,6 +144,61 @@ class CollectionRemoteSerializer(LastSyncTaskMixin, pulp_viewsets.CollectionRemo
             'client_cert': {'write_only': True},
             'ca_cert': {'write_only': True},
         }
+
+    def to_representation(self, instance):
+        """
+        Splits proxy_url field from DB in 3 fields for representation.
+        proxy_url, proxy_username, proxy_password (write-only)
+        """
+
+        data = super().to_representation(instance)
+
+        if instance.proxy_url is not None:
+            url, username, _ = strip_auth_from_url(instance.proxy_url)
+            data['proxy_url'] = url
+            data['proxy_username'] = username
+
+        return data
+
+    def save(self):
+        """
+        Rejoins 3 fields url, username, password back to proxy_url on DB.
+
+        If proxy_url is null or blank, the field is cleaned up in DB
+        regardless of username, password values.
+
+        If proxy, url, password or username is missing from the payload then
+        should be the same as the existing value in the DB or None.
+
+        If proxy username/password is set to null or blank, the fields are
+        cleaned up in DB.
+        """
+
+        empty = object()
+        proxy_url = self.validated_data.get('proxy_url', empty)
+        proxy_username = self.validated_data.get('proxy_username', empty)
+        proxy_password = self.validated_data.get('proxy_password', empty)
+
+        url_in_db, username_in_db, password_in_db = strip_auth_from_url(
+            self.instance.proxy_url
+        ) if self.instance.proxy_url else (None, None, None)
+
+        if proxy_url is empty:
+            proxy_url = url_in_db
+
+        if proxy_username is empty:
+            proxy_username = username_in_db
+
+        if proxy_password is empty:
+            proxy_password = password_in_db
+
+        self.validated_data['proxy_url'] = join_proxy_url(
+            proxy_url,
+            proxy_username,
+            proxy_password,
+        ) if proxy_url else None
+
+        super().save()
 
     def validate(self, data):
         if not data.get('requirements_file') and any(
--- a/galaxy_ng/app/common/proxy_url.py
+++ b/galaxy_ng/app/common/proxy_url.py
@@ -1,0 +1,74 @@
+"""
+This module is temporarily added to address the
+issue AAH-291.
+
+The change that this module is solving
+is going to be implemented on pulpcore
+and then this module and its references
+can be removed or replaces with pulpcore
+implementation.
+"""
+from typing import List, Tuple, Optional
+from urllib3.util import parse_url, Url
+
+
+def get_parsed_url(parsed: Url, auth_items: Optional[List[Optional[str]]]) -> str:
+    """
+    parsed: Url instance
+    auth_items: Optional list of str ["user": "password"]
+    returns: str e.g: http://user:pass@foo.bar:9999
+    """
+    return Url(
+        scheme=parsed.scheme,
+        auth=auth_items and ":".join(auth_items) or None,
+        host=parsed.host,
+        port=parsed.port,
+        path=parsed.path,
+        query=parsed.query,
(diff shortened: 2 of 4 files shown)


rochacbruno:
Todo:

- [ ] Move `to_interval_value` to `.save` because immutable querydict problem.
- [ ] Add unittests

@newswangerd This solution works, but actually brings a new problem which is controlling the empty state of the password on proxy_url.

We need to talk about it and align the changes on UI.

I am actually thinking that **we should not implement this way**, and just remove all passwords and tokens from this endpoint (keeeping only a _meta field  returning its state set|not_set).

If all fields are `not_set` then the API will accept that data on POST to create the entity, however to be able to edit those fields we can create a separate endpoint  `/sync/config/{token|password|proxy_password|cert_key}` and then UI will need to make a call to a different pull request and add different controls whenever user wants to change those passwords.

This avoid the situation where `proxy_url=http://user:pass@foo.bar:9090`, user cleans up proxy_username, but we still have password on database so the end url would end up being `http://:pass@foo.bar:9090` and other variations of this same problem.

newswangerd:
I don't want to split this into different endpoints because this is essentially a shim. Pulp is planning on moving the proxy configuration into separate database fields in the future.

Also splitting this up will make it harder to do validation. Pulp requires a token to be set if the user is changing the auth_url, and we can't validate the object as a whole if it gets broken up into several endpoints.

galaxy_ng/app/api/v3/serializers/sync.py
```diff
@@ -127,6 +144,31 @@ class Meta:
...
+
+        proxy_address = self.validated_data.get('proxy_address')
+        proxy_username = self.validated_data.get('proxy_username')
+        proxy_password = self.validated_data.get('proxy_password')
+
+        if proxy_address is not None:
+            self.validated_data['proxy_url'] = join_proxy_url(
+                proxy_address,
+                proxy_username,
+                proxy_password,  # TODO: Or default to existing password from proxy_url on DB.
```
rochacbruno, line 167: When user changes username but does't provide a password
the password is going to be set to None

to fix that we have 2 options:

- require password to be set whenever username is changed
- default password to the existing password on database

but this required us to do a database lookup here and have UI to manage the empty state (when user wants to clean up the password)

@newswangerd

newswangerd: If `proxy_password` is missing from the api request then proxy password should be the same as the existing value in the DB. If `proxy_password` is set to `null` or blank, the password should be removed.

galaxy_ng/app/common/proxy_url.py
```diff
@@ -0,0 +1,66 @@
+"""
```
newswangerd, line 1: If this isn't used anywhere else, you're probably better off just putting these functions in the serializer.

newswangerd (commented):
Looks good so far

galaxy_ng/app/api/v3/serializers/sync.py
```diff
@@ -78,6 +79,21 @@ class CollectionRemoteSerializer(LastSyncTaskMixin, pulp_viewsets.CollectionRemo
     last_sync_task = serializers.SerializerMethodField()
     created_at = serializers.DateTimeField(source='pulp_created', required=False)
     updated_at = serializers.DateTimeField(source='pulp_last_updated', required=False)
+    proxy_address = serializers.CharField(
```
newswangerd, line 82: if possible, this should be `proxy_url` so that this endpoint remains forwards compatible with pulp when pulp breaks the url up into fields.

galaxy_ng/app/common/proxy_url.py
rochacbruno: I think it is better to keep this as a separate module, because it is temporary, we will remove it anyway when pulp changes internally. Keeping those functions separated do not polute the serializer too much

galaxy_ng/app/api/v3/serializers/sync.py
```diff
@@ -127,6 +145,60 @@ class Meta:
...
+        regardless of username, password values.
+
+        If proxy, url, password or username is missing from the payload then
+        should be the same as the existing value in the DB or None.
+
+        If proxy username/password is set to null or blank, the fields are
+        cleaned up in DB.
+        """
+
+        empty = object()
```
alikins, line 176: :+1: I like the use of 'empty' as a 'unset' / sentinel value.

galaxy_ng/app/common/proxy_url.py
```diff
@@ -0,0 +1,74 @@
...
+is going to be implemented on pulpcore
+and then this module and its references
+can be removed or replaces with pulpcore
+implementation.
+"""
+from typing import List, Tuple, Optional
+from urllib3.util import parse_url, Url
+
+
+def get_parsed_url(parsed: Url, auth_items: Optional[List[Optional[str]]]) -> str:
```
alikins, line 15: Instead of relying on the order in the auth_items to determine username / password, could
usename/password be passed in as a dict? Might give some flexibility if other components get added to the proxy auth later.  Could also use a fstring for building 'auth'.
(Though, this is "temporary" and a "shim" so doesn't matter a ton...)

galaxy_ng/app/common/proxy_url.py
```diff
@@ -0,0 +1,74 @@
...
+        auth=auth_items and ":".join(auth_items) or None,
+        host=parsed.host,
+        port=parsed.port,
+        path=parsed.path,
+        query=parsed.query,
+        fragment=parsed.fragment,
+    ).url
+
+
+def strip_auth_from_url(url: str) -> Tuple[str, Optional[str], Optional[str]]:
```
alikins, line 32: Could also return a dict and join_proxy_url() could use a fstring or similar formatting to build the proxy_url.

Again, don't know if it matters if this is a temp shim.

alikins (approved):
Looks reasonable to me.

My preference might be to use dicts for the  url/username/password data instead of a list/tuple, but current approach seems fine as well.

galaxy_ng/app/api/v3/serializers/sync.py
newswangerd, line 176: Why do you send an empty object instead of just checking if the URL is none in `strip_auth_from_url`?

galaxy_ng/app/api/v3/serializers/sync.py
```diff
@@ -127,6 +145,60 @@ class Meta:
             'ca_cert': {'write_only': True},
         }
 
+    def to_representation(self, instance):
+        """
+        Splits proxy_url field from DB in 3 fields for representation.
+        proxy_url, proxy_username, proxy_password (write-only)
+        """
+
+        if instance.proxy_url is not None:
```
newswangerd, line 154: setting attributes on `instance` overwrites the properties on the underlying object, which causes issues for other methods.

I'd recommend this

```python
        data = super().to_representation(instance)

        if instance.proxy_url is not None:
            url, username, _ = strip_auth_from_url(instance.proxy_url)
            data['proxy_url'] = url
            data['proxy_username'] = username

        return data
```

newswangerd (approved):
Looks good to me aside from one minor change. If you don't want to change the `instance` thing, I can easily do it in my followup PR.

galaxy_ng/app/api/v3/serializers/sync.py
```diff
@@ -78,6 +79,21 @@ class CollectionRemoteSerializer(LastSyncTaskMixin, pulp_viewsets.CollectionRemo
     last_sync_task = serializers.SerializerMethodField()
     created_at = serializers.DateTimeField(source='pulp_created', required=False)
     updated_at = serializers.DateTimeField(source='pulp_last_updated', required=False)
+    proxy_url = serializers.CharField(
```
newswangerd, line 82: I just noticed that there is no validation on the URL. I think this used to be performed by the pulp serializer, but got overridden by this. Maybe use a URLField?

newswangerd (changes requested):

galaxy_ng/app/api/v3/serializers/sync.py
rochacbruno: @newswangerd pulp uses a CharField https://github.com/pulp/pulpcore/blob/master/pulpcore/app/serializers/repository.py#L93

the only url that is validated is `url` field https://github.com/pulp/pulp_ansible/blob/master/pulp_ansible/app/serializers.py#L136

rochacbruno: Not sure if URLField accepts user:pass and port.

```
URLField
A RegexField that validates the input against a URL matching pattern. Expects fully qualified URLs of the form http://<host>/<path>.

Corresponds to django.db.models.fields.URLField. Uses Django's django.core.validators.URLValidator for validation.

Signature: URLField(max_length=200, min_length=None, allow_blank=False)
```

galaxy_ng/app/common/proxy_url.py
rochacbruno: I preferred to keep as a sequence so I can use the same solution for when it has only `user` or both `user` and `password`

```py
":".join(["bruno"])
```

works also for

```py
":".join(["bruno", "1234"])
```

I think by its RFC there is no more components on the auth fragment.

galaxy_ng/app/common/proxy_url.py
rochacbruno: As I used the pattern `var = value if condition else other_value` it was easy to set the `other_value` as `(None, None, None)` instead of creating a whole new empty dict or more if conditions.

galaxy_ng/app/api/v3/serializers/sync.py
newswangerd: @rochacbruno  `proxy_url = serializers.URLField(...)` seems to work. It's just validating proxy_url on the serializer, not the transformed value that gets stored in the DB.

newswangerd (approved):

newswangerd:
Remove the backport-4.3 label since this was merged before we started backporting.

PR #634, opened 2021-02-08
Update requirement files to use new pip-tools format

awcrosby:
Update `requirements/` requirement files to use new pip-tools format while keeping requirement changes to a minimum


alikins (approved):
Good idea to separate the format change and the changes that will update the requirements.

PR #643, opened 2021-02-10
Upping reqs for pulp_container 2.3 (and pulp_ansible 2.7ish and pulpcore 2.10ish)

alikins:


galaxy_ng/app/api/views.py
```diff
@@ -4,7 +4,7 @@
 from rest_framework.response import Response
 from rest_framework.reverse import reverse
 
-from pulp_ansible import __version__ as pulp_ansible_version
+# from pulp_ansible import __version__ as pulp_ansible_version
```
rochacbruno, line 7:
```suggestion
try:
    from pulp_ansible import __version__ as pulp_ansible_version
except ImportError:
    from django.apps import apps
    pulp_ansible_version apps.get_app_config("ansible").version
```

setup.py
```diff
@@ -59,7 +59,7 @@ def run(self):
     "Django~=2.2.18",
     "galaxy-importer==0.2.15",
     "pulpcore<3.11,>=3.10",
-    "pulp-ansible==0.7.0",
+    "pulp-ansible",
```
rochacbruno, line 62: are we going to leave pulp_ansible unpinned?

CHANGES/275.bugfix
```diff
@@ -0,0 +1,5 @@
+Update pulp_container deps to 2.3
+
+And some adjustments to use the newer
+pulpcore 3.10, and pulp_ansible 2.7ish that
+are needed for newer pulp_container
```
awcrosby, line 5: copying comment from another pr:
> towncrier supports more than a single line in a changelog file, but I'm not sure we need to provide more than 1 line in most cases
> For this particular issue, since changelog is focused more towards users than devs, probably don't need extra info

This is a `.bugfix` but in the PR and JIRA issue I don't see a bug described

setup.py
awcrosby: I feel we should keep pulp_ansible pinned, unless we found a reason we need to change

PR #644, opened 2021-02-11
Update pulp_ansible to 0.7.0 and pulpcore 3.10

rochacbruno:
Changes needed to make pulp_ansible 0.7.0 and pulpcore 3.10 to work
on galaxy_ng.

- Added a VERSION static file
- Used django.apps.get_app_config to fetch app versions instead of
  imports
- replace `get_distro_content` with `_distro_content` because of https://github.com/pulp/pulp_ansible/commit/b1d4797c20663c8552a80b3766fde5ad287c996e
- Explicit call to self._deprecation on CollectionViewSet because of change added in https://github.com/pulp/pulp_ansible/commit/b1d4797c20663c8552a80b3766fde5ad287c996e

No-Issue

PR #645, opened 2021-02-11
Add bump2version config to setup.cfg

alikins:
Use bump2version to increment the 'version' string wherever
it is needed. The canonical source of truth for the version
is 'bump2version.current_version' in setup.cfg.

To update version, change that and run 'bump2version'.

Our to just do a 'patch' version bump, can use
'make dev/bumpversion-patch' that will rev the 'patch'
level and update all the files that use it.

- Convience 'make' targets for bumpversion
- Borrow x.y.z-{release} handling from pulp bumpversion
- Setup current version in setup.py, bumpversion will handle
- Add requirements.shipit.txt for reqs needed to build and ship a release


setup.py
```diff
@@ -67,7 +65,7 @@ def run(self):
 ]
 
 package_name = os.environ.get("GALAXY_NG_ALTERNATE_NAME", "galaxy-ng")
-version = os.environ.get("ALTERNATE_VERSION", __version__)
+version = "4.3.0.dev"
```
rochacbruno, line 68: Is there a way to read this version from `setup.cgf` ? parse it and take the `bumpversion.current_version` value?
and also keep the lookup for `ALTERNATE_VERSION` ?

cutwater: > Is there a way to read this version from setup.cgf ? 

I think yes.

> parse it and take the bumpversion.current_version value?

AFAIK bumpversion simply parses file with regexp and replaces version in it. So it should be trivial to move metadata (along with version) from `setup.py` to `setup.cfg`

rochacbruno:
@alikins @newswangerd with #644 merged, lets discuss the inclusion of `bump2version` to manage the version included in `galaxy_ng/app/VERSION` file and how it impacts our current release workflow.

alikins:
> @alikins @newswangerd with #644 merged, lets discuss the inclusion of `bump2version` to manage the version included in `galaxy_ng/app/VERSION` file and how it impacts our current release workflow.

This pr is rebased against master now, so should just be the bits to setup/user bump2version. 

(I may be missing if there is anywhere it may need to be used by the CI scripts, so yell if I have...)

CHANGES/344.misc
```diff
@@ -0,0 +1,8 @@
+Add bump2version config to setup.cfg
+
+Use bump2version to increment the 'version' string wherever
+it is needed. The canonical source of truth for the version
+is 'bump2version.current_version' in setup.cfg.
+
+To update version, change that and run 'bump2version'.
+
```
awcrosby, line 8: `towncrier` supports more than a single line in a changelog file, but I'm not sure we need to provide more than 1 line in most cases

For this particular issue, since changelog is focused more towards users than devs, probably don't need extra info

That said, it is moot overall, `.misc` is considered miscellaneous and `towncrier` does not include text in changelog just a link

requirements/requirements.shipit.txt
```diff
@@ -0,0 +1,2 @@
+bump2version
```
awcrosby, line 1: some minor concern of adding another requirements file just for release, but i guess it is does separate things

setup.cfg
```diff
@@ -0,0 +1,24 @@
...
+[bumpversion:part:release]
+optional_value = prod
+first_value = dev
+values =
+	dev
+	prod
+
+[bumpversion:file:setup.py]
+search = version = "{current_version}"
+replace = version = "{new_version}"
```
awcrosby, line 19: where does it get the `new_version` from? if we change `setup.cfg` `[bumpversion]` `current_version` from `4.3.0.dev` to `4.3.0.dev2` for example

it's nice to not have to update the version number in 2 places, but we should make sure we are not adding more complexity to remember steps of the "right" way to update version and run `bump2version`. (possible alternative with less steps to remember: CI could check version number in 2 places are the same)

Makefile
```diff
@@ -29,3 +29,12 @@ docker/build:     ## Build all development images.
...
 	./compose run api manage test galaxy_ng.tests.unit
+
+dev/bumpversion-patch:
+	bump2version --verbose patch
+
+dev/bumpversion-minor:
+	bump2version --verbose minor
+
+dev/bumpversion-major:
+	bump2version --verbose major
```
awcrosby, line 40: Could we have a single make command that is the canonical way to use bump2version? I feel most of the time we will need to specify the entire version string rather than bumping one version section. Also if `new_version` is in the config, I think we need to specify it. This worked for me:
```
bump2version --verbose --new-version=4.3.1.dev part=[]
```


We will likely be using .a# and .b# in the pre-release sections, I tried:
```
bump2version --verbose --new-version=4.3.0.a1 part=[]
```
and `setup.cfg` was changed as expected but `setup.py` and `__init__.py` had invalid version strings: `version = "version = "4.3.0.a""` `__version__ = "__version__ = "4.3.0.a""`

awcrosby (changes requested):

requirements/requirements.shipit.txt
alikins: Agreed. I'm not sure the best place for the requirement. I mostly want to include it in a requirements file _somewhere_ for automation at some point in the future.

pulpcore and other pulp plugins that use the plugin-template stuff only specify it as a requirement
in the .ci stuff  (https://github.com/pulp/pulpcore/blob/master/.ci/assets/release_requirements.txt ) but
we aren't ported to that yet.

We don't have a general purpose requirements.txt either. Since it isn't needed at app runtime, just at devel/release time and we don't have a dev_requirements.txt etc either.

setup.cfg
alikins: `new_version` is either:

-  from autoincrementing `current_version`  when used with 'bump2version minor' for example.
- specified at the cli with '--new-version 0.6.1' for example.

General, you don't change `current_version` manually, bumpversion itself will increment it to what `new_version` is.
`current_version` is used to build the patterns to find where the `new_version` will be substituted.

Conceptually, `make dev/bumpversion-patch` etc is the 'right' way to bump the version[1]. It then creates `new_version` based on `current_version` and updates all the places where the version needs to be specified.

[1] Or manually running 'bump2version'

Makefile
alikins: `new_version` is generated, so generally no need to specify it. Unless you want/need to bump to an arbitrary version (say, from 4.2.0 -> 5.1.1 etc).

To use '.a#' or .b#' for pre releases, we would need to define that pre-release 'part' ('patch', 'minor', 'major', are default parts and 'release' is defined in the current setup.cfg).

Then you would use 'bumpversion pre-release'. The details of the format are specified with the parse/serialize config for that part.

https://github.com/c4urself/bump2version#optional_value-  has more details. The current config knows about an
optional 'release' for cases like 4.3.0.dev. This is based on what pulp does. 

Haven't tested, but something roughly like:

**UNTESTED** 

``` ini
[bumpversion]
current_version = "4.3.0.dev"
commit = True
tag = True
parse = (?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)(\.(?P<release>[a-z]+))(?P<build>\d+)?
serialize =
	{major}.{minor}.{patch}.{release}{build}
	{major}.{minor}.{patch}

[bumpversion:part:release]
optional_value = stable
first_value = a
values =
	a
        b
	stable

[bumpversion:part:build]
optional_value = 0
first_value = 1

[bumpversion:file:setup.py]
search = version = "{current_version}"
replace = version = "{new_version}"
```

alikins: And 'bumpversion build' to go from 5.3.7.a1 -> 5.3.7.a2
and 'bumpversion release' to from 5.3.7.a1 -> 5.3.7.b1
and another 'bumpversion release' to from from 5.3.7.b1 -> 5.3.7

I think... untested caveat.

alikins: Oh, and the main reason for using the Makefile targets is just to have something known. And for me, so I can tab complete 'make dev/<tab>' and figure the command instead of looking it up.

alikins: I updated this with a config that should handle the cases you mentioned.

setup.cfg
```diff
@@ -0,0 +1,25 @@
...
+values =
+    stable
+    dev
+    a
+    b
+
+[bumpversion:part:build]
+first_value = 0
+
+[bumpversion:file:setup.py]
```
rochacbruno, line 23: Right now the only place where we have version explicitly set is 
https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/VERSION

Can bump2version manage that static file instead of putting version back to setup.py and `__init__.py` ?

alikins:
oh, and a regex tester url for this is https://regex101.com/r/cToWQK/1
(Could include that in a comment in setup.cfg?)

Makefile
alikins: Also updated so that galaxy_ng/__init__.py and galaxy_ng/app/__init__.py are updated.

awcrosby: That sounds good. Can we add something like this in the Makefile?, those examples will be helpful when using in the future

```
# 'bumpversion build' to go from 5.3.7.a1 -> 5.3.7.a2
dev/bumpversion-build:
	bump2version --verbose build

# 'bumpversion release' to from 5.3.7.a1 -> 5.3.7.b1
# another 'bumpversion release' to from from 5.3.7.b1 -> 5.3.7
dev/bumpversion-release:
	bump2version --verbose release
```

setup.cfg
awcrosby: yeah, feels like bump2version should update VERSION unless there is a reason to add it back in the other 2 files.

bump2version will really shine when the UI gets absorbed into this repo and there will be additional version strings scattered around

rochacbruno: And if we go with putting versions back on .py files, then VERSION file must be deleted and removed from MANIFEST.in

alikins: Using galaxy_ng/app/VERSION requires setup.py and galaxy_ng/app/__init__.py to read the version number at runtime. If we make those versions static with bump2version, that's two less runtime couplings that can break.

If we want to move more of the setup.py data into setup.cfg[1] (seems to be currently suggested approach by the docs), then setup.cfg can't run python code to read in the version.

[1] or pyproject.toml or whatever

rochacbruno: @alikins 

good points

1) setup.py already needs to read some static files and also find packages, so there is no problem to have it loading one more.

2) for `__init__`  you have a good point, we can change the code there, as setup.py outputs the version already read and parsed we can do.

```py
import pkg_resources
pkg_resources.get_distribution("galaxy_ng").version
```

awcrosby: @rochacbruno do any of the pulp bits require a `galaxy_ng/app/VERSION` be present?

awcrosby (approved):
Added one question about VERSION, other than that, LGTM

setup.cfg
rochacbruno: @awcrosby no, pulp doesn't require it, pulp only requires version to be defined in the `AppConfig` class. VERSION was included just to ensure we have version written in a single place and that it allows us to read it from anywhere even if package is not installed.

We need to read the version in `setup.py` in `/galaxy_ng/app/__init__.py` and in `galaxy_ng/app/api/views.py` so to avoid duplicating it in multiple files and the need to import it, we added a static file.

Other solution would be having it **only** on `setup.py` and in other places use `pkg_resources.get_distribution("galaxy_ng").version` but that requires `galaxy_ng` package to be always installed.

rochacbruno (approved):

PR #646, opened 2021-02-12
Add build_deploy.sh and pr_check.sh scripts

cutwater:
These scripts are part of the new CI pipeline and are needed
to build and push container images to Quay.

### DIFF
--- a/build_deploy.sh
+++ b/build_deploy.sh
@@ -1,0 +1,36 @@
+#!/bin/bash
+
+set -o errexit
+set -o nounset
+set -o verbose
+
+
+readonly DOCKER_CONF="${PWD}/.docker"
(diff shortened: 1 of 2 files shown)


rochacbruno (approved):

awcrosby (approved):
LGTM

I'm not sure the method that we currently deploy to quay, will that be removed after we fully move to the new pipeline?

bmclaughlin (approved):

cutwater:
> LGTM
> 
> I'm not sure the method that we currently deploy to quay, will that be removed after we fully move to the new pipeline?

@awcrosby Right now images are built in the `buildfactory` project in old OCPv3 and then pushed to Quay.
Once app-interface Jenkins jobs are configured to use these scripts, I'll disable old quay-copiers from e2e-deploy.

PR #648, opened 2021-02-14
Add Unpaginated Collection and Metadata endpoints for better sync performance

rochacbruno:
Issue: AAH-224

## Content Sync Performance Feature

- Added v3/collections/all for UnpaginatedCollections
- Added v3/collection_versions/all for UnpaginatedCollectionVersions
- Added v3/ for RepoMetadata

### Collections/all
collections_all

### collection_versions/ll
versions_all

### v3/
metadata

### DIFF
--- a/galaxy_ng/app/api/v3/serializers/collection.py
+++ b/galaxy_ng/app/api/v3/serializers/collection.py
@@ -14,6 +14,7 @@ from pulp_ansible.app.galaxy.v3.serializers import (
     CollectionSerializer as _CollectionSerializer,
     CollectionRefSerializer as _CollectionRefSerializer,
     CollectionVersionSerializer as _CollectionVersionSerializer,
+    UnpaginatedCollectionVersionSerializer as _UnpaginatedCollectionVersionSerializer,
     CollectionVersionListSerializer as _CollectionVersionListSerializer,
 )
 
@@ -26,10 +27,12 @@ class HrefNamespaceMixin:
 
         view_namespace = self.context["view_namespace"]
 
-        # Handle the case where the /api/automation-hub/v3/collections/  urls do not
+        # Handle the case where the /api/automation-hub/v3/collections/ urls do not
         # include a "<str:path>" path param. ie, the bacwares compatible default "golden" repo
-        if "/<str:path>/v3/" not in self.context["view_route"]:
-            del kwargs["path"]
+        if "/<str:path>/v3/" in self.context["view_route"]:
+            kwargs.setdefault("path", self.context["path"])
+        else:
+            kwargs.pop("path", None)
 
         return reverse(
             f"{view_namespace}:{url_name}",
@@ -42,25 +45,19 @@ class CollectionSerializer(_CollectionSerializer, HrefNamespaceMixin):
         ref_name = "CollectionWithFixedHrefsSerializer"
 
     def get_href(self, obj):
-        """Get href."""
-        kwargs = {"path": self.context["path"], "namespace": obj.namespace, "name": obj.name}
-        return self._get_href("collections-detail", **kwargs)
+        return self._get_href("collections-detail", namespace=obj.namespace, name=obj.name)
 
     def get_versions_url(self, obj):
         """Get a link to a collection versions list."""
-        kwargs = {"path": self.context["path"], "namespace": obj.namespace, "name": obj.name}
-        return self._get_href("collection-versions-list", **kwargs)
+        return self._get_href("collection-versions-list", namespace=obj.namespace, name=obj.name)
 
     def get_highest_version(self, obj):
         """Get a highest version and its link."""
         data = super().get_highest_version(obj)
-        kwargs = {
-            "path": self.context["path"],
-            "namespace": obj.namespace,
-            "name": obj.name,
-            "version": data['version'],
-        }
-        data['href'] = self._get_href("collection-versions-detail", **kwargs)
+        data['href'] = self._get_href(
+            "collection-versions-detail",
+            namespace=obj.namespace, name=obj.name, version=data['version']
+        )
         return data
 
 
@@ -70,9 +67,7 @@ class CollectionRefSerializer(_CollectionRefSerializer, HrefNamespaceMixin):
 
     def get_href(self, obj):
         """Returns link to a collection."""
-
-        kwargs = {"path": self.context["path"], "namespace": obj.namespace, "name": obj.name}
-        return self._get_href("collections-detail", **kwargs)
+        return self._get_href("collections-detail", namespace=obj.namespace, name=obj.name)
 
 
 class CollectionVersionListSerializer(_CollectionVersionListSerializer, HrefNamespaceMixin):
@@ -81,14 +76,21 @@ class CollectionVersionListSerializer(_CollectionVersionListSerializer, HrefName
 
     def get_href(self, obj):
         """Get href."""
-        kwargs = {
-            "path": self.context["path"],
-            "namespace": obj.namespace,
-            "name": obj.name,
-            "version": obj.version,
-        }
+        return self._get_href(
+            "collection-versions-detail",
+            namespace=obj.namespace, name=obj.name, version=obj.version
+        )
 
-        return self._get_href("collection-versions-detail", **kwargs)
+
+class UnpaginatedCollectionVersionSerializer(_UnpaginatedCollectionVersionSerializer, HrefNamespaceMixin):
+
+    collection = CollectionRefSerializer(read_only=True)
+
+    class Meta(_UnpaginatedCollectionVersionSerializer.Meta):
+        ref_name = "UnpaginatedCollectionVersionWithFixedHrefsSerializer"
+
+    def get_href(self, obj):
+        return self._get_href("collections-detail", namespace=obj.namespace, name=obj.name)
 
 
 class CollectionVersionSerializer(_CollectionVersionSerializer, HrefNamespaceMixin):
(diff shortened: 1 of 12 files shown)


galaxy_ng/app/api/v3/viewsets/collection.py
```diff
@@ -52,11 +52,6 @@ class ViewNamespaceSerializerContextMixin:
     def get_serializer_context(self):
         """Inserts distribution path to a serializer context."""
 
-        self._deprecation
```
rochacbruno, line 55: @fao89 looks like this call is needed only when we have no collections, if my db is empty and I try to call `/collections` I will get an error here. After a sync the error does not occur, as this field is populated on the filter I think that happens because empty queryset does not invoke the filters.

galaxy_ng/app/api/v3/views/sync.py
```diff
@@ -47,7 +47,8 @@ def post(self, request: Request, *args, **kwargs) -> Response:
             kwargs={
                 "remote_pk": remote.pk,
                 "repository_pk": distro.repository.pk,
-                "mirror": True
+                "mirror": True,
+                "optimize": True,
```
rochacbruno, line 51: @bmbouter is it right ot always pass `optimize=True` as this attribute is not optional.

galaxy_ng/app/api/v3/viewsets/collection.py
```diff
@@ -107,23 +114,27 @@ def list(self, request, *args, **kwargs):
...
-        artifact = ContentArtifact.objects.get(content=instance)
 
-        context = self.get_serializer_context()
-        context["content_artifact"] = artifact
+class UnpaginatedCollectionVersionViewSet(api_base.LocalSettingsMixin,
+                                          ViewNamespaceSerializerContextMixin,
+                                          CollectionListMixin,
+                                          pulp_ansible_views.UnpaginatedCollectionVersionViewSet):
+    pagination_class = None
+    serializer_class = CollectionVersionSerializer
```
rochacbruno, line 123: @fao89 do we need a custom serializer here? I am asking because the default is returning 114000 records with only 3 collections see; https://gist.github.com/rochacbruno/031d624d0356e57fa0d0cb3f6b54b978

galaxy_ng/app/api/v3/views/sync.py
fao89: the default is true, you don't need to pass it, we set it at serializer level: https://github.com/pulp/pulp_ansible/blob/master/pulp_ansible/app/serializers.py#L101

galaxy_ng/app/api/v3/viewsets/collection.py
fao89: it is the right serializer, it exposes all versions from all collections, and I could see `vmware.vmware_rest` has 68 versions.
Beyond that, we are exposing `MANIFEST.json` and `FILES.json` which made the response bigger.
But I'm seeing manifest field repeating data from other fields, we may need to work on this, FYI @bmbouter @daviddavis

galaxy_ng/app/api/v3/views/sync.py
rochacbruno: Ok, so in our case we must provide it to enqueue the task.

galaxy_ng/app/api/v3/viewsets/collection.py
newswangerd: I don't think it's reasonable to include the `MANIFEST.json` and `FILES.json` as part of this API endpoint. That's just way too much data for even a medium sized deployment like hub on cloud.redhat.com.

`FILES.json` is especially problematic. Collections don't have a ignore file and it's very possible that someone could accidentally include their `node_modules` directory in a collection which would add something like 100,000 files for a single collection version.

newswangerd: @fao89 @bmbouter are manifest and files required for sync?

newswangerd: @rochacbruno tells me that `files` was added for the `ansible-galaxy` client. Can @jborean93, @fao89 and @bmbouter  let me know what this is being used for?

bmbouter: I believe we can remove the `files` and `manifest` data from this metadata API. The sync pipeline populates that data during sync for items it needs to freshly fetch here: https://github.com/pulp/pulp_ansible/blob/master/pulp_ansible/app/tasks/collections.py#L842-L844

Can someone open a bugfix request for that against pulp_ansible?

I believe these are to enable a CLI search command at some point to help solve the problem of "where did my content go?" @jborean93 would need to give more details on the plan and timeline tho.

fao89: it was an old request https://pulp.plan.io/issues/7572 
some background here: https://github.com/pulp/pulp_ansible/pull/505#issuecomment-771903712
@newswangerd you have really valid points, and I can see it was not well designed, maybe we need to move it to another endpoint as docs-blob case, or make a parametrizable endpoint as @rochacbruno suggested

newswangerd: I calculated that exposing the metadata without files would return about 150mb of data with 100,000 collection versions, which isn't unreasonable with proper caching and streamed API responses, so I'm fine with keeping the metadata in, if it's something that will improve the client.

The request in pulp/pulp_ansible#505 (comment) seems reasonable, if we can isolate the exact set of collection information that the client needs, rather than just returning everything in metadata and files.

bmbouter: Overall I am in favor of removing it if the only thing using it would be pulp's sync code. Pulp's sync code doesn't even use it. We need input from @jborean93 on if the CLI needs this in the unpaginated metadata response or not.

fao89: I believe for the CLI needs, we can keep it only on retrieve, and have another serializer for unpaginated list

jborean93: There's no plans or timelines in place to add the `provides` command to the `ansible-galaxy` cli. Maybe someone will work on it in the 2.12 development cycle but nothing has been planned or talked about yet. One of the problems with adding a new command that relies on new APIs is that it will only work for v3 endpoints leaving galaxy.ansible.com out in the cold. This might make a sales proposition for AAP/AH but IMO there is little desire to implement a feature that will not work for galaxy.

As for the API itself and what data returns there are 2 issues that seem to be at play here

* Getting the version list for a collection can take some time, as time goes on there will be more versions added requiring even more API calls
    * This certainly doesn't need the collection content, just a collection with their versions. I'm not sure if we enumerate the deps for all versions or just once we select the candidate
* Getting collection content information for a `provides` like command
    * This certainly requires information about the collections plugins.
    * Doesn't necessarily have to be the `FILES.json` just a list of plugins, their type, and maybe description

I'm no API expert so I don't know of an efficient way of exposing this information but something must be out there if tools like `dnf` can do this.

rochacbruno:
Depending on https://github.com/pulp/pulp_ansible/pull/540 to be ready for review

galaxy_ng/tests/unit/api/test_api_v3_collections.py
```diff
@@ -127,8 +188,37 @@ def test_collection_versions_list(self):
...
+
+    #     # Ensure some fields are in
+    #     for field in ('metadata', 'namespace', 'name', 'artifact'):
+    #         with self.subTest(field=field):
+    #             self.assertIn(field, response.data[0])
+
+    #     # Ensure some fields are not in
+    #     for field in ('manifest', 'files'):
+    #         with self.subTest(field=field):
+    #             self.assertNotIn(field, response.data[0])
```
rochacbruno, line 224: **NOTE** to enable those tests we have to include `assets/` and import a real collection, or create a mocked `ContentArtifact`. This probably will be part of a funtional `sync` test case later, left here as reference.

galaxy_ng/app/api/v3/serializers/collection.py
```diff
@@ -26,10 +27,12 @@ def _get_href(self, url_name, **kwargs):
...
 
-        # Handle the case where the /api/automation-hub/v3/collections/  urls do not
+        # Handle the case where the /api/automation-hub/v3/collections/ urls do not
         # include a "<str:path>" path param. ie, the bacwares compatible default "golden" repo
-        if "/<str:path>/v3/" not in self.context["view_route"]:
-            del kwargs["path"]
+        if "/<str:path>/v3/" in self.context["view_route"]:
+            kwargs.setdefault("path", self.context["path"])
+        else:
+            kwargs.pop("path", None)
```
alikins, line 35: Unrelated to this pr and not something that needs to be addressed now, but eventually I'd like to change 'path' to something more specific  (ie, 'distro_base_path' or slightly inaccurately 'repo_name' or 'repo_path')

galaxy_ng/app/api/v3/urls.py
```diff
@@ -31,6 +31,17 @@
 ]
 
 urlpatterns = [
+    path("", viewsets.RepoMetadataViewSet.as_view({"get": "retrieve"}), name="repo-metadata"),
+    path(
+        'collections/all/',
+        viewsets.UnpaginatedCollectionViewSet.as_view({'get': 'list'}),
+        name='all-collections-list'
+    ),
+    path(
```
alikins, line 40: Design question, could the unpaginated collection list viewsets be implemented with a query param to CollectionViewSet and CollectionVersionViewSet?

ie, in lieu of new viewsets.

I think either is fine, but was curious if there was a reason for one or the other.

rochacbruno: @alikins we just followed what `pulp_ansible` implemented to keep our side conforming with that. I suggested having a dynamic endpoint e,g: `collections/?per_page=all&exclude=manifest&exclude=files` but the way it is was already implemented.

I also think that a dynamic endpoint would be hard to implement and test.

@fao89 any input?

alikins (approved):
I think it looks good to me as is.  Have a few questions/comments I'm curious about, but that's just for extra context.

pr as is LGTM

newswangerd (approved):
lgtm

galaxy_ng/app/api/v3/urls.py
fao89: no special reason, when I was implementing I didn't think about a query_param to change the pagination, but as @rochacbruno mentioned it could be hard to test, `CollectionViewSet` is getting bigger and bigger

PR #649, opened 2021-02-15
Add docker registry token auth configuration to dev environment.

newswangerd:
Add the pulp_container token auth settings: https://docs.pulpproject.org/pulp_container/workflows/authentication.html 

This allows for the `docker` and `podman` clients to correctly authenticate.


dev/common/container_auth_private_key.pem
```diff
@@ -0,0 +1,5 @@
+-----BEGIN EC PRIVATE KEY-----
```
newswangerd, line 1: These keys are just for the developer environment, so it's okay for them to get committed.

rochacbruno (approved):

PR #650, opened 2021-02-16
Update to pulp_container 2.3.1

newswangerd:
2.3.1 includes this fix: https://github.com/pulp/pulp_container/pull/236 which was preventing us from using the pulp_container registry alongside galaxy_ng


alikins (approved):
LGTM

PR #660, opened 2021-02-23
Fixing pulpcore 3.10+ breaking changes and renamings

rochacbruno:
Pulpcore 3.10 introduced some breaking changes

- On serializers `highest_versions` has been removed out of the context
- On download_url `content_artifact` has been also moved out of context
- Sync now requires `optimize=True` attribute

And also some non-breaking changes

- Viewsets `retrieve` is not needed because distro_path is now part of pulp-ansible and can be removed from our side
- `self._deprecation` is not necessary (included by mistake on a
  previous PR)

Issue: AAH-393

### DIFF
--- a/galaxy_ng/app/api/v3/serializers/collection.py
+++ b/galaxy_ng/app/api/v3/serializers/collection.py
@@ -53,15 +53,15 @@ class CollectionSerializer(_CollectionSerializer, HrefNamespaceMixin):
 
     def get_highest_version(self, obj):
         """Get a highest version and its link."""
-        collection = self.context["highest_versions"][obj.pk]
+        data = super().get_highest_version(obj)
         kwargs = {
             "path": self.context["path"],
-            "namespace": collection.namespace,
-            "name": collection.name,
-            "version": collection.version,
+            "namespace": obj.namespace,
+            "name": obj.name,
+            "version": data['version'],
         }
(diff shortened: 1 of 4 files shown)


galaxy_ng/app/api/v3/serializers/collection.py
```diff
@@ -53,15 +53,15 @@ def get_versions_url(self, obj):
...
-            "name": collection.name,
-            "version": collection.version,
+            "namespace": obj.namespace,
+            "name": obj.name,
+            "version": data['version'],
         }
-        href = self._get_href("collection-versions-detail", **kwargs)
-        return {"href": href, "version": str(collection.version)}
+        data['href'] = self._get_href("collection-versions-detail", **kwargs)
+        return data
```
awcrosby, line 64: probably doesn't matter, but does `data` dictionary now return more than just `href` and `version` keys?

rochacbruno: @awcrosby no, on pulp_ansible it returns only href and version

https://github.com/pulp/pulp_ansible/blob/04df5e5ad059f39c5c2c8cd37d8472040af22611/pulp_ansible/app/galaxy/v3/serializers.py#L82

awcrosby (approved):

PR #665, opened 2021-03-03
[PoC] Disable synclist curation.

newswangerd:
This disables synclist curation and fakes synclist repositories by returning them as a dynamic filter instead of as the contents of a repository.


chouseknecht:
Closing due to inactivity.

PR #666, opened 2021-03-04
Reduce excessive number of spawned synclist tasks

cutwater:
Fix large number of enqueued synclist related tasks by doing the
following changes:
  * Call `add_and_remove` tasks directly.
  * Call `curate_synclist_repository` task in batches.
  * Batch size is set to 200.

### DIFF
--- a/galaxy_ng/app/api/ui/viewsets/my_synclist.py
+++ b/galaxy_ng/app/api/ui/viewsets/my_synclist.py
@@ -1,5 +1,6 @@
 import logging
 
+from django.shortcuts import get_object_or_404
 from guardian.shortcuts import get_objects_for_user
 
 from rest_framework.decorators import action
@@ -35,13 +36,12 @@ class MySyncListViewSet(SyncListViewSet):
         )
 
     @action(detail=True, methods=["post"])
-    def curate(self, request, pk=None):
-        locks = [pk]
-        task_args = (pk,)
-        task_kwargs = {}
-
+    def curate(self, request, pk):
+        synclist = get_object_or_404(models.SyncList, pk)
         synclist_task = enqueue_with_reservation(
-            curate_synclist_repository, locks, args=task_args, kwargs=task_kwargs
+            curate_synclist_repository,
+            resources=[synclist.repository],
+            args=(pk, )
         )
 
         log.debug("synclist_task: %s", synclist_task)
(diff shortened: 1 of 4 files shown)


galaxy_ng/app/api/ui/viewsets/my_synclist.py
```diff
@@ -35,13 +35,12 @@ def get_queryset(self):
         )
 
     @action(detail=True, methods=["post"])
-    def curate(self, request, pk=None):
-        locks = [pk]
-        task_args = (pk,)
-        task_kwargs = {}
-
+    def curate(self, request, pk):
+        synclist = models.SyncList.objects.get(pk)
```
newswangerd, line 39: maybe `get_object_or_404` or something similar

galaxy_ng/app/tasks/synclist.py
```diff
@@ -20,7 +19,8 @@
 from galaxy_ng.app import models
 
 log = logging.getLogger(__name__)
-pf = pprint.pformat
+
+SYNCLIST_BATCH_SIZE = 200
```
newswangerd, line 23: should this be configurable in settings.py? That way we can tweak it without a code deployment

bmbouter:
From a high level this contains all the aspects I was hoping for both batching via the introduction of an intermediary task, and the removal of the extra `enqueue_with_reservation`. :+1:

alikins (approved):
Seems okay to me. It reduces the granularity of the tasks, but that is the intention. 
Only concern is it is has any impact on error handling. In particular, if errors raised from curate_synclist_repository() lose any context. I haven't tested it, and can't point to any
cases where it could cause a problem, so assuming it should be okay.

newswangerd (approved):
LGTM

galaxy_ng/app/tasks/synclist.py
```diff
@@ -94,6 +95,11 @@ def curate_all_synclist_repository(upstream_repository_name, **kwargs):
     task_group.finish()
 
 
+def curate_synclist_repository_batch(synclist_pks, **kwargs):
+    for synclist_pk in synclist_pks:
+        curate_synclist_repository(synclist_pk=synclist_pk)
```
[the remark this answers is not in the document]
alikins: So a curate_synclist_repository_batch() will end up running entirely on one worker?

Will it block if any curate_synclist_repository() call blocks?

PR #667, opened 2021-03-05
Add make targets for common tasks on dev env and support for env snapshots

rochacbruno:
## This PR is addresses AAH-387 adding more targets to Makefile

```make
❯ make         
Usage: make <target>

Targets:
help:             ## Show the help.
requirements:     ## Update python dependencies lock files (i.e. requirements.txt).
changelog:        ## Build the changelog
docker/build:     ## Build all development images.
docker/test:      ## Run unit tests.
docker/loaddata:  ## Load initial data from fixtures
docker/migrate:   ## Run django migrations
docker/resetdb:   ## Cleans database
api/get:          ## Make an api get request using 'httpie'
api/shell:        ## Opens django management shell in api container
api/bash:         ## Opens bash session in the api container
api/runserver:    ## Runs api using django webserver for debugging

```

## Also some tweaks to Docker files and setup.py to enable running specific
deps branches.

- if `LOCK_REQUIREMENTS` is disabled, all deps listed on `DEV_SOURCE_PATH` is unppined on `setup.py`

## Adds ability to set a global `suffix` so it is possible to have multiple
environments already built and saved (snapshots)

**Build and run environment to work on specific issue**
```bash
./compose build
./compose up
```

> **OMG** in the middle of my task I have to work on another issue/hotfix, do I need to rebuild my environment? lost my current state? .. No.

**Build and run environment to work on specific issue**
```bash
./compose stop
export DEV_IMAGE_SUFFIX=fixing_an_urgent_issue 
./compose build
./compose up
```
Now you can switch the state of `DEV_IMAGE_SUFFIX` to use whichever environment state you want, (a.k.a snapshot)

Issue: AAH-387

### DIFF
--- a/compose
+++ b/compose
@@ -21,9 +21,9 @@ EOF
 fi
 
 echo "INFO: Using compose profile ${COMPOSE_PROFILE}"
-echo "INFO: Installing from source ${DEV_SOURCE_PATH}"
-echo "INFO: Image suffix ${DEV_IMAGE_SUFFIX}"
-echo "INFO: Volume suffix ${DEV_VOLUME_SUFFIX}"
+echo "INFO: ${DEV_SOURCE_PATH:-No} packages installed from source"
+echo "INFO: Image suffix ${DEV_IMAGE_SUFFIX:-is unset}"
+echo "INFO: Volume suffix ${DEV_VOLUME_SUFFIX:-${DEV_IMAGE_SUFFIX:-is unset}}"
 
 compose_args=(
   -f 'dev/docker-compose.yml'
@@ -48,7 +48,7 @@ declare -xr COMPOSE_CONTEXT=".."
 declare -xr LOCK_REQUIREMENTS="${LOCK_REQUIREMENTS:-1}"
 declare -xr COMPOSE_PROFILE="${COMPOSE_PROFILE}"
 declare -xr DEV_IMAGE_SUFFIX="${DEV_IMAGE_SUFFIX:-}"
-declare -xr DEV_VOLUME_SUFFIX="${DEV_VOLUME_SUFFIX:-$DEV_IMAGE_SUFFIX}"
+declare -xr DEV_VOLUME_SUFFIX="${DEV_VOLUME_SUFFIX:-${DEV_IMAGE_SUFFIX}}"
 declare -xr COMPOSE_PROJECT_NAME="${COMPOSE_PROJECT_NAME:-galaxy_ng${DEV_IMAGE_SUFFIX:-}}"
 
 exec docker-compose "${compose_args[@]}" "$@"
--- a/compose
+++ b/compose
@@ -5,7 +5,8 @@ set -o errexit
 
 if [[ -f '.compose.env' ]]; then
   set -o allexport
-  source .compose.env
+  # avoid override already set variables
+  source <(grep -v '^#' .compose.env | sed -E 's|^(.+)=(.*)$|: ${\1=\2}; export \1|g')
   set +o allexport
 fi
 
@@ -20,6 +21,9 @@ EOF
 fi
 
 echo "INFO: Using compose profile ${COMPOSE_PROFILE}"
+echo "INFO: Installing from source ${DEV_SOURCE_PATH}"
+echo "INFO: Image suffix ${DEV_IMAGE_SUFFIX}"
+echo "INFO: Volume suffix ${DEV_VOLUME_SUFFIX}"
 
 compose_args=(
   -f 'dev/docker-compose.yml'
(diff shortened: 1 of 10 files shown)


setup.py
```diff
@@ -65,6 +66,29 @@ def run(self):
...
+    """
+    DEV_SOURCE_PATH = os.environ.get("DEV_SOURCE_PATH", "").split(":")
+    DEV_SOURCE_PATH += [path.replace("_", "-") for path in DEV_SOURCE_PATH]
+    requirements = [
+        re.sub(r"""(=|^|~|<|>|!)([\S]+)""", "", req)
+        if req.lower().startswith(tuple(DEV_SOURCE_PATH)) else req
+        for req in requirements
+    ]
+    print("Installing with unpinned DEV_SOURCE_PATH requirements", requirements)
+
```
alikins, line 91: This is possibly somewhere we could make use of 'constraint.txt' files.  If memory servers, for our setup
we would have a requirements.txt that is all the unpinned deps. A dev build would use only requirements.txt.

For a compose profile that needs to pin versions, there would be a 'requirements/constraints.standalone.in' and .txt.
`requirements.standalone.in` would provide more specific version requires, and then  `requirements/constraints.standalone.txt` would be the resolved fully pinned requirements. 

Places where we use `pip -r` in the compose and Dockerfiles would need to use '--constraints' instead.

dev/docker-compose.yml
```diff
@@ -83,14 +96,15 @@ services:
...
-    image: "localhost/galaxy_ng/galaxy_ng:${COMPOSE_PROFILE}"
+    image: "localhost/galaxy_ng/galaxy_ng:${COMPOSE_PROFILE}${DEV_IMAGE_SUFFIX:-}"
     command: ['run', 'content-app']
     ports:
       - "24816:24816"
     environment:
       - "WITH_DEV_INSTALL=1"
       - "LOCK_REQUIREMENTS=${LOCK_REQUIREMENTS}"
       - "DEV_SOURCE_PATH=${DEV_SOURCE_PATH}"
+      - "COMPOSE_PROFILE=${COMPOSE_PROFILE}"
```
alikins, line 107: +1 for passing the COMPOSE_PROFILE name into the container. Useful for troubleshooting if nothing else.

dev/Dockerfile.base
```diff
@@ -50,9 +57,12 @@ RUN set -ex; \
 # Install application
 COPY . /app
 
+# When LOCK_REQUIREMENTS is disabled avoid running collectstatic here
+# on that case developer should run collectstatic manually
```
alikins, line 61: 'collectionstatic' will also almost always fail if there are any deps issues iirc.

compose
```diff
@@ -40,8 +44,11 @@ else
 fi
 
 declare -xr DEV_SOURCE_PATH=${DEV_SOURCE_PATH:-galaxy_ng}
-declare -xr COMPOSE_PROJECT_NAME="${COMPOSE_PROJECT_NAME:-galaxy_ng}"
 declare -xr COMPOSE_CONTEXT=".."
 declare -xr LOCK_REQUIREMENTS="${LOCK_REQUIREMENTS:-1}"
+declare -xr COMPOSE_PROFILE="${COMPOSE_PROFILE}"
+declare -xr DEV_IMAGE_SUFFIX="${DEV_IMAGE_SUFFIX:-}"
+declare -xr DEV_VOLUME_SUFFIX="${DEV_VOLUME_SUFFIX:-$DEV_IMAGE_SUFFIX}"
+declare -xr COMPOSE_PROJECT_NAME="${COMPOSE_PROJECT_NAME:-galaxy_ng${DEV_IMAGE_SUFFIX:-}}"
```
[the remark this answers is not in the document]
rochacbruno: @alikins I am working on making it possible to select a different range of ports for a second run.

setup.py
rochacbruno: @alikins the only problem is that we still need `setup.py` to hold pinned deps, so releases are based only on setup.py list, this is what is going to the RPMs build.

We can read a requirements file from setup.py, and also it is possible to make it dynamic as long as we include it as a static file on MANIFEST.in

Probably at some point we might want to move to pyproject.toml / Poetry ?

newswangerd:
`make changelog` would be really nice. This is how the changelog is built: https://github.com/ansible/galaxy_ng/wiki/Updating-the-Change-Log

I think it just has to run the `towncrier` command.

cutwater (approved):
lgtm, nice job

compose
```diff
@@ -20,6 +21,9 @@ EOF
 fi
 
 echo "INFO: Using compose profile ${COMPOSE_PROFILE}"
+echo "INFO: Installing from source ${DEV_SOURCE_PATH}"
+echo "INFO: Image suffix ${DEV_IMAGE_SUFFIX}"
+echo "INFO: Volume suffix ${DEV_VOLUME_SUFFIX}"
```
cutwater, line 26: I'd be nice to print debug log to `stderr`.

rochacbruno: @cutwater what is the advantage of routing it to stderr?

ironfroggy: Could `DEV_IMAGE_SUFFIX` and `DEV_VOLUME_SUFFIX` be given default values?

ironfroggy (changes requested):

compose
rochacbruno: @ironfroggy it already defaults to Blank Value so it keeps compatibility with current environments that has no suffix.

alikins (approved):
Looks good to me.

Assuming it works for an existing checking without any manual intervention. If it does need some manual intervention, need some docs about it (I may have missed them).

ZitaNemeckova:
@rochacbruno I got this when running anything `./compose`
```
ERROR: The $COMPOSE_PROFILE environment variable MUST be set.
       The the easiest way to to this is to create a .compose.env file by running:

           $ cp .compose.env.example .compose.env
```

I tried `$ cp .compose.env.example .compose.env` but it didn't help. So I guess it needs more fool-proof instructions for people like me :D

PR #672, opened 2021-03-09
Add OpenShift job template to run migrations

cutwater:
Issue: AAH-145


alikins (approved):
Not sure how to test that, but looks okay to me.

cutwater:
@alikins the only way to test so far it is to deploy it to stage

PR #687, opened 2021-03-15
Container access policies 278

alikins:
Very much work in progress.

Currently, mainly adding some management commands for introspect access_policy  ('access-policy-views')
and bits and pieces of sample access policy statements to use  (currently, more or less just the default perms, but
you can:

`http --auth admin:admin GET http://localhost:5001/pulp/api/v3/access_policies/?viewset_name=pulp_container/namespaces`

to get the AccessPolicy for pulp_container/namespaces viewset and it's pulp_href

`http -v --auth admin:admin PATCH http://localhost:5001${ACCESS_POLICY_PULP_HREF}/ @galaxy_ng/app/access_control/plugins/pulp_container/pulp_container_namespaces_viewset.json`

To patch existing policy with the locally store version.

Some related pulpcore / pulpcontainer issues  (not really blockers, but kind of annoying):
- https://pulp.plan.io/issues/8408
- https://pulp.plan.io/issues/8405

### DIFF
--- a/CHANGES/278.feature
+++ b/CHANGES/278.feature
@@ -1,0 +1,1 @@
+Add galaxy style access_policy for pulp_container
--- a/galaxy_ng/app/access_control/access_policy.py
+++ b/galaxy_ng/app/access_control/access_policy.py
@@ -222,6 +222,9 @@ class CollectionAccessPolicy(AccessPolicyBase):
         namespace = models.Namespace.objects.get(name=collection.namespace)
         return request.user.has_perm('galaxy.upload_to_namespace', namespace)
 
+class CollectionUploadAccessPolicy(AccessPolicyBase):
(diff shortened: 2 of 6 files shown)


alikins:
Some notes:

1. Currently example still uses the pulp AccessPolicyFromDB and it's associated helper methods (either via that class or the global conditions. I think it will probably be required to subclass the pulp_container viewset classes if we want to change their 'permission_classes' to use `galaxy_ng.app.access_control.access_policy.AccessPolicy` instead of pulps AccessPolicyFromDB.
2. There appears to be some mechanisms in pulpcore to support overriding the default access policy _statements_ but no mechanism to change the AccessPolicy implementation used in permission_classes.

newswangerd:
Closing in favor of #705

chouseknecht:
Closing due to inactivity.

PR #688, opened 2021-03-15
Add readmes for container distributions.

newswangerd:
GET / PUT `http://localhost:5001/api/automation-hub/_ui/v1/execution-environments/repositories/<repo_name>/_content/readme/`

```
{
  "updated": "2021-03-15T20:28:36.315778Z",
  "created": "2021-03-15T19:02:16.017415Z",
  "text": "#Hello World\n this is another a readme"
}
```


compose
```diff
@@ -6,7 +6,7 @@ set -o errexit
 if [[ -f '.compose.env' ]]; then
   set -o allexport
   # avoid override already set variables
-  source <(grep -v '^#' .compose.env | sed -E 's|^(.+)=(.*)$|: ${\1=\2}; export \1|g')
+  source .compose.env
```
rochacbruno, line 9: this is fixed by https://github.com/ansible/galaxy_ng/pull/689

newswangerd: I accidentally committed that. Oops haha

rochacbruno (approved):
LGTM 

I would like to see a simple test case in the same way we did for collections https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/tests/unit/api/test_api_v3_collections.py 

Just to see that endpoints being hit during CI and ensure future changes to pulp_container isn't breaking it.

ZitaNemeckova (approved):
LGTM from UI's point of view 👍

newswangerd:
@rochacbruno I tried to add tests, but I can't create a ContainerDistribution in pulp container. 

When a distro is created with
```
container_models.ContainerDistribution.objects.create(
            name=name, base_path=name, repository=repo)
```

The container access policy throws an error because the current access policy runs this [function](https://github.com/pulp/pulp_container/blob/master/pulp_container/app/models.py#L490) which attempts to create an owners group and add the authenticated user to the group. Since this create isn't happening within the context of a request, there isn't a user and the operation fails.

I'll create an issue to add unit tests for the execution environment apis. We can add tests once @alikins finishes this https://github.com/ansible/galaxy_ng/pull/687.

PR #691, opened 2021-03-16
compose: print INFO messages to stderr, not stdout

himdel:
Figured out how to dump the database from a running container...

    ./compose exec postgres pg_dump -U galaxy_ng galaxy_ng -Fc > pg.dump

Except that dump is invalid because it starts with the 4 INFO lines compose outputs.

Since the 5th line is only going to stderr, adding >&2 to the other INFO outputs as well.

Cc @rochacbruno

### DIFF
--- a/compose
+++ b/compose
@@ -26,10 +26,10 @@ EOF
   exit 1
 fi
 
-echo "INFO: Using compose profile ${COMPOSE_PROFILE}"
-echo "INFO: ${DEV_SOURCE_PATH:-No} packages installed from source"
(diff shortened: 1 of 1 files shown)


himdel:
(Travis fails on `ERROR: Commit 23112dc5 has no issue attached`, not sure how to fix that?)

rochacbruno (approved):

rochacbruno:
@himdel you need to have your commits signed https://docs.github.com/en/github/authenticating-to-github/signing-commits

After you do that config, you can `git commit -s <commit>` and then `push -f `

himdel:
Thanks!, updated with a signed version, merging :)

PR #695, opened 2021-03-18
Add models for ContainerRegistry Sync Config

rochacbruno:
- Add `ContainerRegistryRemote` as a subclass of `core.Remote` to hold
  configuration for syncing container repositories.
- Add the link model `ContainerRegistryRepos` as a one-to-many linking
  `ContainerRegistryRemote` to `container.Remote`
- Add `ContainerSyncTask` to save the history of sync tasks

Issue: AAH-432

galaxy dot

Graph made with

```bash
# make api/bash
./entrypoint.sh manage graph_models \
                        -g --dot --include-models "Container*,Remote*" \
                        -o src/galaxy.dot \ 
                        core container galaxy
```
then
```py
from graphviz import render
render('dot', 'png', 'galaxy.dot')
```


alikins:
Kind of feels like this should live pulp_container?  

Aside from basing the models from our access_policy mixins, is there anything else that makes this not suitable for pulp_container?

The need for using the access_policy mixins could be significant enough to require it in galaxy_ng, mostly asking out of curiosity.

galaxy_ng/app/models/container.py
```diff
@@ -23,3 +23,39 @@ class ContainerDistroReadme(models.Model):
     created = models.DateTimeField(auto_now_add=True)
     updated = models.DateTimeField(auto_now=True, null=True)
     text = models.TextField(blank=True)
+
+
+class ContainerRegistryRemote(
+    pulp_models.Remote,
+    LifecycleModelMixin,
```
newswangerd, line 30: is `LifecycleModelMixin` required? I think all the pulp base classes inherit from `LifecycleModel`

newswangerd (approved):

PR #696, opened 2021-03-22
Fix groups endpoint viewable only by admin

awcrosby:
pulpcore 3.10.0 introduced [queryset_filtering_required_permission](https://github.com/pulp/pulpcore/blob/master/pulpcore/app/viewsets/user.py#L109)
attribute to GroupViewSet with a permission not used by galaxy_ng.

Currently it can not be overridden simply by setting to None,
so the attribute is removed from the subclass.

Issue: AAH-453

### DIFF
--- a/galaxy_ng/app/api/ui/viewsets/group.py
+++ b/galaxy_ng/app/api/ui/viewsets/group.py
@@ -29,6 +29,12 @@ class GroupViewSet(LocalSettingsMixin, viewsets.GroupViewSet):
     filterset_class = GroupFilter
     permission_classes = [access_policy.GroupAccessPolicy]
 
+    # TODO(awcrosby): replace this by setting attribute to None
+    # after https://pulp.plan.io/issues/8438 is resolved
+    def _remove_attr(self):
+        raise AttributeError
+    queryset_filtering_required_permission = property(_remove_attr)
+
     @transaction.atomic
     def create(self, request, *args, **kwargs):
         name = request.data['name']
--- a/CHANGES/453.bugfix
+++ b/CHANGES/453.bugfix
@@ -1,0 +1,1 @@
+Fix groups endpoint viewable only by admin


fao89:
@mdellweg is more familiarized with this code than me

bmbouter:
If the inability to use the default permissions of Django and the default Group model is a given, this PR looks fine. I get the feeling like we're going down an unsustainable road though with so much subclassing in galaxy_ng.

galaxy_ng/app/api/ui/viewsets/group.py
```diff
@@ -29,6 +29,11 @@ class GroupViewSet(LocalSettingsMixin, viewsets.GroupViewSet):
     filterset_class = GroupFilter
     permission_classes = [access_policy.GroupAccessPolicy]
 
+    # override attribute of parent class
+    def _remove_attr(self):
+        raise AttributeError
+    queryset_filtering_required_permission = property(_remove_attr)
```
mdellweg, line 35: This will make all groups visible to everyone (able to call the list endpoint.)

awcrosby: We use permissions like `galaxy.delete_group` not `auth.delete_group` (which is why we needed to override this, it seems to now require use of `auth.view_group` - but correct me if i am mistaken), and in this case `GroupViewSet` `list` is already visible to everyone: https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/access_control/statements/standalone.py#L176-L180

awcrosby:
@bmbouter I’m not sure if `queryset_filtering_required_permission` attribute is considered a default permission of Django, as it looks like it was a [feature](https://docs.pulpproject.org/pulpcore/plugins/plugin-writer/concepts/rbac/queryset_scoping.html#enabling-queryset-scoping) added in pulpcore for use by plugins… when pulpcore GroupViewSet sets it, it seems the plugin either must use as-is or replace it’s value and must use it. 

Since pulpcore seems to support the attribute being [used or not](https://github.com/pulp/pulpcore/blob/master/pulpcore/app/viewsets/base.py#L310-L313) maybe pulpcore should allow it to be set to `None` instead of just checking for [AttributeError](https://github.com/pulp/pulpcore/blob/master/pulpcore/app/viewsets/base.py#L331)

bmbouter:
This permission is defined by Django actually. https://docs.djangoproject.com/en/2.2/topics/auth/default/#default-permissions

awcrosby:
ohh, the `auth.*` permissions for the model, I see.

bmbouter:
Is the user experiencing the issue an admin user or no? With 3.10 iirc, only admin users are allowed unless RBAC grants non-admin users permissions specifically.

awcrosby:
With pulpcore 3.10 no galaxy_ng user can view the GroupViewSet unless they are an admin.

bmbouter:
That change was here https://github.com/pulp/pulpcore/pull/1064 (see comments where we also asked galaxy_ng to confirm it was ok).

So rather than disabling this, how can galaxy_ng be made to work with it?

awcrosby:
From my testing, I think the breaking change was here https://github.com/pulp/pulpcore/pull/1094, where it sets `queryset_filtering_required_permission = "auth.view_group"`. I don't want to disable this, I want to not use the optional `queryset_filtering_required_permission` attribute which seems to be something the [docs say is supported](https://github.com/pulp/pulpcore/blob/master/pulpcore/app/viewsets/base.py#L310-L313).

ipanova:
> With pulpcore 3.10 no galaxy_ng user can view the GroupViewSet unless they are an admin.

you can change this per endpoint like it has been done here for example https://github.com/pulp/pulpcore/pull/1105/files

awcrosby:
@ipanova yes we do that already https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/api/ui/viewsets/group.py#L30 the problem I find is that pulpcore GroupViewSet sets the `queryset_filtering_required_permission` attribute and it cannot be unset by the plugin.

bmbouter:
@awcrosby I see what you mean, it should be supported to unset it. Thank you for identifying this.

The root issue is actually [here](https://github.com/pulp/pulpcore/blob/master/pulpcore/app/viewsets/base.py#L331-L332). Since the pulpcore viewset defines the attribute, a subclass cannot reasonably give it an attribute error.

So what I recommend doing is fixing the problem where it stems from. Can you open an issue against pulpcore and open a PR to adjust that portion to treat `None` like its treating `AttributeError` today? Also if you need that fix backported to a version of Pulp please [open a backport request](https://docs.pulpproject.org/pulpcore/bugs-features.html#for-backport-requests).

awcrosby:
@bmbouter will do, thank you.

newswangerd (approved):

PR #697, opened 2021-03-22
Add healthcheck management command.

newswangerd:
Adds `django-admin healthcheck` command which returns an exit code of 1 if the expected repositories can't be reached.

```
[galaxy@bddd42dc27f3 /]$ django-admin healthcheck
Authenticating with: admin
Healthcheck passed.
```

A `--user` flag can also be passed to the command to run the healthcheck using a specific user. Otherwise the first user in the database is used for authentication.

### DIFF
--- a/CHANGES/448.feature
+++ b/CHANGES/448.feature
@@ -1,0 +1,1 @@
+Add `healthcheck` command to verify that the app is running.
--- a/galaxy_ng/app/management/commands/healthcheck.py
+++ b/galaxy_ng/app/management/commands/healthcheck.py
@@ -1,0 +1,56 @@
+from django.urls.base import reverse
+
+from django.contrib.auth import get_user_model
+from django.core.management import BaseCommand
+from rest_framework.test import APIClient
+from rest_framework import status as http_code
+
+User = get_user_model()
+
+
+class Command(BaseCommand):
+    """
+    Django management command for creating groups
+    """
+
+    help = 'Verify that the app has initialized successfully.'
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--user',
+            default=None,
+            help='Specify a username to perform tests with.'
+        )
+
+    repos_to_verify = [
+        'published',
+        'staging',
+        'rejected',
+        'community',
+        'rh-certified'
+    ]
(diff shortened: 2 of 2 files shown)


newswangerd:
@fao89 @chambridge please review.

fao89 (approved):
Looks good to me!

fao89:
> Adds `django-admin healthcheck` command which returns an exit code of 1 if the expected repositories can't be reached.
> 
> ```
> [galaxy@bddd42dc27f3 /]$ django-admin healthcheck
> Authenticating with: admin
> Healthcheck passed.
> ```
> 
> A `--user` flag can also be passed to the command to run the healthcheck using a specific user. Otherwise the first user in the database is used for authentication.

During our CI test, we only have the admin user

rochacbruno (approved):

newswangerd:
> > Adds `django-admin healthcheck` command which returns an exit code of 1 if the expected repositories can't be reached.
> > ```
> > [galaxy@bddd42dc27f3 /]$ django-admin healthcheck
> > Authenticating with: admin
> > Healthcheck passed.
> > ```
> > 
> > 
> > A `--user` flag can also be passed to the command to run the healthcheck using a specific user. Otherwise the first user in the database is used for authentication.
> 
> During our CI test, we only have the admin user

Using the admin user makes me uneasy, so I like to give the option for people to pick a user if there is a known one that will always be available.

chambridge:
@newswangerd Thanks for the PR. It provides the sanity check info I needed.

I think it is simple enough though just to create a script in the operator CI that checks for a 200 against the expected repos with an http call to  `/api/galaxy/content/{path}/v3/collections/`. No need to pull this into the product for something so simple; let's close it out and I'll pull the content over.

newswangerd:
@chambridge is it worth keeping this in just in case the parameters around the health check change? I could see us changing the repos we inspect in the future.

chambridge:
I think we revisit if/when that becomes an issue. Branch & PR will be available the history if in the future we need to revive it.

newswangerd:
@chambridge fair enough. We'll try to remember to not break your healthcheck :)

PR #703, opened 2021-03-25
Initial data for remote, repo, distro are now loaded by data migration.

rochacbruno:
- Added 0017 migration to populate initial data for repo, remote and
  distro.
- Clean up fixtures (now it serves only to create a group and a user)
- Fixed test cases that were creating remotes

Issue: AAH-281

## Notes for reviewers:

- Rebinding `.objects` to `objects.using(db_alias)` because I read it is a good practice if in future we have a multi/distributed database deployment. (however not all our migrations uses it)
- **not using transaction context manager** on the operation because I saw other migrations are not using it.
- reverse is `noop` so we don't accidentally delete content from running systems.

## Note for QE/Devs

- When running on existing dev environment it is expected to have an extra remote created, that happens because the old fixtures was creating a remote named `Red Hat Certified` while galaxy_collection names it `rh-certified`.

## Follow ups (after merge)

- [ ] Update https://github.com/ansible/galaxy_collection/blob/devel/roles/post_install_config/tasks/configure_remotes.yml removing those tasks (that roles will be basically for creating galaxy_importer directories)
- [ ] Update WIKI page


galaxy_ng/app/migrations/0017_populate_repos_and_remotes.py
```diff
@@ -0,0 +1,128 @@
...
+                next_version=1,
+                **repo_data,
+            )
+        )
+
+        if not RepositoryVersion.objects.filter(repository__name=repository.name):
+            RepositoryVersion.objects.create(
+                repository=repository,
+                number=0,
+                complete=True
```
bmbouter, line 104: I thought Pulp would create this for you. All repositories come with a RepoVersion 0 by default. Would you be willing to test removing this and then confirm with `shell_plus` that the repository_version is created?

rochacbruno: @bmbouter I added this after having errors because no version was found.

galaxy_ng/app/migrations/0017_populate_repos_and_remotes.py
```diff
@@ -0,0 +1,128 @@
...
+            )
+
+        if not AnsibleDistribution.objects.filter(name=repository.name):
+            AnsibleDistribution.objects.create(
+                name=repository.name,
+                base_path=repository.name,
+                remote=remote,
+                repository=repository,
+                pulp_type="ansible.ansible",
```
bmbouter, line 113: The model definition provides this field by default on the AnsibleDistribution subclass.

galaxy_ng/app/migrations/0017_populate_repos_and_remotes.py
```diff
@@ -0,0 +1,128 @@
...
+                or
+                CollectionRemote.objects.create(**remote_data)
+            )
+
+        repository = (
+            AnsibleRepository.objects.filter(name=repo_data["name"]).first()
+            or
+            AnsibleRepository.objects.create(
+                remote=remote,
+                next_version=1,
```
bmbouter, line 95: This next_version is automatically created at creation time after RepoVersion 0 is created. This could be checked as part of the repo version 0 removal test.

galaxy_ng/app/migrations/0017_populate_repos_and_remotes.py
```diff
@@ -0,0 +1,128 @@
...
+        "name": "community",
+        "url": "https://galaxy.ansible.com/api/",
+        "requirements_file": (
+            "---\n"
+            "collections:\n"
+            "  - name: namespace.collection"
+        ),
+        "token": None,
+        "auth_url": None,
+        "pulp_type": "ansible.ansible",
```
bmbouter, line 16: This field is defined on the model definition and not needed here.

galaxy_ng/app/migrations/0017_populate_repos_and_remotes.py
```diff
@@ -0,0 +1,128 @@
...
+    'rh-certified': {
+        "name": "rh-certified",
+        "url": "https://cloud.redhat.com/api/automation-hub/",
+        "requirements_file": None,
+        "token": None,
+        "auth_url": (
+            "https://sso.redhat.com/auth/realms/"
+            "redhat-external/protocol/openid-connect/token"
+        ),
+        "pulp_type": "ansible.ansible",
```
bmbouter, line 27: similar thing here, this is not needed.

galaxy_ng/app/migrations/0017_populate_repos_and_remotes.py
```diff
@@ -0,0 +1,128 @@
...
+    }
+}
+
+REPOSITORIES = [
+    {
+        "name": "published",
+        "description": "Certified content repository",
+        "pulp_type": "ansible.ansible",
```
bmbouter, line 35: also here.

galaxy_ng/app/migrations/0017_populate_repos_and_remotes.py
```diff
@@ -0,0 +1,128 @@
...
+    },
+    {
+        "name": "community",
+        "description": "Community content repository",
+        "pulp_type": "ansible.ansible",
```
bmbouter, line 40: here

galaxy_ng/app/migrations/0017_populate_repos_and_remotes.py
```diff
@@ -0,0 +1,128 @@
...
+    },
+    {
+        "name": "rh-certified",
+        "description": "Red Hat certified content repository",
+        "pulp_type": "ansible.ansible",
```
bmbouter, line 45: here

bmbouter:
@rochacbruno I thought get_or_create would handle multiple keys just fine. https://docs.djangoproject.com/en/3.1/ref/models/querysets/#get-or-create

I'm ok if you want to leave it, but that way I believe would work and seemed easier. It's up to you tho.

galaxy_ng/app/migrations/0017_populate_repos_and_remotes.py
rochacbruno: @bmbouter I also added this after getting an error, I see other existing migrations also passing this field.

bmbouter: Ok that'll work then. Ty for confirming it's needed.

galaxy_ng/app/migrations/0017_populate_repos_and_remotes.py
bmbouter: Interesting. I guess the API handles it and the db doesn't auto-do it in the manager. Thanks for confirming.

bmbouter (approved):
Thanks @rochacbruno ! :+1:

rochacbruno:
@bmbouter Te problem is that `get_or_create` used all the keys for lookup, so, if there is an existing remote named `community` but with a `token` set different than `None` it is not going to find it and we will be trying to create a new one.

We need to lookup using only `name` and if not found create using all the other fields, is that possible with `get_or_create` ?

bmbouter:
> @bmbouter Te problem is that `get_or_create` used all the keys for lookup, so, if there is an existing remote named `community` but with a `token` set different than `None` it is not going to find it and we will be trying to create a new one.
> 
> We need to lookup using only `name` and if not found create using all the other fields, is that possible with `get_or_create` ?

I understand now. +1 to what you've done this makes sense given the need. `get_or_create` doesn't conditionally use some keys for querying and all keys for setting. Given this data inconsistency this migration deals with it's doubly good we're doing this now.

rochacbruno:
@bmbouter I just learned that `get_or_create` can do that.

What I did

```py
remote = (
    CollectionRemote.objects.filter(name=remote_data["name"]).first()
    or
    CollectionRemote.objects.create(**remote_data)
)
```

Using `get_or_create`

```py
CollectionRemote.objects.get_or_create(
    name=remote_data["name"], 
    defaults=remote_data
)
```

or 

```py
CollectionRemote.objects.filter(
    name=remote_data["name"]
).get_or_create(defaults=remote_data)
```

I think I am going to try it and if works I will update the PR.

bmbouter:
@rochacbruno you just blew my mind! :exploding_head:

bmbouter:
@rochacbruno What I think would also be nice is if instead of looping, the data was immediately available in a dict to the `defaults` call. It will make this migration very easy to read for humans. I'm ok if you don't want to do that also.

fao89:
> @rochacbruno you just blew my mind!

now you have the feeling of the Brazilian Python community!

newswangerd (approved):
LGTM. Thanks for taking care of this.

awcrosby (approved):

PR #704, opened 2021-03-26
[dev env] Auto complete broken on django shell with ipython

rochacbruno:
ref: https://github.com/ipython/ipython/issues/12740#issuecomment-751273584

solution: pin jedi to 0.17.2

No-Issue


alikins (approved):
LGTM

PR #705, opened 2021-03-29
Pulp container RBAC

newswangerd:
UI PR for container model permissions: https://github.com/ansible/ansible-hub-ui/pull/335

### pulp_container access policies
- Allow anyone to create a container namespace. This simplifies the new repo creation process by only requiring that the user have `create_containerdistribution` permissions to create a new container repo.
- Require authentication for pulling containers
- Remove group creation. The user that creates the repository is added as an owner on the repository's namespace instead of creating a new group for the user.

### Groups
Filtering pulp container groups turned out to be a real challenge, so I've opted to remove the pulp container groups entirely.

### Container Repo endpoint
- Removed groups since permissions will be set at the namespace level
- Added a `my_perms` filter that allows querying repos based on permissions that the user has. Ex: `?my_perms=namespace_push_containerdistribution` returns all repos that the current user has push permissions on.
- Updated with the following fields
  - `my_permissions`: returns the permissions my user has on the namespace object. Used for the UI to toggle edit buttons.
  - `owners`: list of users who have permissions on the object outside of the assigned groups. This shows who created and owns the namespace.

```
            "namespace": {
                "name": "a",
                "my_permissions": [
                    "namespace_modify_content_containerpushrepository",
                    "namespace_add_containerdistribution",
                    "namespace_pull_containerdistribution",
                    "view_containernamespace",
                    "namespace_delete_containerdistribution",
                    "change_containernamespace",
                    "namespace_view_containerdistribution",
                    "namespace_push_containerdistribution",
                    "namespace_change_containerdistribution",
                    "delete_containernamespace",
                    "namespace_view_containerpushrepository"
                ],
                "owners": [
                    "test"
                ]
            },
```

### Container namespace endpoint
Add `/api/automation-hub/_ui/v1/execution-environments/namespaces/` endpoint which allows viewing and updating permissions on container namespaces.

Permissions for containers can be set at two levels: container distributions and container namespaces. Namespaces grant a user access to all containers in a namespace.

We are only going to allow users to set permissions at the namespace level for the following reasons:
- Simplicity: setting permissions on the container distribution level is more complicated because some permissions have to be set on the distribution and others have to be set on the container push repository, which would require extra api calls.
- Consistency: permissions are set at a namespace level for collections, so it makes sense from a UX perspective to do the same
- Predictability: only setting permissions at the repo level will lead to confusing situations where a user might think they have permissions on a namespace when they don't.


rochacbruno (approved):

galaxy_ng/app/api/ui/viewsets/group.py
```diff
@@ -28,6 +28,7 @@ class GroupViewSet(LocalSettingsMixin, viewsets.GroupViewSet):
     filter_backends = (DjangoFilterBackend,)
     filterset_class = GroupFilter
     permission_classes = [access_policy.GroupAccessPolicy]
+    queryset = Group.objects
```
alikins, line 31: [requires response]

Not 'Group.objects.all()' ?  

If so, kind of curious what the difference is.

galaxy_ng/app/migrations/0019_pulp_container_access_policies.py
```diff
@@ -0,0 +1,194 @@
...
+                ],
+            },
+        ],
+    }
+}
+
+
+def set_pulp_container_access_policy(apps, schema_editor):
+    AccessPolicy = apps.get_model("core", "AccessPolicy")
+    print(AccessPolicy)
```
alikins, line 165: intentional or leftover debugging?

galaxy_ng/app/api/ui/serializers/execution_environment.py
```diff
@@ -73,6 +88,10 @@ def get_pulp(self, distro):
             }
         }
 
+    def get_owners(self, distro):
+        name = f'container.distribution.owners.{distro.pulp_id}'
+        return [user.username for user in models.User.objects.filter(groups__name=name)]
```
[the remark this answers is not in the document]
newswangerd: oh, this function should be removed. I must have missed that.

galaxy_ng/app/migrations/0019_pulp_container_access_policies.py
```diff
@@ -0,0 +1,194 @@
...
+    dependencies = [
+        ('galaxy', '0018_set_rate_limit_default'),
+        ('container', '0018_containerdistribution_description')
+    ]
+
+    operations = [
+        migrations.RunPython(
+            code=set_pulp_container_access_policy,
+        ),
+        migrations.CreateModel(
```
alikins, line 182: Kind of want to suggest the model creation should be in a separate migration script, but :shrug: I don't think it matters.

alikins: Creating the ContainerNamespace model will also create it's django_permissions right?  Is there any ordering issues with creating the model after the AccessPolicy item update?

The perms used in the policy are all really perms created with pulp_container model creation and already exist right?

galaxy_ng/app/models/container.py
```diff
@@ -16,6 +16,16 @@ class Meta:
         default_related_name = "%(app_label)s_%(model_name)s"
 
 
+class ContainerNamespace(
+        container_models.ContainerNamespace,
```
alikins, line 20: Mild warning about subclassing classes and using the same name... It can get confusing sometimes. Especially for code that does typechecking `isinstance(foo, ContainerNamespace)` etc. Can be confusing to read.

If it works and or is required to work, leave it and don't worry about it.

alikins: Didn't notice this was a proxy model. Not sure if that confuses me less or more. But if that is the idiomatic approach, works for me.

alikins (changes requested):
One or two requested changes (or really, verifications). Most of the rest are vague suggestions you can take or leave.

galaxy_ng/app/migrations/0019_pulp_container_access_policies.py
newswangerd: The ContainerNamespace is just getting created as a proxy model that we can use to add some extra lifecycle methods on. The actual ContainerNamespace is from pulp_container.

galaxy_ng/app/api/ui/viewsets/execution_environment.py
```diff
@@ -36,6 +40,12 @@ class Meta:
             'description': ['exact', 'icontains', 'contains', 'startswith'],
         }
 
+    def has_permissions(self, queryset, name, value):
```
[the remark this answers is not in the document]
newswangerd: would `with_permissions` make more sense?

galaxy_ng/app/api/ui/viewsets/group.py
newswangerd: I'm actually not sure what the difference is. I noticed `Model.objects` was used in pulpcore. For consistency I'll put the all() back

galaxy_ng/app/models/container.py
newswangerd: This is a good point. We should have done this with all our proxy models. I'm not sure what the right thing to do is

newswangerd: I think I'll leave it as is, since we aren't renaming any of the other proxy models, and I think renaming them will create more migrations

galaxy_ng/app/api/ui/viewsets/execution_environment.py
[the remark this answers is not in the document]
alikins: Looks like the django_filter docs tend to name them 'filter_whatever'  (https://django-filter.readthedocs.io/en/master/ref/filters.html#method). 

'filter_by_permission'  ? 'filter_accesible_namespaces' ?  

Your call ;->

galaxy_ng/app/migrations/0019_pulp_container_access_policies.py
alikins: :+1:

alikins (approved):
LGTM

mdellweg:
I was asked to take a look at this and as far as i understood, almost everything looks sane to me. Especially the changes to the access_policy.
However a warning is in order: Running `pulpcore-manager flush` will remove the access policies and still prevent this migration to run. That is the reason, why we decided to switch creating them to a post-migrate signal.

galaxy_ng/app/migrations/0019_pulp_container_access_policies.py
```diff
@@ -0,0 +1,193 @@
+from django.db import migrations
+import django_lifecycle.mixins
+import galaxy_ng.app.access_control.mixins
+
+
+viewsets = {
+    # Note. This is the default Pulp Continer access policy with some modifications.
+    # Our changes have been marked with comments.
+    "distributions/container/container": {
```
ipanova, line 9: The policy for container-push repos is omitted here (tag/untag/remove image) - this will be added later whenever image management will be enabled, correct?

galaxy_ng/app/migrations/0019_pulp_container_access_policies.py
```diff
@@ -0,0 +1,193 @@
...
+            {
+                "action": ["retrieve"],
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": [
+                    "has_namespace_or_obj_perms:container.view_containerdistribution",
+                ],
+            },
+            {
+                "action": ["pull"],
```
ipanova, line 36: there are 2 policies for pull operation - should you want to distinguish rules between private and public repo - it makes sense to keep this as it is proposed, otherwise it could be consolidated into one rule where the authentication is required regardless of the private/public repo type

```
{
                "action": ["pull"],
                "principal": "authenticated",
                "effect": "allow",
            },
```

galaxy_ng/app/migrations/0019_pulp_container_access_policies.py
newswangerd: @ipanova are there things in the container push repo access policies that we should change? it looked like the defaults would work to me.

ipanova: clarified on the irc, resolved.

PR #715, opened 2021-04-08
Replace PULP_REDIS* env variables with PULP_REDIS_URL env variable

bmclaughlin:
This change should be removed once Pulp recognizes REDIS_SSL parameter when building settings for RQ in `pulpcore/rqconfig.py`

Issue: AAH-486

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -97,7 +97,19 @@ run_manage() {
 
 
 redis_connection_hack() {
-    PULP_REDIS_URL="rediss://${PULP_REDIS_PASSWORD}@${PULP_REDIS_HOST:-localhost}:${PULP_REDIS_PORT:-6379}/0"
+    if [[ "$PULP_REDIS_SSL" == "true" ]]; then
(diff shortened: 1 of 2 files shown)


docker/entrypoint.sh
```diff
@@ -96,11 +96,20 @@ run_manage() {
 }
 
 
+redis_connection_hack() {
+    PULP_REDIS_URL="rediss://${PULP_REDIS_PASSWORD}@${PULP_REDIS_HOST:-localhost}:${PULP_REDIS_PORT:-6379}/0"
```
cutwater, line 100: 1. Please select schema `redis` or `rediss` depending on a PULP_REDIS_SSL variable value.
2. The `PULP_REDIS_PASSWORD` parameter is optional. Please add in conditionally. Also it has to be preseded by semicolon.

cutwater (changes requested):

bmclaughlin:
@cutwater, ready for another look.

cutwater (approved):

PR #718, opened 2021-04-09
Fix ordering of last sync task

rochacbruno:
When task is in `running` state it doen't have a `finished_at` value so what we want to show last is the task that was last updated by pulp.

No-Issue


newswangerd (approved):

galaxy_ng/app/migrations/0019_auto_20210410_1749.py
```diff
@@ -0,0 +1,17 @@
+# Generated by Django 2.2.20 on 2021-04-10 17:49
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('galaxy', '0018_set_rate_limit_default'),
```
fao89, line 9:
```suggestion
        ('galaxy', '0019_containernamespace'),
```
https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/migrations/0019_containernamespace.py
and rename this migration to 0020_...

newswangerd (approved):

PR #731, opened 2021-04-16
dev/Dockerfile.base: install python deps as galaxy user, don't chown after

himdel:
In the dev docker environment, `compose build` takes ages, on the `chown -R ...` step.
That's because we have to chown every single python dependency that we just downloaded as root, fixing that.

Before:

	(root) python -m venv /venv	# creates and populates a venv
	(root) pip install ...		# installs deps in there
	(root) chown -R galaxy /venv	# changes owner of all these files

After:

	(root) mkdir /venv
	(root) chown galaxy /venv	# while still empty
	(galaxy) python -m venv /venv
	(galaxy) pip install		# no chown for these
	

We can also skip chowning `/app`, using `COPY --chown` instead.

`django-admin collectstatic` needs to run as root however (fails on `/var/lib/pulp` access otherwise),
and keeping the final chown for the rest of the files.

No-Issue

---

Running the original version except with `chown -c | wc -l` and `time`, and comparing with this PR...

| version | files chowned | time chown | total build time |
|-|-|-|-|
| master | 70866 | 53m53.118s | 54m41.144s |
| this PR | 197 | 0m7.687s | 5m14.362s |

### DIFF
--- a/dev/Dockerfile.base
+++ b/dev/Dockerfile.base
@@ -42,7 +42,12 @@ RUN set -ex; \
     && rm -rf /var/cache/dnf/ \
     && rm -f /var/lib/rpm/__db.* \
     \
-    && python3 -m venv /venv
+    && mkdir /venv \
+    && chown ${USER_NAME}:${USER_GROUP} /venv
+
+USER ${USER_NAME}:${USER_GROUP}
+RUN set -ex; \
+    python3 -m venv /venv
 
 ENV PATH="/venv/bin:${PATH}" \
     VIRTUAL_ENV="/venv"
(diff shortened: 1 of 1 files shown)


alikins:
My previous unfinished attempt along this line can be found at https://github.com/ansible/galaxy_ng/compare/master...alikins:clean_docker_file_group_perms 
(albeit in a roughly rebased and untested state at the moment).

I only mention it as a reference point for a branch with similar intents, albeit one that never fully worked.

alikins:
Further test with different requirements shows this branch is consistently 30s faster (on my machine)

alikins (approved):
LGTM. Definitely speeds up the `./compose build` step.

cutwater (approved):

cutwater:
I wonder though how did you get an almost 1h build time.

himdel:
Thanks both of you! :) 

> I wonder though how did you get an almost 1h build time.

@cutwater I don't know either, on my previous laptop it also took way too long but it was under half an hour.
I suppose part of it may be ext4 vs btrfs (or whatever fedora is using these days),
or maybe I need to figure out if I have the filesystem block size different from the actual SSD block size (this being a new notebook),
the problem is definitely not CPU-bound but that's about all I know for now :).

PR #733, opened 2021-04-19
[WIP] Update pr_check.sh

jameswnl:


awcrosby:
We plan to soon move our CI bits to GitHub Actions, not sure if that can be added here instead: https://github.com/ansible/galaxy_ng/tree/master/.github/workflows

jameswnl:
> We plan to soon move our CI bits to GitHub Actions, not sure if that can be added here instead: https://github.com/ansible/galaxy_ng/tree/master/.github/workflows

I assume Hub's current CI bits are centered around python unit test framework?

So what pr_check job can do that is not available (or difficult to do) in GitHub Actions workflow is to validate that the PR branch will build and is deployable to OSD4 
 - the deployment is in an environment that is very similar to that of the production
 - have its own UI deployment
 - execute IQE smoke test suites
This is especially useful and important for the c.r.c's app-interface CI/CD toolchain.

You can have the regular GitHub Actions doing its unit test coverage (and e.g. linting etc) and pr_check job focusing on the CI/CD aspect.  (FYI, Analytics is doing just like that)

PR #738, opened 2021-04-21
Make waiting-for-migrations script call optional

bmclaughlin:
Issue: AAH-424


docker/entrypoint.sh
```diff
@@ -82,7 +83,7 @@ run_service() {
 
     if [[ "${WITH_MIGRATIONS}" -eq "1" ]]; then
         django-admin migrate
-    else
+    elif [[ "${WAIT_FOR_MIGRATIONS}" -ne "0" ]]; then
```
cutwater, line 86: please explicitly compare with "1"

cutwater (changes requested):
`entrypoint.sh` script is also used in local development environment. since you're changing default behavior, please make sure `WAIT_FOR_MIGRATIONS` is set correctly for local environment.

cutwater (approved):

PR #746, opened 2021-04-23
Download correct version of UI during package builds.

newswangerd:
During the python package build:
- If the `ALTERNATE_UI_DOWNLOAD_URL` environment variable is set, attempt to download the specified UI.
- Attempt to download the version of the UI that matches the current version of the API.
- If no version can be found, download the UI from https://github.com/ansible/ansible-hub-ui/releases/tag/dev

This is a companion to https://github.com/ansible/ansible-hub-ui/pull/372, which provides a build of the master branch on the 'dev' github release.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,7 @@ class PrepareStaticCommand(Command):
         with tempfile.NamedTemporaryFile() as download_file:
             log.info(f"Downloading UI distribution to temporary file: {download_file.name}")
 
-            if (self.ALTERNATE_UI_DOWNLOAD_URL):
+            if self.ALTERNATE_UI_DOWNLOAD_URL:
                 log.info(f"Downloading UI from {self.ALTERNATE_UI_DOWNLOAD_URL}")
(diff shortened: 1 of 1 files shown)


awcrosby (approved):

setup.py
```diff
@@ -37,12 +44,25 @@ def run(self):
 
         with tempfile.NamedTemporaryFile() as download_file:
             log.info(f"Downloading UI distribution to temporary file: {download_file.name}")
-            urllib.request.urlretrieve(self.UI_DOWNLOAD_URL, filename=download_file.name)
+
+            if (self.ALTERNATE_UI_DOWNLOAD_URL):
```
cutwater, line 48: Parenthesis are redundant here

cutwater (changes requested):

setup.py
newswangerd: woops. This is what I get from constantly switching between javascript and python

PR #750, opened 2021-04-29
Fix schema view

fao89:
Currently `/pulp/api/v3/` breaks with:
```
AssertionError: Expected view UserViewSet to be called with a URL keyword argument named "pk".
```

No-Issue

### DIFF
--- a/galaxy_ng/app/access_control/access_policy.py
+++ b/galaxy_ng/app/access_control/access_policy.py
@@ -218,6 +218,8 @@ class CollectionAccessPolicy(AccessPolicyBase):
     NAME = 'CollectionViewSet'
 
     def can_update_collection(self, request, view, permission):
+        if not view.kwargs.get("path"):
+            return False
         collection = view.get_object()
         namespace = models.Namespace.objects.get(name=collection.namespace)
         return request.user.has_perm('galaxy.upload_to_namespace', namespace)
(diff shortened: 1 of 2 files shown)


fao89:
@himdel this fixes the error we saw when hitting `/pulp/api/v3/`
Not sure if it is the best way to fix

galaxy_ng/app/access_control/access_policy.py
```diff
@@ -218,6 +218,8 @@ class CollectionAccessPolicy(AccessPolicyBase):
     NAME = 'CollectionViewSet'
 
     def can_update_collection(self, request, view, permission):
+        if not view.kwargs.get("path"):
+            return False
```
alikins, line 222: Whats the scenario where this gets called on a CollectionViewSet without a path set?

The non /content/ case like `/api/automation-hub/v3/collections` should get a 'path' kwarg set to the default.

galaxy_ng/app/access_control/access_policy.py
```diff
@@ -239,11 +241,17 @@ class UserAccessPolicy(AccessPolicyBase):
...
+        else:
+            user = request.user
         return user.is_superuser
 
     def is_current_user(self, request, view, action):
-        return request.user == view.get_object()
+        if view.kwargs.get("pk"):
+            return request.user == view.get_object()
+
+        return True
```
alikins, line 254: Should this default to True? Feels like it should default False until proven that `request.user == view.get_object()` ?

But I'm not sure how this get's called with no pk, since it is only used from UserViewSet delete and I'm making the assumption that DELETE is only valid method in /_ui/users/<pk>/ .

galaxy_ng/app/access_control/access_policy.py
fao89: on `/pulp/api/v3/` when generating the OpenAPI schema:
image

PR #759, opened 2021-05-06
Update all deps, including galaxy-importer

alikins:
Issue: AAH-576


alikins:
Note: There will still need to be another rev of deps to bring in django >=2.2.23 and newer pulpcore if needed.

awcrosby (commented):
It looks like we will have another galaxy-importer release before 4.3 code freeze, so we can probably close this and wait for that

awcrosby (approved):

PR #771, opened 2021-05-12
Speedup dev docker builds by using g+s to avoid chown -R

alikins:
No-Issue

This builds on #731. 

Use `mkdir --mode` to make most of the dirs group 'sticky' so group perms stick, so user can continue to write to subdirs even if ROOT unintentionally creates dirs 0755. 

Some rearranging of dir creation order so we can create parent dirs first so we can make them group sticky.
For `/var/lib/pulp` for example, this removes the need for a `chown -R /var/lib/pulp`.

On top of #731, it seems to shave about ~25secs off a `./compose build` down to about ~1m15sec.  Down from about ~3m30s before either #731 or this change.

### DIFF
--- a/dev/Dockerfile.base
+++ b/dev/Dockerfile.base
@@ -41,9 +41,10 @@ RUN set -ex; \
     && dnf clean all \
     && rm -rf /var/cache/dnf/ \
     && rm -f /var/lib/rpm/__db.* \
-    \
-    && mkdir /venv \
-    && chown ${USER_NAME}:${USER_GROUP} /venv
+    && mkdir --mode=2775 /venv \
+    && chown ${USER_NAME}:${USER_GROUP} /venv \
+    && mkdir --mode=2775 /app \
+    && chown ${USER_NAME}:${USER_GROUP} /app
 
 USER ${USER_NAME}:${USER_GROUP}
 RUN set -ex; \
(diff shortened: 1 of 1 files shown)


dev/Dockerfile.base
```diff
@@ -75,19 +77,19 @@ RUN set -ex; \
...
-             /etc/ansible \
-             /entrypoints.d \
-    && chown -R ${USER_NAME}:${USER_GROUP} \
-        /var/lib/pulp \
+    && chown ${USER_NAME}:${USER_GROUP} \
         /tmp/ansible \
         /etc/ansible \
-    && chmod -R 0775 /var/lib/pulp \
-                     /app/docker/entrypoint.sh \
+    && chmod -R 2775 /app/docker/entrypoint.sh \
```
himdel, line 91: Not sure about this one, do we want g+s on a shell script here?

himdel:
I'm also seeing a timing improvement, from 6:26.48 down to 5:15.89 :+1:

dev/Dockerfile.base
alikins: good catch. Looks like I got overzealous when fixing up the merge. Will update.

dev/Dockerfile.base
```diff
@@ -55,7 +56,8 @@ ENV PATH="/venv/bin:${PATH}" \
 COPY ./requirements/requirements.common.txt /tmp/requirements.txt
 
 RUN set -ex; \
-    pip install --no-cache-dir --upgrade pip \
+    chmod 2775 /venv/bin \
```
alikins, line 59: Looking now, I'm wondering if this `chmod 2775 /venv/bin` is redundant, since /venv/ should already be 2775 and owned by USER_NAME.USER_GROUP.

I'll test and see if it's required.

dev/Dockerfile.base
alikins: Updated with a fix for this chmod.

dev/Dockerfile.base
alikins: Yeah, the 'chmod 2775 /venv/bin' was unneeded.  Updated now.

alikins:
@himdel
> I'm also seeing a timing improvement, from 6:26.48 down to 5:15.89

I'm curious what the setup / config is for a `./compose build` to take roughly 5x longer than on my system (~2year old thinkpad t460p).

Is that timing including installing pulpcore/pulp_ansible/pulp_container/etc from source?
(ie, `DEV_SOURCE_PATH='galaxy_ng:pulpcore:pulp_ansible:pulp_container:django-extensions'` or similar)

himdel:
> I'm curious what the setup / config is for a ./compose build to take roughly 5x longer than on my system (~2year old thinkpad t460p).

Mine is a t590, no idea why it would be slower. 

> Is that timing including installing pulpcore/pulp_ansible/pulp_container/etc from source?

My `.compose.env` is just..

```
COMPOSE_PROFILE=standalone
DEV_SOURCE_PATH='galaxy_ng'
LOCK_REQUIREMENTS=1
WAIT_FOR_MIGRATIONS=1
```

And I'm measuring after `docker system -a prune`, otherwise the build times range anywhere from 10 seconds to about 4 minutes, depending on what is getting cached and what is getting rebuilt.

If you'd like to take a look at the timing, this is my `./compose build | ts -s` output: https://gist.github.com/himdel/d8d1642a0f6cc55b6ac2dac13c69e970

Highlights.. 
donwloading containers 20s
installing dnf dependencies ~2m
installing python dependencies ~1m
installing more python dependencies ~2m

Not sure how that differs from anyone else :)

chouseknecht (approved):

PR #774, opened 2021-05-14
Pin click==7.1.2, click>=8.0.0 req py38

alikins:
pulpcore uses 'rq' which uses 'click'.

click version 8.0.0 or higher don't work with
python < 3.8 and we currently use python~=3.6

If click>=8.0.0 is used with python 3.6. it will
eventually raise an exception like:

ValueError: 'default' must be a list when 'multiple' is true.

See https://pulp.plan.io/issues/8744

### DIFF
--- a/CHANGES/637.bugfix
+++ b/CHANGES/637.bugfix
@@ -1,0 +1,1 @@
+Pin 'click' version to 7.1.2 for 'rq' compat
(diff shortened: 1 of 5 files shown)


rochacbruno (approved):

newswangerd (approved):

cutwater (approved):
LGTM, but @alikins could you please give more descriptive PR titles.

bmclaughlin (approved):

PR #777, opened 2021-05-24
Pin pulp-container

fao89:
This is currently breaking pulp-operator
https://github.com/pulp/pulp-operator/runs/2652733659?check_suite_focus=true

pulp-container 2.6 requires pulpcore >=3.12.1
https://github.com/pulp/pulp_container/blob/2.6.0/requirements.txt

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -83,7 +83,8 @@ requirements = [
     "pulp-ansible==0.7.3",
     "django-prometheus>=2.0.0",
     "drf-spectacular",
-    "pulp-container>=2.5.2",
+    # pulp-container 2.6 requires pulpcore >=3.12.1
+    "pulp-container>=2.5.2,<2.6.0",
     # click 8 requires py38,
     # can be removed once we require >=py38
     "click==7.1.2",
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -80,10 +80,11 @@ def run(self):
     "Django~=2.2.23",
     "galaxy-importer==0.3.2",
     "pulpcore<3.12,>=3.11.1",
+    "click~=7.1.2",
```
rochacbruno, line 83:
```suggestion
    "click~=7.1.2",  # added as a workaround to fix rq, can be removed when pulpcore gets updated to 3.11.2
```

newswangerd:
We pinned click here: https://github.com/ansible/galaxy_ng/pull/774/files

Might need to rebase

rochacbruno (approved):

alikins:
Also needs requirements/*.txt updated.
ie, run

`PIP_COMPILE_UPDATE_SPEC='--upgrade-package "pulp-container>=2.5.2,<2.6.0"' make requirements`

alikins (changes requested):
needs requirements.*.txt updated

alikins (approved):

PR #781, opened 2021-05-27
Update collection download urls

newswangerd:


galaxy_ng/app/api/v3/serializers/collection.py
```diff
@@ -105,10 +105,11 @@ def get_download_url(self, obj):
         """
         Get artifact download URL.
         """
-        host = settings.CONTENT_ORIGIN.strip("/")
-        prefix = settings.CONTENT_PATH_PREFIX.strip("/")
+
+        host = settings.ANSIBLE_API_HOSTNAME.strip("/")
+        prefix = settings.GALAXY_API_PATH_PREFIX.strip("/")
```
bmbouter, line 110: What was the motivation for changing these two lines?

bmbouter: Actually I realize now you want the not-content-app to hand out the data. That makes the `ANSIBLE_API_HOSTNAME` change look good. I don't think I understand what `GALAXY_API_PATH_PREFIX` is since it's not a pulp_ansible setting https://docs.pulpproject.org/pulp_ansible/settings.html

newswangerd: That points the download URL to the galaxy apis instead of the content app

bmbouter:
Is there an opportunity to only have this go to older release branches instead of master? For master I think we'll want to use the ContentGuard from pulp_ansible.

cutwater (approved):

galaxy_ng/app/api/v3/serializers/collection.py
```diff
@@ -105,10 +105,11 @@ def get_download_url(self, obj):
...
         distro_base_path = self.context["path"]
-        return f"{host}/{prefix}/{distro_base_path}/{obj.relative_path}"
+        return f"{host}/{prefix}/v3/artifacts/collections/{distro_base_path}/{obj.relative_path}"
```
alikins, line 112: Before https://github.com/ansible/galaxy_ng/commit/3caff652bb5df817040420ed0e0842a23af790a8#  (ie, in https://github.com/ansible/galaxy_ng/blob/3caff652bb5df817040420ed0e0842a23af790a8/galaxy_ng/app/api/v3/serializers/collection.py)

This method was:
```
    def get_download_url(self, obj):
        """
        Get artifact download URL.
        """

        host = settings.CONTENT_ORIGIN.strip("/")
        prefix = settings.CONTENT_PATH_PREFIX.strip("/")
        distro_base_path = self.context["path"]
        return f"{host}/{prefix}/{distro_base_path}/{obj.relative_path}"
```

The self.context['path'] was set in the 'retrieve' method. And the context was set via
https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/api/v3/viewsets/collection.py#L50-L67

That was done to override the default pulp 'download_url' to include the distribution.base_path info.

If the 'download_url' from pulp is still incomplete for this use, are there other issues that may still be wrong?
Namely thinking of reverse() url lookups for these endpoints. https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/api/v3/viewsets/collection.py#L50-L67 added that info to the context to be used in the Collection serializer.

PR #783, opened 2021-06-01
Make synclist task with policy 'exclude' include all versions not just is_latest AAH-485

alikins:
Intended fix for 'Syncing from synclist is pulling all content, ignoring sync toggles'

https://issues.redhat.com/browse/AAH-585?focusedCommentId=16232352&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16232352  goes into details about the root cause.

I think the synclist 'curate' is only removing the latest version of the collection from the synclist RepositoryVersion instead of all the versions when the synclist.policy=='exclude'.

This pr is a WIP that sets `is_highest=False` for exclude case.

Issue: AAH-485


galaxy_ng/app/tasks/synclist.py
```diff
@@ -155,4 +163,5 @@ def curate_synclist_repository(synclist_pk, **kwargs):
     else:
         raise RuntimeError("Unexpected synclist policy {}".format(synclist.policy))
 
+    log.debug('task_kwargs: %s', pprint.pformat(task_kwargs))
```
rochacbruno, line 166: do we pprint our log output? `pformat` generates indentation and I think we need to keep log (when output is a file) in to single lines and let formatting to the handlers and/or reader.

galaxy_ng/app/tasks/synclist.py
```diff
@@ -125,19 +125,27 @@ def curate_synclist_repository(synclist_pk, **kwargs):
...
+    is_highest = True
+    if synclist.policy == 'exclude':
+        is_highest = False
+
     collection_versions = CollectionVersion.objects.filter(
         Q(
             repositories=synclist.upstream_repository,
             collection__namespace__in=namespaces,
-            is_highest=True,
+            is_highest=is_highest,
```
rochacbruno, line 137: what if here you use

```py
is_highest=synclist.policy != "exclude"
```

and remove the `if` block above

CHANGES/585.bugfix
```diff
@@ -0,0 +1,27 @@
+Resolve issues syncing all content via synclist
```
rochacbruno, line 1: this CHANGELOG must be a single line now

rochacbruno (changes requested):
pending also a rebase

galaxy_ng/app/tasks/synclist.py
```diff
@@ -125,19 +125,27 @@ def curate_synclist_repository(synclist_pk, **kwargs):
 
     namespaces = synclist.namespaces.filter().values_list("name", flat=True)
 
+    # FIXME: For 'exclude', we want to remove all version of the excluded collection
+    is_highest = True
+    if synclist.policy == 'exclude':
+        is_highest = False
```
rochacbruno, line 131: This whole block is not needed because `synclist.policy == 'exclude'` already evaluates to a boolean

rochacbruno (approved):

galaxy_ng/app/tasks/synclist.py
```diff
@@ -125,16 +125,21 @@ def curate_synclist_repository(synclist_pk, **kwargs):
 
     namespaces = synclist.namespaces.filter().values_list("name", flat=True)
 
+    # include adds only highest version and exclude removes all versions of the collection.
+    is_highest_query_param = {}
```
rochacbruno, line 129: Explanation for reviewers:

When policy is `exclude` this is going to be an empty dict,so the expansion later will make the query `Q(` to not filter by version including on the queryset all the versions of the collection.

When it is `include` then `is_highest = True` is added to the query, to it filters only the latest version of a collection.

awcrosby (approved):

newswangerd (approved):

PR #790, opened 2021-06-02
Use red-hat-certified as the standard for base_path

rochacbruno:
Previous versions provisioned using galaxy_collection roles
used `red-hat-certified` instead of `rh-certified` as the
`base_path` for distro.

So to avoid breaking the migration with DB Integrity Error
we are from now, using `red-hat-certified` as the standard
for this distro `base_path` while keeping `rh-certified`
as its name and related repo/remote names.

No-Issue

### DIFF
--- a/galaxy_ng/app/migrations/0017_populate_repos_and_remotes.py
+++ b/galaxy_ng/app/migrations/0017_populate_repos_and_remotes.py
@@ -40,6 +40,10 @@ REPOSITORIES = [
             ),
             "pulp_type": "ansible.ansible",
             "rate_limit": 8,
+        },
+        "distribution": {
+            "name": "rh-certified",
+            "base_path": "red-hat-certified"
         }
     }
 ]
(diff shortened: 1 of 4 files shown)


galaxy_ng/app/migrations/0017_populate_repos_and_remotes.py
```diff
@@ -40,6 +40,10 @@
             ),
             "pulp_type": "ansible.ansible",
             "rate_limit": 8,
+        },
+        "distribution": {
+            "name": "rh-certified",
```
newswangerd, line 45: when I ran this command from the installer: https://github.com/ansible/galaxy_collection/blob/devel/roles/post_install_config/tasks/configure_remotes.yml#L34-L41 it created a distro with `red-hat-certified` for the base_path and name. I wonder if we could make this red-hat-certified as well.

rochacbruno:
replaced by #792

PR #804, opened 2021-06-18
Remove pulp-container from 4.2.

newswangerd:
- Remove pulp container
- Update to latest django LTS
- Pin to pulp core 3.7

No-Issue


awcrosby (approved):
it looks like our last 4.2.3 release was `pulpcore==3.7.5` so we are not going backward from 3.8 to 3.7, so this LGTM

hendersonreed (approved):
Looks right to me, lgtm

PR #807, opened 2021-06-24
Workaround bonfire stripping resource limits of ClowdApp

cutwater:
This change replaces call to `_deploy_ephemeral_env.sh` script
with it's contents and adds additional required `--no-remove-resources`
parameter to the `bonfire deploy` call.

No-Issue

### DIFF
--- a/docker/etc/settings.py
+++ b/docker/etc/settings.py
@@ -1,4 +1,3 @@
-import json
 import os
 
 DEFAULT_FILE_STORAGE = 'storages.backends.s3boto3.S3Boto3Storage'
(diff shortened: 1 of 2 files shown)


jameswnl (approved):

alikins (approved):
LGTM.  I'm assuming the new code in pr_check.sh is more or less deploy_emphemeral_env.sh but with the --no-remove-resources added.

newswangerd:
@cutwater  Why is this going in `pr_check.sh`?

cutwater:
@newswangerd `pr_check.sh` runs bonfire from `$CICD_ROOT/deploy_ephemeral_env.sh` to deploy ephemeral environment and some testing. Unfortunately `deploy_ephemeral_env.sh` doesn't provide an option of running it with the `--no-remove-resources`, so to workaround that I had to replicate it directly in `pr_chech.sh`

PR #810, opened 2021-06-28
Implement distro based import aah 558

alikins:
For case #2, the goal is to support having an optional ‘inbound’ repo for ever exposed repo (Published, Red Hat Reviews, Community, etc) instead of creating an inbound repo for every potential namespace.


awcrosby:
Closing this as issue AAH-739 has been addressed

PR #816, opened 2021-06-30
Create per namespace inbound repos on all namespace creation

alikins:
This will create a new Repository and Distribution for every
Namespace in Galaxy.

Issue: AAH-739

### DIFF
--- a/CHANGES/739.bugfix
+++ b/CHANGES/739.bugfix
@@ -1,0 +1,1 @@
+Create 'inbound-namespaces' whenever a namespace is created.
--- a/galaxy_ng/app/models/namespace.py
+++ b/galaxy_ng/app/models/namespace.py
@@ -9,7 +9,6 @@ from pulp_ansible.app.models import AnsibleRepository, AnsibleDistribution
 from galaxy_ng.app.access_control import mixins
 from galaxy_ng.app.constants import INBOUND_REPO_NAME_FORMAT
 
-
 __all__ = ("Namespace", "NamespaceLink")
 
 
@@ -48,9 +47,11 @@ class NamespaceManager(models.Manager):
             create_inbound_repo(obj.name)
         return super().bulk_create(objs, **kwargs)
 
-    def delete(self):
-        delete_inbound_repo(self.name)
-        return super().delete()
+    def get_or_create(self, *args, **kwargs):
+        ns, created = super().get_or_create(*args, **kwargs)
+        if created:
+            create_inbound_repo(kwargs['name'])
+        return ns, created
 
 
 class Namespace(LifecycleModel, mixins.GroupModelPermissionsMixin, AutoDeleteObjPermsMixin):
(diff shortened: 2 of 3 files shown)


alikins:
NOTE: This might be heavy handed, since this will create the 'inbound-<namespace>' repo whenever a namespace is created, including sync.

Currently the 'inbound-<namespace>' repos are not created when Namespaces are created by sync.

An alternatively would be to keep the sync behaviour the same (only create the Namespace and not the inbound-<namespace> repo) and to create the inbound-<namespace> repo "when needed".

CHANGES/739.bugfix
```diff
@@ -0,0 +1,3 @@
+Create 'inbound-namespaces' whenever a namespace is created.
+
+This includes namespaces created via sync.
```
awcrosby, line 3: recommend changing this changelog entry to say
`Ensure 'inbound-<namespace>' repo exists when accessing namespace`
since we already create these repos on sync - this just ensures namespaces sync'd before https://github.com/ansible/galaxy_ng/commit/8c5f3f945435a792504adc97df7bd2d2a94b412d get a repo

awcrosby (changes requested):

awcrosby (approved):
@alikins overall LGTM

It seems this edit showed we have a bug in `delete_inbound_repo()` or `test_delete_namespace_deletes_inbound_repo()` (test was passing because it used `get_or_create` and never actually created the repo) - so i believe that is the cause of the CI failure.

alikins:
@awcrosby hmm, I think I ran into that same issue with an earlier version and thought I had fixed it somehow. If that was the same issue (or new issue too of course), I'll track it down and fix it.

alikins:
reproduce CI error

```
$ ./compose run --rm --service-ports api /bin/bash
# in container shell
$ ./entrypoint.sh django-admin test --debug-mode -v 3 --noinput galaxy_ng.tests.unit
<... lots of test output ...>
======================================================================
FAIL: test_delete_namespace_deletes_inbound_repo (galaxy_ng.tests.unit.test_models.TestNamespaceModelManager)
When deleting a Namespace the manager should delete inbound instances.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/app/galaxy_ng/tests/unit/test_models.py", line 52, in test_delete_namespace_deletes_inbound_repo
    self.assertFalse(AnsibleRepository.objects.filter(name=inbound_name).exists())
AssertionError: True is not false

----------------------------------------------------------------------
Ran 116 tests in 9.535s

FAILED (failures=1, skipped=19)
```

alikins:
Looks like the test case is calling delete() on the Namespace queryset (Namespace.objects.filter(names=ns_name).delete().

I believe that is deleting all members of that queryset based on https://docs.djangoproject.com/en/3.2/topics/db/queries/#deleting-objects
Seems to imply that makes for 'bulk' deletes as well.

Think NamespaceManager.delete() isn't needed, and that delete() should be on the Namespace model.
And test case update to get the right instance.

alikins:
Namespace delete unit test should work now  (and more importantly, some_namespace.delete() will delete the inbound repos correctly).

awcrosby (approved):
After another review, LGTM except for CI issues, which are not currently related to this PR.

newswangerd (approved):

PR #821, opened 2021-07-06
Replace enqueue_with_reservation with dispatch for pulpcore 3.14.0 compat

alikins:
Replace enqueue_with_reservation() with dispatch()

Primary difference is that enqueue_with_reservation()
returns an 'rq' job, and 'rq' support is being removed
in pulpcore.

The new 'dispatch()' returns a pulpcore.plugin.models.Task
object. dispatch() also no longer supports the 'options' arg.

pulpcore 3.13 removes 'enqueue_with_reservation'
https://docs.pulpproject.org/pulpcore/changes.html#id96

pulpcore 3.12.0 deprecated 'enqueue_with_reservation'
https://docs.pulpproject.org/pulpcore/changes.html#id152

See https://docs.pulpproject.org/pulpcore/plugins/api-reference/tasking.html#pulpcore.plugin.tasking.dispatch
for docs on dispatch().

More explanation of the diff between enqueue_with_reservation()
and dispatch() can be seen at:
https://github.com/pulp/pulpcore/commit/753e4ddb184efa290d71b4a53b419d66a0b60745
and https://pulp.plan.io/issues/8496

Issue: AAH-706


rochacbruno:
@alikins you need to tell the CI to use the proper versions of pulpcore

https://github.com/ansible/galaxy_ng/blob/master/.travis/before_install.sh#L73

and

https://github.com/ansible/galaxy_ng/blob/master/.travis/before_install.sh#L103

rochacbruno (changes requested):
The upgrade of dependencies made migration to break

```bash
File "/usr/local/lib/python3.6/site-packages/galaxy_ng/app/migrations/0017_populate_repos_and_remotes.py", line 104, in populate_initial_repos
    "pulp_type": "ansible.ansible"
```

we need a fix for that migration as part of this PR to make the CI to pass.

```__fake__.DoesNotExist: AnsibleDistribution matching query does not exist.```

alikins:
@rochacbruno 

> we need a fix for that migration as part of this PR to make the CI to pass.
> 
> __fake__.DoesNotExist: AnsibleDistribution matching query does not exist.

hmm, interesting. Did not expect that.

AnsibleDistribution no longer exists?

rochacbruno:
@alikins the entity exists, but the SQL query made on the migration is failing for some reason. Did we changed something on `get_or_create` logic?

rochacbruno (approved):
LGTM

Waiting to hear from @cutwater related to the appSRE failure.

PR #826, opened 2021-07-06
Add unix socket support to collection version download view

awcrosby:
Issue: AAH-743

### DIFF
--- a/galaxy_ng/app/api/utils.py
+++ b/galaxy_ng/app/api/utils.py
@@ -1,10 +1,15 @@
 from collections import namedtuple
 import re
+from requests.adapters import HTTPAdapter
+import socket
+from urllib3.connection import HTTPConnection
+from urllib3.connectionpool import HTTPConnectionPool
 
 from galaxy_importer.schema import MAX_LENGTH_NAME, MAX_LENGTH_VERSION
 
 CollectionFilename = namedtuple("CollectionFilename", ["namespace", "name", "version"])
 
+LOCALHOST = "localhost"
 FILENAME_REGEXP = re.compile(
     r"^(?P<namespace>\w+)-(?P<name>\w+)-" r"(?P<version>[0-9a-zA-Z.+-]+)\.tar\.gz$"
 )
@@ -53,3 +58,31 @@ def parse_collection_filename(filename):
         raise ValueError(f"Expected version to be max length of {MAX_LENGTH_VERSION}")
 
     return CollectionFilename(namespace, name, version)
+
+
+class SocketHTTPConnection(HTTPConnection):
+    def __init__(self, socket_file):
+        self.socket_file = socket_file
+        super().__init__(LOCALHOST)
+
+    def connect(self):
+        self.sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+        self.sock.connect(self.socket_file)
+
+
+class SocketHTTPConnectionPool(HTTPConnectionPool):
+    def __init__(self, socket_file):
+        self.socket_file = socket_file
+        super().__init__(LOCALHOST)
+
+    def _new_conn(self):
+        return SocketHTTPConnection(self.socket_file)
+
+
+class SocketHTTPAdapter(HTTPAdapter):
+    def __init__(self, socket_file):
+        self.socket_file = socket_file
+        super().__init__()
+
+    def get_connection(self, url, proxies=None):
+        return SocketHTTPConnectionPool(self.socket_file)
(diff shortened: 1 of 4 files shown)


galaxy_ng/app/api/v3/viewsets/collection.py
```diff
@@ -240,7 +250,10 @@ def get(self, request, *args, **kwargs):
             filename=self.kwargs['filename'],
         )
 
-        response = requests.get(url, stream=True, allow_redirects=False)
+        if settings.PULP_CONTENT_BIND:
```
rochacbruno, line 253: Isn't it a good idea to check also if this `PULP_CONTENT_BIND` starts with `"unix:"` ?

According to gunicorn docs https://docs.gunicorn.org/en/stable/settings.html#bind 

```
A string of the form: HOST, HOST:PORT, unix:PATH, fd://FD. An IP is a valid HOST.

Changed in version 20.0: Support for fd://FD got added.
```

and pulp_installer maps it directly to that gunicorn argument https://github.com/pulp/pulp_installer/tree/master/roles/pulp_content#role-variables

```suggestion
        if settings.PULP_CONTENT_BIND and settings.PULP_CONTENT_BIND.startswith("unix:"):
```

awcrosby: sounds like it must be a unix socket path and not the other forms, so yes we can check for this

newswangerd (approved):

rochacbruno (approved):

galaxy_ng/app/api/v3/viewsets/collection.py
```diff
@@ -229,6 +230,15 @@ class CollectionArtifactDownloadView(api_base.APIView):
     permission_classes = [access_policy.CollectionAccessPolicy]
     action = 'retrieve'
 
+    def _get_tcp_response(self, url):
+        return requests.get(url, stream=True, allow_redirects=False)
+
+    def _get_unix_socket_response(self, url):
+        socket_file = settings.PULP_CONTENT_BIND.split(':')[1]
```
rochacbruno, line 237:
```suggestion
        socket_file = settings.CONTENT_BIND.split(':')[1]
```

awcrosby: thanks!

bmbouter:
This looks reasonable to me.

galaxy_ng/app/api/v3/viewsets/collection.py
```diff
@@ -229,6 +230,15 @@ class CollectionArtifactDownloadView(api_base.APIView):
     permission_classes = [access_policy.CollectionAccessPolicy]
     action = 'retrieve'
 
+    def _get_tcp_response(self, url):
+        return requests.get(url, stream=True, allow_redirects=False)
+
+    def _get_unix_socket_response(self, url):
+        socket_file = settings.CONTENT_BIND.split(':')[1]
+        session = requests.Session()
+        session.mount("http://", SocketHTTPAdapter(socket_file))
```
alikins, line 239: Could this session.mount() mount a 'unix://' adapter?

It's been a while since I used the requests / urllib3 Adapter stuff, but I was thinking you
could do that and then just pass in the 'unix://...'  "url" to request.get()

galaxy_ng/app/api/utils.py
```diff
@@ -53,3 +58,31 @@ def parse_collection_filename(filename):
         raise ValueError(f"Expected version to be max length of {MAX_LENGTH_VERSION}")
 
     return CollectionFilename(namespace, name, version)
+
+
```
alikins, line 62: First thought with this stuff was maybe add a galaxy_ng/app/common/unix_socket_http_adapter.py or whatever.

But I'm not sure there is really any place currently that is an obvious place, so /api/utils.py is probably as good as any.

I would be tempted to split it to it's own file just to keep the list of imports small, but doesn't matter a ton.

galaxy_ng/app/api/utils.py
```diff
@@ -53,3 +58,31 @@ def parse_collection_filename(filename):
...
+class SocketHTTPConnection(HTTPConnection):
+    def __init__(self, socket_file):
+        self.socket_file = socket_file
+        super().__init__(LOCALHOST)
+
+    def connect(self):
+        self.sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+        self.sock.connect(self.socket_file)
```
alikins, line 70: Does HTTPConnection() take care of closing self.sock?  (probably, but haven't verified).

Not sure if it would be an issue even without the close, but maybe worth verifying

PR #842, opened 2021-07-15
apply partial match to user filter

ShaiahWren:
Issue: AAH-355
Apply partial filter to user filter.

Before:
Screen Shot 2021-07-15 at 11 52 01 AM
After:
Screen Shot 2021-07-15 at 11 51 06 AM

### DIFF
--- a/galaxy_ng/app/api/ui/viewsets/user.py
+++ b/galaxy_ng/app/api/ui/viewsets/user.py
@@ -11,6 +11,14 @@ from galaxy_ng.app.api import base as api_base
 
 
 class UserFilter(filterset.FilterSet):
+    username = filters.CharFilter(field_name='username', lookup_expr='icontains')
+    email = filters.CharFilter(field_name='email', lookup_expr='icontains')
+    first_name = filters.CharFilter(field_name='first_name', lookup_expr='icontains')
+    last_name = filters.CharFilter(field_name='last_name', lookup_expr='icontains')
+    date_joined = filters.CharFilter(field_name='date_joined')
+    groups_name = filters.CharFilter(field_name='groups__name')
+    groups = filters.CharFilter(field_name='groups')
+
     sort = filters.OrderingFilter(
         fields=(
             ('username', 'username'),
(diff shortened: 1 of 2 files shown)


galaxy_ng/app/api/ui/viewsets/user.py
```diff
@@ -11,6 +11,14 @@
 
 
 class UserFilter(filterset.FilterSet):
+    username = filters.CharFilter(field_name='username', lookup_expr='icontains')
+    email = filters.CharFilter(field_name='email', lookup_expr='icontains')
+    first_name = filters.CharFilter(field_name='first_name', lookup_expr='icontains')
+    last_name = filters.CharFilter(field_name='last_name', lookup_expr='icontains')
+    date_joined = filters.CharFilter(field_name='date_joined')
+    groups_name = filters.CharFilter(field_name='groups__name')
```
rochacbruno, line 19: probably this needs to be `groups__name` but I guess this field is not being used so maybe we can remove, can you confirm @newswangerd ?

on the UI filters component there are only username, email, first_name, last_name

rochacbruno (approved):
👍🏻

galaxy_ng/app/api/ui/viewsets/user.py
```diff
@@ -11,6 +11,14 @@
 
 
 class UserFilter(filterset.FilterSet):
+    username = filters.CharFilter(field_name='username', lookup_expr='icontains')
```
newswangerd, line 14: We can't accept these changes because they are technically a breaking change (by changing the behaviour of an existing filter) and they aren't consistent with the rest of our filtering on the API.

If you want to update the UI to use icontains filtering instead of exact filtering, then the UI should be updated to perform `?<field_name>__icontains=foo` lookups instead of changing the default behavior of `?<fieldname>=foo` lookups.

There are cases where clients may need to look up a result using an exact match and replacing the exact filter with one that performs an icontains lookup under the hood will break that.

newswangerd (changes requested):

galaxy_ng/app/api/ui/viewsets/user.py
rochacbruno: @newswangerd so https://issues.redhat.com/browse/AAH-355 is an invalid issue?

newswangerd: @rochacbruno it's mostly valid. The changes should happen on the UI tho, not the API. the API already supported icontains lookups, you just had to do `?username__icontains` instead of `?username` for example.

rochacbruno: ok, so I think we can close this PR ? @ShaiahWren

ShaiahWren: @rochacbruno close the pr and DO NOT MERGE. haha. My new pr for this issue is over on https://github.com/ansible/ansible-hub-ui/pull/670

PR #844, opened 2021-07-19
Remove click requirement

bmclaughlin:
Remove `click` from our requirements list, it's no longer needed after updating to Python 3.8

Issue: AAH-603

### DIFF
--- a/CHANGES/603.misc
+++ b/CHANGES/603.misc
@@ -1,1 +1,1 @@
-Update the development environment to Python 3.8 and base image to ubi8:8.2
+Update the environments to Python 3.8 and update base image to ubi8.
--- a/setup.py
+++ b/setup.py
@@ -85,9 +85,6 @@ requirements = [
     "drf-spectacular",
     # pulp-container 2.6 requires pulpcore >=3.12.1
     "pulp-container>=2.7.0,<2.8.0",
-    # click 8 requires py38,
-    # can be removed once we require >=py38
-    "click==7.1.2",
 ]
 
 
@@ -121,7 +118,7 @@ setup(
     author="Red Hat, Inc.",
     author_email="info@ansible.com",
     url="https://github.com/ansible/galaxy_ng/",
-    python_requires=">=3.8",
+    python_requires=">=3.8,<3.9",
     setup_requires=["wheel"],
     install_requires=requirements,
     include_package_data=True,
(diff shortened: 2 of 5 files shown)


CHANGES/603.misc
```diff
@@ -1 +1 @@
-Update the development environment to Python 3.8 and base image to ubi8:8.2
\ No newline at end of file
+Update the development environment to Python 3.8, update base image to ubi8:8.2 and update requirements
```
awcrosby, line 1: Since you are editing this changelog snippet, may want to change to say ubi8:8.4 if that separate PR gets approved/merged. And may be worth removing the "development" piece since the change applies to more than just dev environment.

setup.py
```diff
@@ -121,7 +118,7 @@ def run(self):
     author="Red Hat, Inc.",
     author_email="info@ansible.com",
     url="https://github.com/ansible/galaxy_ng/",
-    python_requires=">=3.8",
+    python_requires="<3.9",
```
awcrosby, line 121: It looks like `python_requires` can use the same format as `requirements` https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires i.e. `python_requires=">=3.8,<3.9"` and I assume we want to require 3.8

awcrosby (changes requested):

setup.py
```diff
@@ -85,9 +85,6 @@ def run(self):
     "drf-spectacular",
     # pulp-container 2.6 requires pulpcore >=3.12.1
     "pulp-container>=2.7.0,<2.8.0",
-    # click 8 requires py38,
-    # can be removed once we require >=py38
-    "click==7.1.2",
```
awcrosby, line 90: The requirements files still have this requirement. I recommend not running `make requirements` which updates all to latest via pip, but running these which updates just based on setup.py (i will create a make endpoint to make this eaiser)
```
pip-compile --output-file=requirements/requirements.common.txt setup.py
pip-compile --output-file=requirements/requirements.insights.txt requirements/requirements.insights.in setup.py
pip-compile --output-file=requirements/requirements.standalone.txt requirements/requirements.standalone.in setup.py
```

awcrosby (changes requested):

newswangerd (approved):

awcrosby (approved):

PR #845, opened 2021-07-19
Make dev environment base image consistent with Dockerfile.rhel8

bmclaughlin:
No-Issue

### DIFF
--- a/dev/Dockerfile.base
+++ b/dev/Dockerfile.base
@@ -1,4 +1,4 @@
-FROM quay.io/app-sre/ubi8-ubi:8.2
+FROM registry.access.redhat.com/ubi8
 
 ARG USER_ID=1000
 ARG USER_NAME=galaxy


awcrosby (changes requested):
Please include an update for the dev environment UBI image also so we are using 8.4 everywhere.

Also the commit message needs `No-Issue` to pass the `Galaxy Custom CI / commit-check`

cutwater:
What is the reason for restricting the base image to 8.4?

dev/Dockerfile.base
```diff
@@ -1,4 +1,4 @@
-FROM quay.io/app-sre/ubi8-ubi:8.2
+FROM quay.io/app-sre/ubi8-ubi:8.4
```
cutwater, line 1: Why is the image from `app-sre` mirror is used?

cutwater (changes requested):

dev/Dockerfile.base
bmclaughlin: I incorrectly believed that the image from `registry.access.redhat.com` required either subscription or a vpn connection.  Further testing proved otherwise.  Thanks for catching this, updating.

bmclaughlin:
@cutwater I suppose the reasoning behind restricting to `8.4` was due to it being the currently recommended version as opposed to `8.2`. 

@awcrosby Any thoughts on restricting to `8.4` vs no restriction?

awcrosby:
@bmclaughlin i'd say go with no tag to match how we have been doing it

awcrosby (approved):

cutwater (approved):

PR #858, opened 2021-07-21
Add option to override existing UI when downloading UI from github

newswangerd:
Allows users to run `python setup.py prepare_static` with a `--force-download-ui` flag that removes any existing static files and re downloads them from github.

This will be used for setting up the UI in the pulp dev environment.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,9 +31,11 @@ class PrepareStaticCommand(Command):
     TARGET_DIR = "galaxy_ng/app/static/galaxy_ng"
 
     user_options = [
-        ('force-download-ui',
-        None,
-        'Replace any existing static files with the ones downloaded from github.'),
+        (
+            'force-download-ui',
+            None,
+            'Replace any existing static files with the ones downloaded from github.'
+        ),
     ]
 
     def initialize_options(self):
--- a/setup.py
+++ b/setup.py
@@ -4,6 +4,7 @@ import os
 import re
 import tempfile
 import tarfile
+import shutil
 import urllib.request
 import urllib.error
 from distutils import log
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -29,18 +30,26 @@ class PrepareStaticCommand(Command):
...
-    user_options = []
+    user_options = [
+        ('force-download-ui',
+        None,
+        'Replace any existing static files with the ones downloaded from github.'),
+    ]
 
     def initialize_options(self):
-        pass
+        self.force_download_ui = False
```
awcrosby, line 40: minor: since option `force-download-ui` gets mapped to var `self.force_download_ui`, is it necessary to initialize this as `self.force_download_ui = False` since the default is `None`?

setup.py
```diff
@@ -118,7 +127,7 @@ def run(self):
     author="Red Hat, Inc.",
     author_email="info@ansible.com",
     url="https://github.com/ansible/galaxy_ng/",
-    python_requires=">=3.8,<3.9",
+    python_requires=">=3.8",
```
awcrosby, line 130: may need to rebase as i think this change has already been committed to `master`

newswangerd: oopsie. Didn't mean to commit that

setup.py
newswangerd: I _think_ the second argument here

```
    user_options = [
        ('force-download-ui',
        None,
        'Replace any existing static files with the ones downloaded from github.'),
    ]
 ```

defines a single character flag like `-f`, not the default value.

newswangerd: yeah, it's long name, short name, description

https://dankeder.com/posts/adding-custom-commands-to-setup-py/

awcrosby: makes sense

awcrosby (approved):

PR #864, opened 2021-07-26
Fix missing ansible-importer configuration

cutwater:
No-Issue

### DIFF
--- a/CHANGES/815.bugfix
+++ b/CHANGES/815.bugfix
@@ -1,0 +1,1 @@
+Fixed missing galaxy-importer configuration in Clowder template.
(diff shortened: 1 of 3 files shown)


cutwater:
@jameswnl With this fix that requires adding the NAMESPACE parameter, we can move Route to the OpenShift template as well.

awcrosby (approved):

jameswnl:
As commented in slack, just a reminder:  `ConfigMap` needed to be in `managedResources` section in deploy-clowder.yml in app-int

jameswnl (approved):

PR #873, opened 2021-07-30
Move from travis to Github Actions CI

rochacbruno:
Workflow is now managed by pulp/plugin-template

Issue: AAH-460

Depends on: https://github.com/pulp/plugin_template/pull/463


.github/workflows/scripts/update_manifest.sh
```diff
@@ -1,11 +1,12 @@
...
 
+
 set -o nounset
 set -o errexit
 
 
-readonly TRAVIS_PULL_REQUEST="${TRAVIS_PULL_REQUEST:-}"
-readonly TRAVIS_BRANCH="${TRAVIS_BRANCH:-}"
+readonly GITHUB_PULL_REQUEST="${GITHUB_PULL_REQUEST:-}"
+readonly GITHUB_BRANCH="${GITHUB_BRANCH:-}"
```
fao89, line 9: Github variables are different,
these are some of the translations I tried: https://github.com/fao89/pulp_file/blob/6066/.github/workflows/ci.yaml#L47-L60

Ref: https://docs.github.com/en/actions/reference/environment-variables

pyproject.toml
```diff
@@ -22,3 +22,41 @@ exclude = '''
...
+ignore = [
+    ".bumpversion.cfg",
+    "CHANGES/**",
+    "CONTRIBUTING.rst",
+    "HISTORY.rst",
+    "dev_requirements.txt",
+    "doc_requirements.txt",
+    "functest_requirements.txt",
+    "unittest_requirements.txt",
+    "requirements/**",
```
fao89, line 36: maybe this can be helpful

pyproject.toml
```diff
@@ -22,3 +22,41 @@ exclude = '''
...
+    ".travis/**",
+    ".travis.yml",
+    ".flake8",
+    ".pep8speaks.yml",
+    ".ci/**",
+    ".github/**",
+    "dev/**",
+    "docker/**",
+    "openshift/**",
+    "galaxy_ng/app/static/**",
```
fao89, line 50: I believe you need this

.github/workflows/scripts/update_manifest.sh
rochacbruno: I exported those variables as part of the job 

```yaml
  env:
    GITHUB_PULL_REQUEST: 873
    GITHUB_BRANCH: ci_template_test
    GITHUB_REPO_SLUG: ansible/galaxy_ng
```

Those vars are present, the error is happening because file misses `+x` permission.

fao89 (approved):
LGTM, but I would confirm with the building team about the MANIFEST.in

pyproject.toml
rochacbruno: we have lots of requirements on that folder and they are specific to the docker environment `standalone|insights` I think it can be kept out and the py package carries only what is defined on setup.py

pyproject.toml
rochacbruno: added

pyproject.toml
rochacbruno: I included only the requirements/requirements.common.txt on the MANIFEST

newswangerd:
@rochacbruno I'm not sure where to start with this. Does the check to make sure if the issue exists on jira still work? I know that was custom to our setup.

awcrosby:
@newswangerd it looks like that check is still called, just within a Pulp CI github action:  `.github/workflows/scripts/check_commits.py → ..scripts/validate_commit_message_custom.py`

awcrosby (approved):

rochacbruno:
@newswangerd yes, that check runs as a pre-job, I added support for pre-jobs on this PR https://github.com/pulp/plugin_template/pull/463

newswangerd (approved):

rochacbruno:
@cutwater @newswangerd I need your help to understand this part.

We had previously on TRAVIS https://app.travis-ci.com/github/ansible/galaxy_ng/builds/233990489#L230 that variables were exported by https://github.com/ansible/galaxy_ng/blob/master/.travis.yml#L17

On Github Actions it looks like that variables might go to the repo settings: https://docs.github.com/en/actions/reference/encrypted-secrets
and then I will probably need to update this file https://github.com/ansible/galaxy_ng/pull/873/files#diff-2b9f00f4b9f3ba028931b4f521eace9a7d1a98308b185f55cf4d4887bb652a55 adding the matching variables.

newswangerd:
> @cutwater @newswangerd I need your help to understand this part.
> 
> We had previously on TRAVIS https://app.travis-ci.com/github/ansible/galaxy_ng/builds/233990489#L230 that variables were exported by https://github.com/ansible/galaxy_ng/blob/master/.travis.yml#L17
> 
> On Github Actions it looks like that variables might go to the repo settings: https://docs.github.com/en/actions/reference/encrypted-secrets
> and then I will probably need to update this file https://github.com/ansible/galaxy_ng/pull/873/files#diff-2b9f00f4b9f3ba028931b4f521eace9a7d1a98308b185f55cf4d4887bb652a55 adding the matching variables.

I think @cutwater wrote this so maybe he can elaborate. I think those secrets are for publishing the manifest, but I don't think there is a way to extract them from travis. We'll have to generate new tokens.

cutwater (commented):
@rochacbruno You're free to merge it if you will. I won't approve it, as it's to large to effectively review it.

cutwater:
As for uploading manifests, I don't think I have them backed up, so yes, we'll have to re-generate the token.

rochacbruno:
@awcrosby  @cutwater I don't have admin permissions to merge it, and this is blocked by the github branch protection rules.

I think after we merge it, we can change the rules and deal with the manifest secrets (it might require a change on the `post-job-template` file to point to the correct secrets variables.

PR #883, opened 2021-08-05
Update CI from Plugin Template

rochacbruno:
Depends on https://github.com/pulp/plugin_template/pull/470

No-Issue


.github/workflows/scripts/push_branch_and_tag_to_github.sh
```diff
@@ -7,7 +7,7 @@ ref_string=$(git show-ref --tags | grep refs/tags/$1)
 
 SHA=${ref_string:0:40}
 
-remote_repo=https://ansible:${RELEASE_TOKEN}@github.com/${GITHUB_REPOSITORY}.git
+remote_repo=https://pulpbot:${RELEASE_TOKEN}@github.com/${GITHUB_REPOSITORY}.git
```
fao89, line 10: I think you got the wrong user

template_config.yml
```diff
@@ -67,7 +68,8 @@ pydocstyle: true
 pypi_username: null
 python_version: '3.8'
 redmine_project: null
-release_user: ansible
+release_email: pulp-infra@redhat.com
+release_user: pulpbot
```
fao89, line 72: I believe you need to change this

.github/workflows/scripts/push_branch_and_tag_to_github.sh
rochacbruno: We don´t have yet an `galaxybot` user, the token should be created for the specific user? maybe we need to create that user and add to the org.

fao89: not sure, as we have pulpbot we always used it, maybe other user will work, but I never tested

PR #891, opened 2021-08-11
Add dependency django-automated-logging

chouseknecht:
Adds dependency django-automated-logging

Issue: AAH-849


cutwater (approved):
LGTM, though I'm concerned why do we need another not very well known dependency in the project.

PR #897, opened 2021-08-17
Try to run main/master branches on CI

rochacbruno:
Trying to see if we are able to default to `master/main` branches for repos on CI


rochacbruno:
I don't know why the error, locally the migrations work.

```
# check for any uncommitted migrations
echo "Checking for uncommitted migrations..."
cmd_prefix bash -c "django-admin makemigrations --check --dry-run"
Migrations for 'galaxy':
  usr/local/lib/python3.8/site-packages/galaxy_ng/app/migrations/0021_alter_user_first_name.py
    - Alter field first_name on user
Error: Process completed with exit code 1.
```
https://github.com/ansible/galaxy_ng/pull/897/checks?check_run_id=3356377778#step:11:1654

galaxy_ng/app/api/ui/viewsets/collection.py
```diff
@@ -59,6 +59,9 @@ def get_queryset(self):
             "collection_id",
             "version",
         )
+        deprecations = AnsibleCollectionDeprecated.objects.filter(
+            pk__in=self._distro_content
+        )
```
rochacbruno, line 64: @awcrosby @fao89 this change was needed to make it compatible with latest changes made on pulp_ansible, now Deprecated is a content type, but I need another eyes on this change to confirm I did the right thing.

Before it was

```
deprecated_query = AnsibleCollectionDeprecated.objects.filter(
            collection=OuterRef("collection"), repository_version=self._repository_version
        )
```

fao89: I believe you need reference something from the main query, maybe:
```suggestion
        deprecations = AnsibleCollectionDeprecated.objects.filter(
            namespace=OuterRef("namespace"), name=OuterRef("name"), pk__in=self._distro_content
        )
```

rochacbruno: done!

galaxy_ng/app/api/ui/viewsets/collection.py
```diff
@@ -70,16 +76,13 @@ def get_queryset(self):
             return CollectionVersion.objects.none().annotate(
                 # AAH-122: annotated fields must exist in all the returned querysets
                 #          in order for filters to work.
-                deprecated=Exists(AnsibleCollectionDeprecated.objects)
+                deprecated=Exists(deprecated_query)
```
fao89, line 79: shouldn't just return `deprecated=False`?

fao89 (approved):
Great work with the CI! 🚀

newswangerd (approved):
This looks fine to me. I'd like to get @awcrosby's feedback too

template_config.yml
```diff
@@ -1,33 +1,36 @@
...
-    branch: 2.7.1
-  - name: galaxy-importer
-    bindings: false
-    org: ansible
-    branch: v0.3.4
+- bindings: true
+  revision: "0b6440a9ddd1048fd1d6a8945c969a7c28d2a4c9"
+  branch: main
+  name: pulp_ansible
+  org: pulp
```
awcrosby, line 11: @rochacbruno if we push a PR with a `galaxy_ng` change that is compatible with `pulp_ansible` master at this revision, it will pass CI. But if that `galaxy_ng` change conflicts with the `pulp_ansible` version that is pinned in `setup.py`, do we have any way of knowing that?

rochacbruno: @awcrosby no, this PR makes master to be always the bleeding edge of repos, the versions pinned on setup.py for those 4 repos are bypassed and requirements files are completely ignored.

You have a good point and what we can do is:

```bash
if PR sent against master branch (${{ github.event.pull_request.base.ref == 'master' }}) 
then
   install from latest compatible main branches
else
   install from pinned versions from setup.py and requirements files
fi
```

with that logic on `.github/workflows/scripts/post_before_install.sh` we can ensure that a PR sent against other branches `stable-x.y.z` will always use pinned versions, but master will always be the latest.

We need master to catch latest updates to keep us in sync with pulp, but we cannot break this compatibility by adding a code change that is compatible only with previous versions. We will only know when the PR is sent against a stable-... branch.

rochacbruno: The other option, would be to have another branch called `next` and this PR would go there, and changes targetting `next` pulp release would be done there, this will generate the need to backport/cherry pick multiple things.

rochacbruno: this also would require more customization to pulp/plugin_template @fao89 @daviddavis I think we can add a conditional version matrix to the `pulpcore_branch|revision` and `additional_repos.branch|revision`.

e.g:

```yaml
pulpcore_branch:
    - name: master
      when: "${{ github.event.pull_request.base.ref == 'master' }}"
    - name: 3.14
      when: "${{ github.event.pull_request.base.ref != 'master' }}"

additional_repos:
    - name: pulp_ansible
      branch:
        - name: main
          when: "${{ github.event.pull_request.base.ref == 'master' }}"
        - name: 0.7.0
          when: "${{ github.event.pull_request.base.ref != 'master' }}"
```

then we need to add a new job to each item on matrix and condition goes to the job after it is rendered by Jinja

rochacbruno: We just talked about it and decided that we are adding this as a nightly build and not as a job to run for every PR, I will work on the changes.

cutwater (approved):

awcrosby (approved):

rochacbruno:
@awcrosby @cutwater  I see `You’re not authorized to merge this pull request.`

PR #898, opened 2021-08-18
new api endpoints for getting and listing container registries

ShaiahWren:
See issue: 
https://issues.redhat.com/browse/AAH-434

This pr creates new api endpoints for getting, listing, and updating container registries.


rochacbruno:
@ShaiahWren is thsi ready to be tested?
example: https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/tests/unit/api/test_api_ui_sync_config.py 

If we can add some testing following this example, simple tests just to crate the database entries and calls the api asserting fields are returned.

ShaiahWren:
@rochacbruno Yes, ready for testing. 

I will look into writing a sample test. May have some questions, stay tuned

galaxy_ng/app/api/ui/serializers/__init__.py
```diff
@@ -64,5 +66,7 @@
     'ContainerManifestSerializer',
     'ContainerManifestDetailSerializer',
     'ContainerReadmeSerializer',
-    'ContainerNamespaceDetailSerializer'
+    'ContainerNamespaceDetailSerializer',
+    'ContainerRegistryRemote',
```
awcrosby, line 70: we shouldn't need the model passed through the serializers, and it doesn't look like it is used in this PR, so it and the import can be removed.

galaxy_ng/app/api/ui/serializers/execution_environment.py
```diff
@@ -218,29 +200,74 @@ def _content_info(self, content):
...
+        ]
+        extra_kwargs = {
+            'name': {'read_only': True},
+            'client_key': {'write_only': True},
+        }
+
+    def get_write_only_fields(self, obj):
+        return utils.get_write_only_field(self, obj)
+
+    def get_last_sync_task_queryset(self, obj):
```
awcrosby, line 269: the serializer method name should be `get_<field_name>` so `get_last_sync_task` to match `last_sync_task`

galaxy_ng/app/api/ui/viewsets/execution_environment.py
```diff
@@ -210,3 +210,11 @@ class ContainerReadmeViewSet(ContainerContentBaseViewset):
     def get_object(self):
         distro = self.get_distro()
         return self.queryset.get_or_create(container=distro)[0]
+
+
+class SyncConfigContainerViewSet(
```
awcrosby, line 215: Can this viewset be renamed to `ContainerRegistryRemoteViewSet` to more closely match the name of the model `ContainerRegistryRemote` and serializer `ContainerRegistryRemoteSerializer`?

galaxy_ng/app/api/ui/serializers/execution_environment.py
awcrosby: Oh I think I see now... the `LastSyncTaskMixin` has `get_last_sync_task()` already defined, but needs the `get_last_sync_task_queryset()` implemented.

galaxy_ng/app/api/ui/urls.py
```diff
@@ -64,6 +64,19 @@
         "namespaces/",
         viewsets.ContainerNamespaceViewSet.as_view({'get': 'list'}),
         name='container-namespace-list'),
+    path(
+        "registries/<str:pk>/",
+        viewsets.SyncConfigContainerViewSet.as_view({'get': 'retrieve', 'put': 'update'}),
+        name='execution-environments-registries-detail'),
```
awcrosby, line 70: I _think_ the convention here is to have this singular, so `registry-detail` and `registry-list`

rochacbruno (changes requested):
Linter is failing, run `flake8 --config flake8.cfg` locally to see where changes are needed.

+1 to @awcrosby suggestions

galaxy_ng/app/api/ui/urls.py
```diff
@@ -64,6 +64,19 @@
...
+        "registries/<str:pk>/",
+        viewsets.ContainerRegistryRemoteViewSet.as_view({'get': 'retrieve', 'put': 'update'}),
+        name='execution-environments-registries-detail'),
+    path(
+        "registries/",
+        viewsets.ContainerRegistryRemoteViewSet.as_view({'get': 'list', 'post': 'create'}),
+        name='execution-environments-registries-list'),
+    # path(
+    #     "registries/<name>/sync/",
+    #     viewsets.SyncConfigContainerViewSet.as_view({'post': 'create',}),
```
newswangerd, line 77: Lets delete this bit, since this endpoint is coming in a future PR.

newswangerd (commented):
Looks good so far. I'll re review when tests are finished.

ShaiahWren:
@awcrosby @newswangerd The edits are in and the tests are completed. Lmk when you can take a look

galaxy_ng/app/api/ui/serializers/execution_environment.py
awcrosby: @ShaiahWren I don't think any changes were needed here, but it looks like `get_last_sync_task_queryset()` was renamed to `get_last_sync_task()`. By using `get_last_sync_task_queryset()` we get a [standard set of data](https://github.com/ansible/galaxy_ng/blob/b483f2780d1ae21c6d6e922b0595b757758e0d12/galaxy_ng/app/api/v3/serializers/sync.py#L40-L46) rather than whatever it displays with just returning just an object in the `last_sync_task` field

galaxy_ng/tests/unit/api/test_api_ui_container_registry_remote.py
```diff
@@ -0,0 +1,90 @@
...
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+        return response.data['pk']
+
+    def test_get_detail_with_data(self):
+        name = 'my_registry'
+        pk = self.create_registry_remote(name)
+        url = get_current_ui_url('execution-environments-registry-detail', kwargs={'pk': pk})
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response.data['name'], name)
```
awcrosby, line 47: I recommend testing at at least one other field. You could test something with the `last_sync_task` field, that would actually fail when you rename the `ui/serializers/execution_environment.py` method from `last_sync_task_queryset()` to `last_sync_task()` which is really helpful - this test could check the format is as expected.

It would probably look something like this: `self.assertIn("task_id", response.data['last_sync_task'])`

galaxy_ng/tests/unit/api/test_api_ui_container_registry_remote.py
```diff
@@ -0,0 +1,90 @@
...
+    def test_update_detail_fail(self):
+        initial_name = 'intial_registry_name'
+        updated_name = 'updated_registry_name'
+        pk = self.create_registry_remote(initial_name)
+        url = get_current_ui_url('execution-environments-registry-detail', kwargs={'pk': pk})
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response.data['name'], initial_name)
+        # logic for updating the registry
+        response = self.client.put(url, {'name' : updated_name}, format='json')
```
awcrosby, line 89: it might be good to put a comment here on why this fails, i assume it is failing because `url` is required in the PUT, not for example because you are doing a PUT to a `pk` that does not exist

awcrosby (changes requested):

galaxy_ng/tests/unit/api/test_api_ui_container_registry_remote.py
ShaiahWren: The registry that we made with the helper function has a 'last_sync_task' of None, should I just assign it a number? or generate it some other way?

awcrosby: @ShaiahWren not sure if you saw my other comment about `get_last_sync_task_queryset()`... in the django api page see what the `last_sync_task` field looks like before / after that change

ShaiahWren: Screen Shot 2021-08-23 at 2 28 54 PM

awcrosby: Seems we cannot test format of sync task here in tests since a sync task will not exist, this test LGTM

awcrosby (approved):
code changes all LGTM

PR #908, opened 2021-08-30
Allow running latest/main branches or revision on PR CI.

rochacbruno:
This is already possible to run a PR CI checks using specific pull-request for repos,
for example, one can add:

```bash
Required PR: https://github.com/pulp/pulpcore/pulls/456
Required PR: https://github.com/ansible/galaxy-importer/pulls/999
```

And the CI will checkout to that specific PR before installing and running the tests.

Right now on galaxy_ng if you try it, you will get a **failure on installation** because pip will Conflict on deps version, to solve that add to commit msg also `env:LOCK_REQUIREMENTS=0` and this will work similar to what we have on our docker based environment, it will remove the pinned versions from galaxy_ng/setup.py and allow pip to install.

So the commit message would contain:

```bash
Required PR: https://github.com/pulp/pulpcore/pulls/456
Required PR: https://github.com/ansible/galaxy-importer/pulls/999
env:LOCK_REQUIREMENTS=0
```

This is very useful when you are working on a feature that depends on unmerged Pull Request.

---

This PR also adds more ways of specifying the checkout for the dependency repos.

**Forcing specific git revision for a repo.**

Add `env:PULPCORE_REVISION=36692a8c628a5d5af220fb1c7bede712727e1e89` to the commit message, this is similar to the `Required PR` but works for any revision, so you can also checkout to a point in the past to run the CI.
 
The above will require the `env:LOCK_REQUIREMENTS=0` on your latest commit message.

--- 

**Shortcut to latest**

If all you want is to run your galaxy_ng PR against all repos checked out at master/main branches ensure you have on your commit message the `RUN_ON_LATEST` variable set to 1.

```bash
This is my latest commit message

- Solves X, Y and Z

Issue: AAH-xxx

env:RUN_ON_LATEST=1
```

(this will set each REPO_REVISION=latest and also ensure the LOCK_REQUIREMENTS=0)

---

**Plus: Added a NIghtly Job which runs pointing to all the latest repo branches.**

Every night 4AM a job will run testing latest branches together, we expect this job to fail most of the times, it is here just to enable us to see earlier when things starts breaking and have the opportunity to fix soon.

---

> NOTE: When running PRs using `master/main` branches, probably the `ci.ext.devshift.net` check will fail, and that is expected as we don't want to have those changes merged in before fully compatible with the release stream we are running. The better is to have this kind of PRs opened as drafts with purpose to get review and feedback earlier.

This is an example of PR using this feature: https://github.com/ansible/galaxy_ng/pull/902 (PR CI uses pulpcore master)

Issue: AAH-871

### DIFF
--- a/.github/workflows/nightly_latest.yml
+++ b/.github/workflows/nightly_latest.yml
@@ -1,0 +1,129 @@
+name: Nightly Latest
+on:
+  schedule:
+    # * is a special character in YAML so you have to quote this string
+    # runs at 4:00 UTC daily
+    - cron: '00 4 * * *'
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        env:
+          - TEST: pulp
+          - TEST: s3
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          # by default, it uses a depth of 1
+          # this fetches all history so that we can read each commit
+          fetch-depth: 0
+
+      - uses: actions/setup-python@v2
+        with:
+          python-version: "3.8"
+
+      - name: Install httpie
+        run: |
+          echo ::group::HTTPIE
+          sudo apt-get update -yq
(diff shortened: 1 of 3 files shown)


awcrosby (approved):
LGTM ... this is really flexible and convenient!

PR #909, opened 2021-08-30
Enable view-only collection browsing

bmclaughlin:
Set the `GALAXY_ENABLE_VIEW_ONLY_ACCESS` in `galaxy_ng/app/settings.py` to `True` and load Hub.
Issue: AAH-881

### DIFF
--- a/galaxy_ng/app/access_control/access_policy.py
+++ b/galaxy_ng/app/access_control/access_policy.py
@@ -63,6 +63,9 @@ class AccessPolicyBase(AccessPolicy):
 class NamespaceAccessPolicy(AccessPolicyBase):
     NAME = 'NamespaceViewSet'
 
+    def view_only_mode_enabled(self, request, view, permission):
+        return settings.GALAXY_ENABLE_VIEW_ONLY_ACCESS
+
 
 class CollectionAccessPolicy(AccessPolicyBase):
     NAME = 'CollectionViewSet'
@@ -79,6 +82,9 @@ class CollectionAccessPolicy(AccessPolicyBase):
         except models.Namespace.DoesNotExist:
             raise NotFound(_('Namespace in filename not found.'))
         return request.user.has_perm('galaxy.upload_to_namespace', namespace)
+
+    def view_only_mode_enabled(self, request, view, permission):
+        return settings.GALAXY_ENABLE_VIEW_ONLY_ACCESS
 
 
 class CollectionRemoteAccessPolicy(AccessPolicyBase):
@@ -101,6 +107,9 @@ class MyUserAccessPolicy(AccessPolicyBase):
 
     def is_current_user(self, request, view, action):
         return request.user == view.get_object()
+
+    def view_only_mode_enabled(self, request, view, action):
+        return settings.GALAXY_ENABLE_VIEW_ONLY_ACCESS
 
 
 class SyncListAccessPolicy(AccessPolicyBase):
--- a/galaxy_ng/app/access_control/access_policy.py
+++ b/galaxy_ng/app/access_control/access_policy.py
@@ -6,7 +6,7 @@ from rest_access_policy import AccessPolicy
 from rest_framework.exceptions import NotFound
 
 from galaxy_ng.app import models
-from galaxy_ng.app.access_control.mixins import ViewOnlyMixin
+from galaxy_ng.app.access_control.mixins import UnauthenticatedCollectionAccessMixin
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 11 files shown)


galaxy_ng/app/api/ui/views/auth.py
```diff
@@ -44,6 +46,10 @@ def get(self, request, *args, **kwargs):
 
     def post(self, request, *args, **kwargs):
         serializer = self.get_serializer(data=self.request.data)
+
+        if str(request.user) == "AnonymousUser" and GALAXY_ENABLE_VIEW_ONLY_ACCESS:
```
rochacbruno, line 50: Here it is also possible to read it from django settings

```suggestion
        if str(request.user) == "AnonymousUser" and settings.GALAXY_ENABLE_VIEW_ONLY_ACCESS:
```

galaxy_ng/app/settings.py
```diff
@@ -134,6 +134,8 @@
 BASE_DIR = os.path.dirname(os.path.dirname(__file__))
 LOCALE_PATHS = (os.path.join(BASE_DIR, 'locale'),)
 
+GALAXY_ENABLE_VIEW_ONLY_ACCESS = os.getenv('GALAXY_ENABLE_VIEW_ONLY_ACCESS', default=False)
```
rochacbruno, line 137: If you default to False here:

```suggestion
GALAXY_ENABLE_VIEW_ONLY_ACCESS = False
```

Then users can overwrite it on `/etc/pulp/settings.py`
It will allow the pulp installer to have 
```yaml
pulp_settings:
  galaxy_enable_view_only_access: true
```

and also

```
export PULP_GALAXY_ENABLE_VIEW_ONLY_ACCESS=true
```

galaxy_ng/app/access_control/statements/standalone.py
```diff
@@ -1,8 +1,32 @@
+from galaxy_ng.app.settings import GALAXY_ENABLE_VIEW_ONLY_ACCESS
```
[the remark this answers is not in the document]
bmclaughlin: I tried importing it via `django.conf` to originally but kept running into `AttributeError: 'Settings' object has no attribute 'GALAXY_ENABLE_VIEW_ONLY_ACCESS'`.  Using `settings.get('')` results in a `AttributeError: 'Settings' object has no attribute 'get'`.

rochacbruno:
LGTM, however I think it would be good if we can add some tests


It might be possible to write a test for this feature by overriding the settings and forcing the reload of access policies.


```py
import galaxy_ng

class ViewOnlyTestCase(TestCase):
    def test_unauthenticated_user_can_browse_collections(self):
        with self.settings(VIEW_ONLY_ACCESS...=True):
            import galaxy_ng
            galaxy_ng.app.AccessPolicyBase._STATEMENTS = None   # of this doesn't work then mocking can be a solution
            # write a test using the client and try to browse collections via API
```

galaxy_ng/app/access_control/access_policy.py
```diff
@@ -63,6 +63,9 @@ def has_rh_entitlements(self, request, view, permission):
 class NamespaceAccessPolicy(AccessPolicyBase):
     NAME = 'NamespaceViewSet'
 
+    def view_only_mode_enabled(self, request, view, permission):
+        return settings.GALAXY_ENABLE_VIEW_ONLY_ACCESS
```
rochacbruno, line 67: Suggestion:

This method repeats on every other Policy

```py
class ViewOnlyMixin:
    def view_only_mode_enabled(self, request, view, permission):
        return settings.GALAXY_ENABLE_VIEW_ONLY_ACCESS
```

then

```py
class CollectionAccessPolicy(ViewOnlyMixin, AccessPolicyBase):
    NAME = 'CollectionViewSet'
```

rochacbruno (approved):

galaxy_ng/app/api/ui/views/auth.py
```diff
@@ -44,6 +45,10 @@ def get(self, request, *args, **kwargs):
 
     def post(self, request, *args, **kwargs):
         serializer = self.get_serializer(data=self.request.data)
+
+        if str(request.user) == "AnonymousUser" and settings.GALAXY_ENABLE_VIEW_ONLY_ACCESS:
```
newswangerd, line 49: What is this in here for? It seems like this should break authentication when running in view only mode.

galaxy_ng/app/settings.py
```diff
@@ -137,6 +137,8 @@
 BASE_DIR = os.path.dirname(os.path.dirname(__file__))
 LOCALE_PATHS = (os.path.join(BASE_DIR, 'locale'),)
 
+GALAXY_ENABLE_VIEW_ONLY_ACCESS = False
```
newswangerd, line 140: I'm trying to think about the best way to set up the configuration for this. I'm leaning towards renaming this to `GALAXY_ENABLE_UNAUTHENTICATED_COLLECTION_ACCESS`, since this doesn't enable access to containers. I know this wasn't a requirement from product but I would also like to add a `GALAXY_ENABLE_UNAUTHENTICATED_COLLECTION_DOWNLOAD` option that lets people download collections with an anonymous user as well. Don't know if customers want this or not, but doesn't hurt to give people options.

This opens the door for enabling view only EE access as well if we want.

newswangerd (changes requested):
Looking good!

galaxy_ng/app/api/ui/views/auth.py
bmclaughlin: Doh!  Had that in there when testing things, forgot to remove it.  Thanks!

newswangerd (approved):
Nice work!

PR #918, opened 2021-09-01
Clowder: Add pulp encryption secret

cutwater:
Mount pulp encryption secret in ClowdApp and database migration
template.

NOTES:
  1. For stage and production deployment a secret must exist
     before the deployment is executed.
  2. Migtation template ref must be updated along with image tag
     in the app-interface to pick up changes in this commit.

No-Issue

### DIFF
--- a/openshift/clowder/clowd-app.yaml
+++ b/openshift/clowder/clowd-app.yaml
@@ -5,6 +5,15 @@ metadata:
   name: automation-hub
 objects:
 - apiVersion: v1
+  kind: Secret
+  metadata:
+    name: pulp-key
+    namespace: automation-hub
+  data:
+    database_fields.symmetric.key: |
+      DNmNdwgyZugTax9S64J0FITTr9IHPxbuoF1F1CGPr68=
+
+- apiVersion: v1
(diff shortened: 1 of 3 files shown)


openshift/clowder/clowd-app.yaml
```diff
@@ -4,6 +4,15 @@ kind: Template
...
   name: automation-hub
 objects:
+- apiVersion: v1
+  kind: Secret
+  metadata:
+    name: pulp-key
+    namespace: automation-hub
+  data:
+    database_fields.symmetric.key: |
+      DNmNdwgyZugTax9S64J0FITTr9IHPxbuoF1F1CGPr68=
```
rochacbruno, line 14: Our security teams runs monitoring for leaked keys.

To avoid this to be catched by the monitoring we have this file https://github.com/ansible/galaxy_ng/blob/master/.gitleaks.toml#L3

We need to add `openshift/clowder/clowd-app.yaml` to that list of paths.

rochacbruno (changes requested):

rochacbruno (approved):

PR #920, opened 2021-09-01
galaxy custom access control statements should be lazily loaded.

rochacbruno:
Tl:DR; 

```py
# doing this import, as this module is not part of django
from galaxy_ng.app.access_control.statements import PULP_CONTAINER_VIEWSETS, ANYTHINGELSE
# happens before django.apps.ready() is executed
# So pulp settings is not loaded yet.
```

---

Galaxy has a custom accesscontrol statements module
if that module is imported before the django settings is configured
by pulp/dynaconf, then it is impossible to access settings on that
module.

It is also safer to load extra-django things only after the app.ready()
is executed.

No-Issue


bmclaughlin (approved):
Fixes the issue that I ran into in #909 when accessing `django.settings`

PR #938, opened 2021-09-07
Use paths, not files syntax for gitleaks

rooftopcellist:
The `files` only variable only looks at individual file names, regardless of path.  We should be using the `path` variable here in the `.gitleaks.toml` conf.

### DIFF
--- a/.gitleaks.toml
+++ b/.gitleaks.toml
@@ -1,6 +1,6 @@
 [allowlist]
   description = "Our test install exports a test only MINIO ACCESS KEY"
-  files = [
+  paths = [
     ".github/workflows/scripts/install.sh",
     "openshift/clowder/clowd-app.yaml",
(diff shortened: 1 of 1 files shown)


awcrosby:
@rooftopcellist the commit signature is showing as `Unverified`

awcrosby:
@rooftopcellist our `check-commit` CI job needs a linkage to JIRA in commit message https://github.com/ansible/galaxy_ng/wiki/Contributing#preparing-pull-requests or in this case since there is none: `No-Issue`

rooftopcellist:
@awcrosby thanks for the reminder, I have it in my commit message at the bottom.  Hopefully that is the right syntax.

rochacbruno (approved):

awcrosby (approved):

gundalow:
Thanks for fixing this

PR #939, opened 2021-09-07
Collection Version Deletion

rochacbruno:
Allow DELETE requests on collection version endpoint.

1. Perform Dependency Check to verify that the collection version can be deleted
2. If it can, perform Collection Repository Cleanup tasks
3. If the collection version can’t be deleted, return the reason why
4. If the version being deleted is the last collection version in the collection,
   remove the collection object as well.

Issue: AAH-710

### DIFF
--- a/galaxy_ng/app/api/v3/viewsets/collection.py
+++ b/galaxy_ng/app/api/v3/viewsets/collection.py
@@ -2,6 +2,9 @@ import logging
 
 import requests
 
+from packaging.specifiers import SpecifierSet
+from packaging.version import Version
+
 from django.core.exceptions import ValidationError, ObjectDoesNotExist
 
 from django.conf import settings
@@ -10,8 +13,10 @@ from django.shortcuts import get_object_or_404
 from django.urls import reverse
 from django.utils.translation import gettext_lazy as _
 
+from rest_framework.request import Request
 from rest_framework.response import Response
 from rest_framework.exceptions import APIException, NotFound
+from rest_framework.mixins import DestroyModelMixin
 
 from pulpcore.plugin.models import Task
 from pulpcore.plugin.tasking import dispatch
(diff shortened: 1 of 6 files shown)


galaxy_ng/app/api/v3/viewsets/collection.py
```diff
@@ -87,21 +92,66 @@ class CollectionViewSet(api_base.LocalSettingsMixin,
...
+                    ).format(
+                        namespace=collection_version.namespace,
+                        name=collection_version.collection.name,
+                    ),
+                    "dependent_collections": [f"{dep.namespace}.{dep.name}" for dep in dependents],
+                },
+                status=400,
+            )
+        # collection repo cleanup
+
```
rochacbruno, line 151: @awcrosby the **Collection Repo Cleanup** tasks are described in Content Deletion arch doc.

on this under review PR https://github.com/ansible/galaxy_ng/pull/958 there is a script to generate some collections with dependencies.

```bash
django-admin create-test-collections --strategy=test --prefix=710 -y
```

You gonna feed your system with collections and dependencies to test.

galaxy_ng/app/api/v3/viewsets/collection.py
```diff
@@ -87,21 +92,66 @@ class CollectionViewSet(api_base.LocalSettingsMixin,
...
 
 
+def get_dependents(parent):
+    """Given a parent collection, return a list of collections that depend on it."""
+    key = f"{parent.namespace}.{parent.name}"
+    dependents = []
+    for child in CollectionVersion.objects.filter(dependencies__has_key=key):
+        spec = SpecifierSet(child.dependencies[key])
+        if Version(parent.version) in spec:
+            dependents.append(child)
```
awcrosby, line 113: @rochacbruno in working this PR, I plan to change the version specificer comparison library from `packaging` to `semantic_version` which we use in `galaxy-importer` and a bit in `galaxy_ng`.

For example we have existing dependencies in the format of `ansible.netcommon: '1.0.0'` but for that case it looks like packaging expects `ansible.netcommon: '==1.0.0'`

awcrosby:
Moving this to #962

PR #941, opened 2021-09-07
Add controllers endpoint for listing connected controllers

newswangerd:
Adds a new `CONNECTED_ANSIBLE_CONTROLLERS` setting that allows users to specify a list of controller hosts that they wish to connect to.

Sample configuration
```
CONNECTED_ANSIBLE_CONTROLLERS = [
            "https://my-controller1.example.com/",
            "https://controllerino.contotrolisimo.com/",
            "https://boring-url.example.com/",
        ]
```

Adds new `_ui/v1/controllers` endpoint, which returns a paginated list of the controller instances configured in settings. This endpoint accepts a `host`  and `host__icontains` filter for searching for specific controller instances.


himdel (approved):
Works for me :+1:

It doesn't seem to read `CONNECTED_ANSIBLE_CONTROLLERS` from `.compose.env` (with or without a `PULP_` prefix), but modifying `app/settings.py` works :)

newswangerd:
> Works for me 👍
> 
> It doesn't seem to read `CONNECTED_ANSIBLE_CONTROLLERS` from `.compose.env` (with or without a `PULP_` prefix), but modifying `app/settings.py` works :)

@himdel the environment variables in .compose.env don't get passed into the container

rochacbruno:
> @himdel the environment variables in .compose.env don't get passed into the container
@newswangerd 

Do we want to fix it? I think it is possible to do that.

newswangerd:
> > @himdel the environment variables in .compose.env don't get passed into the container
> > @newswangerd
> 
> Do we want to fix it? I think it is possible to do that.

@rochacbruno this could be helpful. I'm not totally sure how you'd do it though. You'd have to find all env variables that start with `PULP_*` and then dynamically modify that docker compose file to include them in the container.

PR #942, opened 2021-09-07
Allow configuration of KEYCLOAK_HOST_LOOPBACK env var

ironfroggy:
No-Issue


ironfroggy:
I run my local-dev on a separate machine and needed this to get the auth redirects working on a specified IP address instead of the default.

chambridge:
> I run my local-dev on a separate machine and needed this to get the auth redirects working on a specified IP address instead of the default.

I don't have a problem with this, but you should probably also add it to the `.compose.env.example` file

chambridge (changes requested):
I think you need No-Issue in the commit message

chambridge (approved):
LGTM

PR #944, opened 2021-09-09
add filtering to container registries endpoint

ShaiahWren:
Issue: AAH-897

This pr adds filtering to the container registries endpoint.

### DIFF
--- a/galaxy_ng/app/api/ui/viewsets/execution_environment.py
+++ b/galaxy_ng/app/api/ui/viewsets/execution_environment.py
@@ -212,9 +212,29 @@ class ContainerReadmeViewSet(ContainerContentBaseViewset):
         return self.queryset.get_or_create(container=distro)[0]
 
 
+class ContainerRegistryRemoteFilter(filterset.FilterSet):
+    name = filters.CharFilter(field_name='name')
+    url = filters.CharFilter(field_name='url')
+
+    sort = filters.OrderingFilter(
+        fields=(
+            ('name', 'name'),
+            ('url', 'url'),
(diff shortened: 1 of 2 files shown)


galaxy_ng/app/api/ui/viewsets/execution_environment.py
```diff
@@ -212,9 +212,31 @@ def get_object(self):
         return self.queryset.get_or_create(container=distro)[0]
 
 
+class ContainerRegistryRemoteFilter(filterset.FilterSet):
+    name = filters.CharFilter(field_name='name')
+    url = filters.CharFilter(field_name='url')
+    name_icontains = filters.CharFilter(field_name='name', lookup_expr='icontains')
+    url_icontains = filters.CharFilter(field_name='url', lookup_expr='icontains')
```
rochacbruno, line 219: Here I think you don't need to specify the lookup expr, just the field.

```suggestion
    name = filters.CharFilter(field_name='name')
    url = filters.CharFilter(field_name='url')
```

Because on the `class Meta:` there is already a list of supported lookup expressions.

On api call that will enable `?name__icontains=foo` and `?name=foobar` and `?name__startswith=fo` 

I think @newswangerd knows more about filtering, but I am guessing this is how it works.

awcrosby (commented):
if the suggestion from @rochacbruno works in testing then this LGTM

ShaiahWren:
@newswangerd Want to weigh in? I think this might be ready to merge :)

newswangerd (approved):

PR #945, opened 2021-09-09
Fix Keycloak conditional settings using Dynaconf Hooks

rochacbruno:
KEYCLOAK settings was not being loaded from /etc/pulp/settings.py

More info: https://github.com/rochacbruno/dynaconf/issues/654 Dynaconf hooks are available only on 3.1.6+

Issue: AAH-915


galaxy_ng/app/dynaconf_hooks.py
```diff
@@ -0,0 +1,237 @@
+import os
+from typing import Any, Dict
+
+from dynaconf import Dynaconf
+
+
+def post(settings: Dynaconf) -> Dict[str, Any]:
+    """The dynaconf post hook is called after all the settings are loaded and set.
+
+    Post hook is necessary when a setting key depends conditionally on a previouslys et variable.
```
chambridge, line 10:
```suggestion
    Post hook is necessary when a setting key depends conditionally on a previously set variable.
```

chambridge (approved):
One typo, but otherwise looks good.

ironfroggy (approved):

PR #946, opened 2021-09-09
Add configuration for API access logging to dev environment

bmclaughlin:
Adds logging file with permissions and owner.  Now a dev only needs to update the `GALAXY_ENABLE_API_ACCESS_LOG` to enable logging in the container dev environment.

No-Issue

### DIFF
--- a/dev/standalone/Dockerfile
+++ b/dev/standalone/Dockerfile
@@ -11,6 +11,9 @@ RUN set -ex; \
 
 USER root
 
-RUN dnf install -y gettext
+RUN dnf install -y gettext; \
+    touch /var/log/galaxy_api_access.log \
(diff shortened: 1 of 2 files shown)


dev/standalone/Dockerfile
```diff
@@ -11,6 +11,9 @@ RUN set -ex; \
 
 USER root
 
-RUN dnf install -y gettext
+RUN dnf install -y gettext; \
```
rochacbruno, line 14: Maybe it can go in https://github.com/ansible/galaxy_ng/blob/master/dev/Dockerfile.base ?

So it is reused by insights, keycloak and other images?

rochacbruno (approved):

PR #947, opened 2021-09-09
Pass .compose.env variables to the containers

rochacbruno:
docker compose allows `env_file` to be a list of .env files
those files are loaded in order.

It makes easier to have custom local settings.

Ex:

`.compose.env`
```bash
PULP_GALAXY_ENABLE_API_ACCESS_LOG=true
```

will enable logs to be written to `/var/log/galaxy_api_access.log`

No-Issue

### DIFF
--- a/dev/standalone-keycloak/docker-compose.yml
+++ b/dev/standalone-keycloak/docker-compose.yml
@@ -5,18 +5,21 @@ services:
   api:
     env_file:
       - './standalone-keycloak/galaxy_ng.env'
+      - '../.compose.env'
     environment:
       SOCIAL_AUTH_KEYCLOAK_PUBLIC_KEY: ${SOCIAL_AUTH_KEYCLOAK_PUBLIC_KEY}
       KEYCLOAK_HOST_LOOPBACK: ${KEYCLOAK_HOST_LOOPBACK}
   worker:
     env_file:
       - './standalone-keycloak/galaxy_ng.env'
+      - '../.compose.env'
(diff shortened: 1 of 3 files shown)


awcrosby (approved):

dev/standalone-keycloak/docker-compose.yml
```diff
@@ -5,18 +5,21 @@ services:
   api:
     env_file:
       - './standalone-keycloak/galaxy_ng.env'
+      - '../.compose.env'
     environment:
       SOCIAL_AUTH_KEYCLOAK_PUBLIC_KEY: ${SOCIAL_AUTH_KEYCLOAK_PUBLIC_KEY}
       KEYCLOAK_HOST_LOOPBACK: ${KEYCLOAK_HOST_LOOPBACK}
```
chambridge, line 11: The assumption is that these values would come in from the `.compose.env` so you could probably remove each of this from the file below.

bmclaughlin (approved):

dev/standalone-keycloak/docker-compose.yml
rochacbruno: @chambridge this other PR is changing it a little https://github.com/ansible/galaxy_ng/pull/945 the variable will need to be prefixed with `PULP_`

chambridge: ok.

chambridge (approved):

PR #948, opened 2021-09-09
Disable single_commit_check

daviddavis:
No-Issue


rochacbruno (approved):
Nice!

Thanks for the PR

We allow multiple commits on PRs because we enforce `squash and merge` and for huge PRs having multi commits makes easier for reviewer.

Also changes applied after a reviewer request are easier to detect in a subsequent commit.

PR #961, opened 2021-09-13
Clowder: add prometheus_multiproc_dir

drodowic:


awcrosby:
@drodowic you will need to add `No-Issue` as part of your commit message (since there is no jira issue associated with this change)

awcrosby (approved):
LGTM after CI passes

PR #962, opened 2021-09-14
Support deleting a collection version

awcrosby:
Allow DELETE requests on collection version endpoint.

Continuing from #939

Issue: AAH-710

### DIFF
--- a/galaxy_ng/app/api/v3/viewsets/collection.py
+++ b/galaxy_ng/app/api/v3/viewsets/collection.py
@@ -2,6 +2,9 @@ import logging
 
 import requests
 
+from packaging.specifiers import SpecifierSet
+from packaging.version import Version
+
 from django.core.exceptions import ValidationError, ObjectDoesNotExist
 
 from django.conf import settings
@@ -10,8 +13,10 @@ from django.shortcuts import get_object_or_404
 from django.urls import reverse
 from django.utils.translation import gettext_lazy as _
 
+from rest_framework.request import Request
 from rest_framework.response import Response
 from rest_framework.exceptions import APIException, NotFound
+from rest_framework.mixins import DestroyModelMixin
 
 from pulpcore.plugin.models import Task
 from pulpcore.plugin.tasking import dispatch
@@ -87,12 +92,26 @@ class CollectionViewSet(api_base.LocalSettingsMixin,
     serializer_class = CollectionSerializer
 
 
-class UnpaginatedCollectionVersionViewSet(api_base.LocalSettingsMixin,
-                                          ViewNamespaceSerializerContextMixin,
-                                          pulp_ansible_views.UnpaginatedCollectionVersionViewSet):
+class UnpaginatedCollectionVersionViewSet(
+    api_base.LocalSettingsMixin,
+    ViewNamespaceSerializerContextMixin,
+    DestroyModelMixin,
+    pulp_ansible_views.UnpaginatedCollectionVersionViewSet,
+):
     pagination_class = None
     serializer_class = UnpaginatedCollectionVersionSerializer
     permission_classes = [access_policy.CollectionAccessPolicy]
+
+
+def get_dependents(parent):
+    """Given a parent collection, return a list of collections that depend on it."""
+    key = f"{parent.namespace}.{parent.name}"
+    dependents = []
+    for child in CollectionVersion.objects.filter(dependencies__has_key=key):
+        spec = SpecifierSet(child.dependencies[key])
+        if Version(parent.version) in spec:
+            dependents.append(child)
+    return dependents
 
 
 class CollectionVersionViewSet(api_base.LocalSettingsMixin,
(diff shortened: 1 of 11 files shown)


galaxy_ng/app/api/v3/viewsets/collection.py
```diff
@@ -87,21 +92,71 @@ class CollectionViewSet(api_base.LocalSettingsMixin,
...
+                status=400,
+            )
+
+        dispatch(
+            delete_collection_version,
+            exclusive_resources=collection_version.repositories.all(),
+            kwargs={"collection_version_pk": collection_version.pk},
+        )
+
+        return Response(status=status.HTTP_202_ACCEPTED)
```
awcrosby, line 158: Docs mentioned a `204` status here, but in other cases pulp returns a `202` for tasks since the task won't be completed when the response is sent.

galaxy_ng/app/tasks/deletion.py
```diff
@@ -0,0 +1,36 @@
+import logging
+
+from pulpcore.app.tasks import orphan_cleanup
```
awcrosby, line 3: this will need to be moved to pulpcore.plugin for CI to pass

rochacbruno: @awcrosby you can add here https://github.com/ansible/galaxy_ng/blob/master/template_config.yml#L28 and regenerate ci template

awcrosby: oh great, i'll do that

rochacbruno: anyway I think it is a good idea to add it to pulpcore plugin

galaxy_ng/app/api/v3/viewsets/collection.py
```diff
@@ -87,21 +92,71 @@ class CollectionViewSet(api_base.LocalSettingsMixin,
     serializer_class = CollectionSerializer
 
 
-class UnpaginatedCollectionVersionViewSet(api_base.LocalSettingsMixin,
-                                          ViewNamespaceSerializerContextMixin,
-                                          pulp_ansible_views.UnpaginatedCollectionVersionViewSet):
+class UnpaginatedCollectionVersionViewSet(
+    api_base.LocalSettingsMixin,
+    ViewNamespaceSerializerContextMixin,
+    DestroyModelMixin,
```
rochacbruno, line 98: this Mixin is in the wrong class, it would go to the Collection Version Viewset, but looks like is not needed.

rochacbruno (changes requested):

galaxy_ng/app/api/v3/viewsets/collection.py
```diff
@@ -87,21 +91,72 @@ class CollectionViewSet(api_base.LocalSettingsMixin,
...
+            return Response(
+                {
+                    "detail": _(
+                        "Collection {namespace}.{name} could not be deleted "
+                        "because there are other collections that require it."
+                    ).format(
+                        namespace=collection_version.namespace,
+                        name=collection_version.collection.name,
+                    ),
+                    "dependent_collection_versions": [f"{dep.namespace}.{dep.name} {dep.version}" for dep in dependents],
```
awcrosby, line 147: changed from `dependent_collections` to `dependent_collection_versions` since we are returning collection_versions in `get_dependents()` and this helps users track down which versions depend on the collection_version they are trying to delete

galaxy_ng/app/api/v3/viewsets/collection.py
```diff
@@ -87,21 +85,75 @@ class CollectionViewSet(api_base.LocalSettingsMixin,
...
+                    ],
+                },
+                status=400,
+            )
+
+        dispatch(
+            delete_collection_version,
+            exclusive_resources=collection_version.repositories.all(),
+            kwargs={"collection_version_pk": collection_version.pk},
+        )
```
rochacbruno, line 153: At the top of the file

```py
from pulpcore.plugin.viewsets import OperationPostponedResponse
```

then

```suggestion
        async_result = dispatch(
            delete_collection_version,
            exclusive_resources=collection_version.repositories.all(),
            kwargs={"collection_version_pk": collection_version.pk},
        )
        return OperationPostponedResponse(async_result, request)
```

This makes the proper format with the task link on the response, I just learned this, I will update the namespace deletion properly, we have to do that on every endpoint where a task is dispatched, so the response is always the link to the task (clickable link on API ui)

galaxy_ng/app/api/v3/viewsets/collection.py
```diff
@@ -87,21 +85,75 @@ class CollectionViewSet(api_base.LocalSettingsMixin,
...
+
+
 class CollectionVersionViewSet(api_base.LocalSettingsMixin,
                                ViewNamespaceSerializerContextMixin,
                                pulp_ansible_views.CollectionVersionViewSet):
     serializer_class = CollectionVersionSerializer
     permission_classes = [access_policy.CollectionAccessPolicy]
     list_serializer_class = CollectionVersionListSerializer
 
+    def destroy(self, request: Request, *args, **kwargs) -> Response:
```
rochacbruno, line 118:
```py
from pulpcore.plugin.serializers import AsyncOperationResponseSerializer
```

```suggestion
    @extend_schema(
        description="Trigger an asynchronous delete task",
        responses={202: AsyncOperationResponseSerializer},
    )
    def destroy(self, request: Request, *args, **kwargs):
```

rochacbruno (approved):
Only an optional suggestion

PR #937, opened 2021-09-06
Run pulp-ansible functional tests

fao89:
Issue: AAH-873

PR #969, opened 2021-09-15
Add Container Repo and Container Image Deletion

rochacbruno:
Issue: AAH-712
Issue: AAH-713

Added support for `DELETE` request on routes:

```bash
/api/automation-hub/_ui/v1/execution-environments/repositories/<base_path>/
/api/automation-hub/_ui/v1/execution-environments/repositories/<base_path>/_content/images/<str:manifest_ref>/
```

**plus**

Added a `make api/push-test-images` to feed system with some images for manual testing,

It is not possible **yet** to automate container testing as it is pending on #937

So this PR relies on manual testing only.

### DIFF
--- a/galaxy_ng/app/tasks/deletion.py
+++ b/galaxy_ng/app/tasks/deletion.py
@@ -1,7 +1,8 @@
 import logging
 
 from pulp_ansible.app.models import Collection, CollectionVersion
-from pulpcore.app.tasks import orphan_cleanup
+from pulp_container.app import tasks as pulp_container_tasks
+from pulpcore.app.tasks import orphan_cleanup, reclaim_space
 from pulpcore.plugin.tasking import add_and_remove
 
 log = logging.getLogger(__name__)
@@ -60,3 +61,26 @@ def delete_collection(collection_pk):
 
     log.info("Deleting collection {}".format(collection))
     collection.delete()
+
+
+def delete_container_distribution(instance_ids):
+    """Deletes a container distribution and push repository related."""
+
+    log.info("Running container.general_multi_delete to delete distro and repo")
+    pulp_container_tasks.general_multi_delete(instance_ids=instance_ids)
+
+    log.info("Running orphan_cleanup to delete Container objects and artifacts")
+    orphan_cleanup(content_pks=None, orphan_protection_time=0)
+
+
+def delete_container_image_manifest(repository_pk, content_unit_pks):
+    """Deletes a container image manifest."""
+
+    log.info(f"Running delete manifest for {repository_pk}")
+    pulp_container_tasks.recursive_remove_content(
+        repository_pk=repository_pk,
+        content_units=content_unit_pks,
+    )
+
+    log.info(f"Reclaiming disk space for {repository_pk}")
+    reclaim_space(repo_pks=[repository_pk], force=True)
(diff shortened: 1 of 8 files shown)


galaxy_ng/app/viewsets.py
```diff
@@ -0,0 +1,15 @@
+from pulpcore.plugin import viewsets as pulp_viewsets
+from rest_framework import mixins
+
+from galaxy_ng.app import models
+from galaxy_ng.app.api.ui import serializers
+
+
+class ContainerDistributionViewSet(
```
rochacbruno, line 8: I had to create this Viewset here, even not being used is necessary so Pulp can dynamically import it during the removal tasks, it needs it to generate the URL.

template_config.yml
```diff
@@ -29,6 +29,7 @@ core_import_allowed:
 - pulpcore.app.*viewsets
 - pulpcore\.app.*admin
 - ProgressReportSerializer
+- pulpcore.app.*tasks
```
rochacbruno, line 32: This change is needed until we have the issue https://pulp.plan.io/issues/9418 is resolved

galaxy_ng/app/tasks/deletion.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+def delete_container_distribution(instance_ids):
+    """Deletes a container distribution and push repository related."""
+
+    log.info("Running container.general_multi_delete to delete distro and repo")
+    pulp_container_tasks.general_multi_delete(instance_ids=instance_ids)
+
+    log.info("Running orphan_cleanup to delete CollectionVersion objects and artifacts")
+    orphan_cleanup(content_pks=None, orphan_protection_time=0)
```
rochacbruno, line 16: I was unsure if I needed to call `reclaim_space` here after the call to `orphan_cleanup`

rochacbruno: c/c @ipanova

galaxy_ng/app/tasks/deletion.py
```diff
@@ -0,0 +1,29 @@
...
+    """Deletes a container image manifest."""
+
+    log.info(f"Running delete manifest for {repository_pk}")
+    pulp_container_tasks.recursive_remove_content(
+        repository_pk=repository_pk,
+        content_units=content_unit_pks,
+    )
+
+    log.info(f"Reclaiming disk space for {repository_pk}")
+    reclaim_space(repo_pks=[repository_pk])
```
ipanova, line 29: you need to set the force flag to True https://github.com/pulp/pulpcore/blob/master/pulpcore/app/tasks/reclaim_space.py#L23

galaxy_ng/app/tasks/deletion.py
ipanova: There is no need to call the `reclaim_space`.  The space can be reclaimed on a repo, since you've just removed it there is nothing to reclaim on..
Orphan clean up will clean up everything what does not belong to any of the repo, including content + artifacts.

ipanova: re: orphan_protection_time, willingly you can set it globally https://github.com/pulp/pulpcore/blob/master/pulpcore/app/settings.py#L248 https://github.com/pulp/pulpcore/blob/master/docs/configuration/settings.rst#orphan_protection_time

galaxy_ng/app/tasks/deletion.py
rochacbruno: Thanks @ipanova 💯

newswangerd (approved):

PR #973, opened 2021-09-16
update_manifest workflow: fix & use deploy_key instead of user token

himdel:
The main change should match ansible/ansible-hub-ui#946 .. switch from user&token to a deploy-key encrypted with a passphrase - the gpg-encrypted key is added to the repo, encrypted by a passphrase added to secrets.

It also fixes a couple of issues causing the task to never run...

GITHUB_PULL_REQUEST is set to a PR number during the pull_request event, and nothing during the push event,
the script expects `false`
fixing by testing GITHUB_PULL_REQUEST for nonempty

GITHUB_BRANCH is set to github.head_ref, only available during pull_request events,
the script tests for "master" or "stable"
fixing by updating GITHUB_BRANCH to use github.ref sans the refs/heads/ prefix

---

Cc @newswangerd , @rochacbruno


.github/workflows/ci.yml
```diff
@@ -197,18 +197,16 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v2
         with:
-          ref: ${{ github.event.after }}  # for PR avoids checking out merge commit
           fetch-depth: 0  # include all history
 
+      - name: "Set GITHUB_BRANCH"
+        run: |
```
rochacbruno, line 203: we should not change this file, if we need changes we need to add to template_config.ymlz there is a ci_env there to add additional environment variables.

If something needs fix then we have to fix on https://github.com/pulp/plugin_template

we need to use plugin_template to generate some of the CI files

rochacbruno: update manifest is on .github/post_job_templatwz you can edit that file and then run plugin_template

rochacbruno: https://github.com/ansible/galaxy_ng/blob/master/.github/post-job-template.yml.j2

himdel: Aah, thanks!

Updated that file, `./plugin-template --github galaxy_ng` no longer complains (the j2 syntax clash with github's `${{` is unfortunate, and leads to a rather cryptic `jinja2.exceptions.UndefinedError: 'github' is undefined`), and `ci.yml` stays the same :+1:

rochacbruno (approved):

rochacbruno:
@himdel one of your commits are not signed

PR #974, opened 2021-09-16
Fix collection version filter chaining with dependencies

awcrosby:
No-Issue


awcrosby:
@jerabekjiri i think this should fix dep chaining here, please check if this works for you.

jerabekjiri:
@awcrosby thanks, the filtering by name works now :+1:, but what if I need to target a specific collection version? right now if I include a version it narrows down the result or maybe I'm missing something?

awcrosby:
@jerabekjiri i recommend for any additional requirements / use cases to add them to the issue https://issues.redhat.com/browse/AAH-902 - right now the filter takes a Collection not a CollectionVersion

jerabekjiri (approved):

PR #975, opened 2021-09-20
Fix download url for V3 unpaginated collection versions endpoint

gerrod3:
No-Issue

I filed an issue in redmine, before figuring out this is a galaxy_ng problem: https://pulp.plan.io/issues/9405. Do you need me to create and link this to a jira issue? 

Also, are there tests for the V3 unpaginated endpoints? I ran into this issue while trying to implement synclist on the AH staging environment.

### DIFF
--- a/galaxy_ng/app/api/v3/serializers/collection.py
+++ b/galaxy_ng/app/api/v3/serializers/collection.py
@@ -22,6 +22,15 @@ log = logging.getLogger(__name__)
 
 
 class HrefNamespaceMixin:
+
+    def _get_download_url(self, obj):
+        """Get artifact download URL."""
+
+        host = settings.ANSIBLE_API_HOSTNAME.strip("/")
+        prefix = settings.GALAXY_API_PATH_PREFIX.strip("/")
+        distro_base_path = self.context["path"]
+        return f"{host}/{prefix}/v3/artifacts/collections/{distro_base_path}/{obj.relative_path}"
+
     def _get_href(self, url_name, **kwargs):
         """Generic get_*_href that uses context["view_namespace"] to reverse the right url"""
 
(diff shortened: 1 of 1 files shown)


rochacbruno:
Thanks @gerrod3 

We require all commits to be signed.

galaxy_ng/app/api/v3/serializers/collection.py
```diff
@@ -90,6 +90,16 @@ class UnpaginatedCollectionVersionSerializer(_UnpaginatedCollectionVersionSerial
...
 
+    def get_download_url(self, obj):
+        """
+        Get artifact download URL.
+        """
+
+        host = settings.ANSIBLE_API_HOSTNAME.strip("/")
+        prefix = settings.GALAXY_API_PATH_PREFIX.strip("/")
+        distro_base_path = self.context["path"]
+        return f"{host}/{prefix}/v3/artifacts/collections/{distro_base_path}/{obj.relative_path}"
```
rochacbruno, line 101: @gerrod3 what if we move this whole method https://github.com/ansible/galaxy_ng/blob/d1883acc636e7fa304211ca759e7aa3feeed4547/galaxy_ng/app/api/v3/serializers/collection.py#L104-L112

To the HrefNamespaceMixin https://github.com/ansible/galaxy_ng/blob/d1883acc636e7fa304211ca759e7aa3feeed4547/galaxy_ng/app/api/v3/serializers/collection.py#L24

?

rochacbruno:
@gerrod3  The base branch requires all commits to be signed. [Learn more about signing commits.](https://docs.github.com/articles/about-gpg/)

rochacbruno:
Also notice that this is about to change on https://github.com/ansible/galaxy_ng/pull/977

gerrod3:
> Also notice that this is about to change on #977

I think this PR should still be merged and have that one be updated, so that there is one source of truth for the download_url across the different V3 endpoints.

PR #976, opened 2021-09-20
Add content guard

bmclaughlin:
- [x] Add content guard to check authentication before a collection can be downloaded.
- [x] Create a migration to add content guard to existing Distributions
- [x] Create a signal to add content guard to new Distributions

Issue: AAH-923


galaxy_ng/app/migrations/0025_add_content_guard_to_distributions.py
```diff
@@ -0,0 +1,34 @@
+from django.db import migrations
+
+def add_content_guard(apps, schema_editor):
+    """
+    Enforce that all distributions have a content guard applied
+    """
+    AnsibleDistribution = apps.get_model('ansible', 'AnsibleDistribution')
+    ContentGuard = apps.get_model('galaxy', 'CollectionDownloadContentGuard')
+
+    cg = ContentGuard(
```
newswangerd, line 10: This should use `get_or_create` to initialize the object if it doesn't exist.

galaxy_ng/app/models/contentguard.py
```diff
@@ -0,0 +1,25 @@
...
+
+    def permit(self, request):
+        """
+        Authorize the specified request based on if the request is authenticated.
+        """
+        from galaxy_ng.app.api.v3.viewsets import CollectionArtifactDownloadView
+
+        view = CollectionArtifactDownloadView()
+        setattr(view, "get_object", lambda: self)
+        setattr(view, "action", "download")
```
newswangerd, line 18: I believe the action on this view is already set to "download"

galaxy_ng/app/models/contentguard.py
```diff
@@ -0,0 +1,25 @@
...
+    """
+
+    def permit(self, request):
+        """
+        Authorize the specified request based on if the request is authenticated.
+        """
+        from galaxy_ng.app.api.v3.viewsets import CollectionArtifactDownloadView
+
+        view = CollectionArtifactDownloadView()
+        setattr(view, "get_object", lambda: self)
```
newswangerd, line 17: This is overriding the `get_object` function so that it always returns the content guard instance. What we need to do is make it so that `get_object` returns the collection that is being accessed.

However, looking through the access policy, it looks like the download action doesn't take any conditions, so we can probably get away without defining a `get_object` method on the viewset until we get to a point where we need to add conditions to the download function.
```
        {
            "action": ["download"],
            "principal": 'authenticated',
            "effect": "allow",
        },
```

newswangerd (changes requested):

galaxy_ng/app/migrations/0025_add_content_guard_to_distributions.py
```diff
@@ -0,0 +1,34 @@
...
+        pulp_type='ansible.ansible'
+    )
+    cg.save()
+
+    AnsibleDistribution.objects.filter(
+        content_guard=None,
+        pulp_type='ansible.ansible'
+    ).update(
+        content_guard=ContentGuard.objects.first(),
```
rochacbruno, line 19: I suggest it to be more specific, use the `cg` object created above instead of getting the first content guard

galaxy_ng/app/migrations/0025_add_content_guard_to_distributions.py
```diff
@@ -0,0 +1,31 @@
...
+    AnsibleDistribution = apps.get_model('ansible', 'AnsibleDistribution')
+    ContentGuard = apps.get_model('galaxy', 'CollectionDownloadContentGuard')
+
+    AnsibleDistribution.objects.filter(
+        content_guard=None,
+        pulp_type='ansible.ansible'
+    ).update(
+        content_guard=ContentGuard.objects.get_or_create(
+            pulp_type='ansible.ansible'
+        )[0]
```
chouseknecht, line 16: Is '[0]' required here? I may be completely wrong, but I think it implies that "ContentGuard.objects.get_or_create()" returns a queryset when in fact "objects.get()" and "objects.get_or_create()" return a single object.

galaxy_ng/app/signals/handlers.py
```diff
@@ -22,6 +22,17 @@ def ensure_retain_repo_versions_on_repository(sender, instance, created, **kwarg
...
 
 
+@receiver(post_save, sender=AnsibleDistribution)
+def ensure_content_guard_exists_on_distribution(sender, instance, created, **kwargs):
+    """Ensure distribution have a content guard when created."""
+
+    if created and instance.content_guard is None:
+        instance.content_guard = CollectionDownloadContentGuard.objects.get_or_create(
+            pulp_type='ansible.ansible'
+        )[0]
```
chouseknecht, line 32: Same question as above. Do we need the "[0]" ?

galaxy_ng/app/migrations/0025_add_content_guard_to_distributions.py
rochacbruno: `get_or_create` returns a tuple @chouseknecht 

commonly we use like:
```py
content_guard, created = ContentGuard.objects.get_or_create(...)
```
content_guard: ContentGuard
created: bool # indicate whether the object was get or created 

It is not wrong to access the `[0]` index but to be more clear I recommend

```py
content_guard, _ = ContentGuard.objects.get_or_create(pulp_type='ansible.ansible')
...
.update(content_guard=content_guard)
```

bmclaughlin: Much clearer, I was not aware of that pattern.  Update coming.

bmclaughlin:
@chouseknecht, @rochacbruno updated migration and signal.

rochacbruno:
ci erroring on 
```
django.core.exceptions.ImproperlyConfigured: Field name `content-guard` is not valid for model `AnsibleDistribution`.
```

galaxy_ng/app/models/contentguard.py
```diff
@@ -11,11 +11,13 @@ def permit(self, request):
         """
         Authorize the specified request based on if the request is authenticated.
         """
+        if not (drequest := request.get("drf_request", None)):
```
rochacbruno, line 14: The Walrus `:=` for the first time on our codebase :) 💯

galaxy_ng/app/settings.py
```diff
@@ -65,6 +65,16 @@
 # set to 'insights' for cloud.redhat.com deployments
 GALAXY_DEPLOYMENT_MODE = 'standalone'
 
+REST_FRAMEWORK = {
+    'DEFAULT_AUTHENTICATION_CLASSES': (
+        "rest_framework.authentication.SessionAuthentication",
+        "rest_framework.authentication.TokenAuthentication",
+        "rest_framework.authentication.BasicAuthentication",
+        "dynaconf_merge",
```
rochacbruno, line 73: I think this `dynaconf_merge` inside the list is not needed, you already have `dynaconf_merge: True` at the root of the dictionary.

I just tested


```py
>>> from dynaconf import Dynaconf
>>> settings = Dynaconf(REST_FRAMEWORK={"DEFAULT_AUTHENTICATION_CLASSES": ("a", "b", "c")})
>>> settings.REST_FRAMEWORK
<Box: {'DEFAULT_AUTHENTICATION_CLASSES': ['a', 'b', 'c']}>

>>> settings.set("REST_FRAMEWORK", {"DEFAULT_AUTHENTICATION_CLASSES": ("d","e"), "dynaconf_merge": True})
>>> settings.REST_FRAMEWORK
<Box: {'DEFAULT_AUTHENTICATION_CLASSES': ['a', 'b', 'c', 'd', 'e']}>
```

rochacbruno:
@bmclaughlin can you provide a testing workflow for this PR? here in comments, something like

1. Login to hub
2. Try to click on x, y z.
3. Verify that something happened

I want to help testing, but I checked out to this PR and I don't know what to check during my manual testing.

bmclaughlin:
@rochacbruno The manual testing that I've done for this PR started as:
1. Make sure new migrations have been run
2. Verify our core distros in `core_distribution` have the same `content_guard_id` (empty prior to this PR)
3. Upload (or `make api/create-test-collections`), verify `content_guard_id` is populated again, same `pulp_id`

That is working as expected, I attempted to download an uploaded collection which now results in a 403 from the content-app.  The content app is being re-enabled in #977, so I assume downloads should continue to work as they did prior to this PR.

PR #977, opened 2021-09-20
Download collection artifacts from the pulp content app

bmclaughlin:
This reverts #781.  This relies on the content guards from #976.
Issue: AAH-924


bmclaughlin:
Moved to #981

PR #981, opened 2021-09-23
Add pre-authorized redirect content guard

bmclaughlin:
- Added content guard
- Includes commit to re-enable content-app

Issue: AAH-923
Issue: AAH-924


galaxy_ng/app/migrations/0025_add_content_guard_to_distributions.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def add_content_guard(apps, schema_editor):
+    """
+    Enforce that all distributions have a content guard applied
+    """
+    AnsibleDistribution = apps.get_model('ansible', 'AnsibleDistribution')
+    ContentRedirectContentGuard = apps.get_model('galaxy', 'ContentRedirectContentGuard')
+
+    content_guard, _ = ContentRedirectContentGuard.objects.get_or_create(pulp_type='ansible.ansible')
```
newswangerd, line 11: I'd give this a name. It looks like if you leave out the name it defaults to "", which is no good.

galaxy_ng/app/signals/handlers.py
```diff
@@ -22,6 +22,19 @@ def ensure_retain_repo_versions_on_repository(sender, instance, created, **kwarg
         instance.save()
 
 
+@receiver(post_save, sender=AnsibleDistribution)
+def ensure_content_guard_exists_on_distribution(sender, instance, created, **kwargs):
+    """Ensure distribution have a content guard when created."""
+
+    content_guard, _ = ContentRedirectContentGuard.objects.get_or_create(
```
newswangerd, line 29: I think you can just do `objects.first()` here. We can trust that the migrations created the content guard successfully.

newswangerd (commented):
Looking good so far! I'm excited that this seems to be working

fao89 (commented):
pulp_ansible tests broke because it expected a [NamedModelViewSet](https://github.com/pulp/pulpcore/blob/master/pulpcore/app/viewsets/publication.py#L107-L139) for ContentRedirectContentGuard
```
 Could not determine ViewSet base name for model <class 'galaxy_ng.app.models.contentguard.ContentRedirectContentGuard'>
 ```
Not sure if it is really needed here, in any case, I'm disabling pulp_ansible tests:
https://github.com/ansible/galaxy_ng/pull/984

File: galaxy_ng/app/api/v3/viewsets/collection.py [hunk 3]
```diff
@@ -380,7 +383,8 @@ def get(self, request, *args, **kwargs):
         if content_bind and content_bind.startswith("unix:"):
             response = self._get_unix_socket_response(url)
         else:
-            response = self._get_tcp_response(url)
+            distribution = self._get_ansible_distribution(self.kwargs['path'])
```
newswangerd, line 386: Right now we’re proxying the requests for the content app for on prem on cloud. We still have to proxy the requests to `X_PULP_CONTENT_HOST` because console.redhat.com doesn’t expose the content app, but we need to stop proxying the requests when running on premise.

I think the best way to fix this is to check if the app is running in insights mode. If it is, proxy the request to `X_PULP_CONTENT_HOST` like we do now. If it's not, generate a URL for `{CONTENT_ORIGIN}/{CONTENT_PATH_PREFIX}`, pre authenticate it and return a redirect to it.

You can get rid of `_get_tcp_response` and `_get_unix_socket_response` since those were used for proxying on prem, and aren't needed anymore.

galaxy_ng/app/api/v3/viewsets/collection.py
```diff
@@ -365,22 +365,56 @@ def _get_unix_socket_response(self, url):
...
+                    content_type=response.headers['Content-Type']
+                )
+            metrics.collection_artifact_download_failures.labels(status=response.status_code).inc()
+            raise APIException(
+                _('Unexpected response from content app. Code: %s.') % response.status_code
+            )
+        elif settings.GALAXY_DEPLOYMENT_MODE == DeploymentMode.STANDALONE.value:
+            url = '{host}/{prefix}/{distro_base_path}/{filename}'.format(
+                host=settings.CONTENT_ORIGIN.strip("/"),
+                prefix=prefix,
```
newswangerd, line 410: might need to strip("/") prefix

newswangerd: nevermind just noticed line 376

newswangerd (approved):
https://media.giphy.com/media/kyLYXonQYYfwYDIeZl/giphy.gif

PR #983, opened 2021-09-24
Enable indexing EEs from Red Hat Registry

newswangerd:
Add `_ui/v1/execution-environments/registries/{registry_pk}/index/` endpoint.

```
curl \
  -X POST \
  -H 'Content-Type: application/json' \
  curl_flags  -u admin:admin \
  http://localhost:5001/api/automation-hub/_ui/v1/execution-environments/registries/{registry_pk}/index/
```

Launches a task that:
- Checks if the registry is indexable (for now, that just means that it's a red hat registry instance)
- Launches a task that reaches out to catalog.redhat.com and pulls a list of execution environments
- For each EE in the response, launch a task that creates a new remote repository in the registry.

### DIFF
--- a/galaxy_ng/app/api/ui/serializers/execution_environment.py
+++ b/galaxy_ng/app/api/ui/serializers/execution_environment.py
@@ -254,6 +254,7 @@ class ContainerRegistryRemoteSerializer(
     updated_at = serializers.DateTimeField(source='pulp_last_updated', required=False)
     last_sync_task = utils.RemoteSyncTaskField(source="*")
     write_only_fields = serializers.SerializerMethodField()
+    is_indexable = serializers.SerializerMethodField()
 
     class Meta:
         model = models.ContainerRegistryRemote
@@ -277,6 +278,7 @@ class ContainerRegistryRemoteSerializer(
             "proxy_password",
             "write_only_fields",
             "rate_limit",
+            "is_indexable"
         ]
         extra_kwargs = {
             'name': {'read_only': True},
@@ -285,6 +287,11 @@ class ContainerRegistryRemoteSerializer(
 
     def get_write_only_fields(self, obj):
         return utils.get_write_only_fields(self, obj)
+
+    def get_is_indexable(self, obj):
+        if obj.get_registry_backend():
+            return True
+        return False
 
 
 class ContainerRemoteSerializer(
--- a/galaxy_ng/app/api/ui/views/index_execution_environments.py
+++ b/galaxy_ng/app/api/ui/views/index_execution_environments.py
@@ -1,0 +1,49 @@
+from rest_framework.request import Request
+from rest_framework.response import Response
+from rest_framework.exceptions import ValidationError
+
+from django.shortcuts import get_object_or_404
+from django.utils.translation import gettext_lazy as _
+
+from pulpcore.plugin.tasking import dispatch
+from pulpcore.plugin.viewsets import OperationPostponedResponse
+
+from galaxy_ng.app.api import base as api_base
+from galaxy_ng.app.access_control import access_policy
+from galaxy_ng.app import models
+from galaxy_ng.app import tasks
+
+
+class IndexRegistryEEView(api_base.APIView):
+    permission_classes = [access_policy.ContainerRegistryRemoteAccessPolicy]
+    action = 'index_execution_environments'
+
+    def post(self, request: Request, *args, **kwargs) -> Response:
+        registry = get_object_or_404(models.ContainerRegistryRemote, pk=kwargs['pk'])
+
+        if not registry.get_registry_backend():
+            raise ValidationError(
+                detail={
+                    'url':
+                        _('Indexing execution environments is not supported on this registry.')
+                })
+
+        serializable_meta = {}
+        for key in self.request.META:
+            val = self.request.META[key]
+            if type(val) == str:
+                serializable_meta[key] = val
+
+        request_data = {
+            "META": serializable_meta,
+        }
+
+        result = dispatch(
+            tasks.index_execution_environments_from_redhat_registry,
+            kwargs={
+                "registry_pk": registry.pk,
+                "request_data": request_data
+            },
+        )
+
+        return OperationPostponedResponse(result, self.request)
--- a/galaxy_ng/app/api/ui/views/index_execution_environments.py
+++ b/galaxy_ng/app/api/ui/views/index_execution_environments.py
@@ -1,12 +1,16 @@
 from rest_framework.request import Request
 from rest_framework.response import Response
 from rest_framework.exceptions import ValidationError
+from rest_framework import status
+
+from drf_spectacular.utils import extend_schema
 
 from django.shortcuts import get_object_or_404
 from django.utils.translation import gettext_lazy as _
 
 from pulpcore.plugin.tasking import dispatch
 from pulpcore.plugin.viewsets import OperationPostponedResponse
+from pulpcore.plugin.serializers import AsyncOperationResponseSerializer
 
 from galaxy_ng.app.api import base as api_base
 from galaxy_ng.app.access_control import access_policy
@@ -18,6 +22,10 @@ class IndexRegistryEEView(api_base.APIView):
     permission_classes = [access_policy.ContainerRegistryRemoteAccessPolicy]
     action = 'index_execution_environments'
 
+    @extend_schema(
+        description="Trigger an asynchronous delete task",
+        responses={status.HTTP_202_ACCEPTED: AsyncOperationResponseSerializer},
+    )
     def post(self, request: Request, *args, **kwargs) -> Response:
         registry = get_object_or_404(models.ContainerRegistryRemote, pk=kwargs['pk'])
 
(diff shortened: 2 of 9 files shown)


galaxy_ng/app/api/ui/serializers/execution_environment.py
```diff
@@ -284,6 +286,11 @@ class Meta:
     def get_write_only_fields(self, obj):
         return utils.get_write_only_fields(self, obj)
 
+    def get_is_indexable(self, obj):
+        if obj.get_registry_backend():
+            return True
+        return False
```
rochacbruno, line 292:
```suggestion
        return bool(obj.get_registry_backend())
```

has the same effect

galaxy_ng/app/api/ui/serializers/execution_environment.py
```diff
@@ -252,6 +252,7 @@ class ContainerRegistryRemoteSerializer(
     updated_at = serializers.DateTimeField(source='pulp_last_updated', required=False)
     last_sync_task = utils.RemoteSyncTaskField(source="*")
     write_only_fields = serializers.SerializerMethodField()
+    is_indexable = serializers.SerializerMethodField()
```
rochacbruno, line 255: I wonder if it would not be possible to use 
```suggestion
    is_indexable = serializers.BooleanField(source="get_registry_backend")
```


looks like source can be a callable https://www.django-rest-framework.org/api-guide/fields/#source

Will that work?

galaxy_ng/app/tasks/index_registry.py
```diff
@@ -0,0 +1,160 @@
...
+log = logging.getLogger(__name__)
+
+
+CATALOG_API = "https://catalog.redhat.com/api/containers/v1/repositories"
+
+
+class CouldNotCreateContainerError(Exception):
+    def __init__(self, remote_name, error=""):
+        self.message = _(
+            f"Failed to create container {remote_name}. {error}"
```
rochacbruno, line 26: IIRC we should not use `f".."` inside a `_()` translation.
```suggestion
            _("Failed to create container {remote_name}. {error}").format(remote_name=remote_name, error=error)
```

@jerabekjiri can confirm

galaxy_ng/app/tasks/index_registry.py
```diff
@@ -0,0 +1,160 @@
...
+            "upstream_name": container['repository'],
+            "description": container['display_data']['short_description'],
+            "readme": container['display_data']['long_description_markdown'],
+        })
+
+    return containers
+
+
+def _update_distro_readme_and_description(distro, container_data):
+    readme = models.ContainerDistroReadme.objects.get_or_create(container=distro)[0]
```
rochacbruno, line 56:
```suggestion
    readme, _ = models.ContainerDistroReadme.objects.get_or_create(container=distro)
```

galaxy_ng/app/api/ui/views/index_execution_environments.py
```diff
@@ -0,0 +1,49 @@
...
+
+        result = dispatch(
+            tasks.index_execution_environments_from_redhat_registry,
+            kwargs={
+                "registry_pk": registry.pk,
+                "request_data": request_data
+            },
+        )
+
+        return OperationPostponedResponse(result, self.request)
```
rochacbruno, line 49: When returning a `OperationPostponedResponse` is a good practice to extend_schema and provide the proper serializer, this makes OpenAPISpec better.

Example
https://github.com/ansible/galaxy_ng/blob/453d4ecdbecbe9f3098ed16efbab0848f97965bd/galaxy_ng/app/api/v3/viewsets/collection.py#L102-L105

galaxy_ng/app/models/container.py
```diff
@@ -67,6 +71,12 @@ def get_connection_fields(self):
 
         return result
 
+    def get_registry_backend(self):
+        for registry in self.RED_HAT_REGISTY_DOMAINS:
+            if self.url in registry:
+                return 'redhat'
+        return None
```
rochacbruno, line 78: If the idea is to match 

self.url ({https, http}://registry.redhat.io) with (registry.redhat.io) this should be compared in the opposite way,

```suggestion
        for registry in self.RED_HAT_REGISTY_DOMAINS:
            if registry in self.url:
                return 'redhat'
        return None
```

galaxy_ng/app/tasks/index_registry.py
```diff
@@ -0,0 +1,160 @@
...
+from pulp_container.app import models as container_models
+
+from galaxy_ng.app.api.ui import serializers
+from galaxy_ng.app import models
+
+
+log = logging.getLogger(__name__)
+
+
+CATALOG_API = "https://catalog.redhat.com/api/containers/v1/repositories"
```
rochacbruno, line 20: should this go to settings?
```suggestion
CATALOG_API = settings.get("CATALOG_API",  "https://catalog.redhat.com/api/containers/v1/repositories")
```

galaxy_ng/app/tasks/index_registry.py
```diff
@@ -0,0 +1,160 @@
...
+        )
+
+        super().__init__(self.message)
+
+
+def _get_request(request_data):
+    request = HttpRequest()
+
+    for k in request_data:
+        setattr(request, k, request_data[k])
```
rochacbruno, line 36:
```suggestion
    for k, v in request_data.items():
        setattr(request, k, v)
```

rochacbruno (approved):
Small nitpick suggestions

galaxy_ng/app/tasks/index_registry.py
awcrosby: > IIRC we should not use f".." inside a _() translation

that is correct

galaxy_ng/app/api/ui/serializers/execution_environment.py
newswangerd: I kinda like the more explicit `if... then` it's easier to read.

galaxy_ng/app/tasks/index_registry.py
```diff
@@ -0,0 +1,160 @@
...
+        # create a subtask for each remote, so that if one fails, we can throw a usable error
+        # message for the user to look at and prevent the rest of the repositories from failing.
+        dispatch(
+            create_or_update_remote_container,
+            kwargs={
+                "container_data": remote,
+                "registry_pk": registry.pk,
+                "request_data": request_data
+            },
+            exclusive_resources=['/pulp/api/v3/distributions/']
```
awcrosby, line 159: for `exclusive_resources` (or `resources` in the past), iirc we send a list of model objects or list of string pk of a model object https://docs.pulpproject.org/pulpcore/plugins/api-reference/tasking.html#pulpcore.plugin.tasking.dispatch

Is an endpoint considered a resource that it looks up to find objects? and would it lock every distribution in the database? If you need to lock all distros assocated to containers, maybe you could send it something like `models.ContainerDistribution.objects.all()`

galaxy_ng/app/tasks/index_registry.py
newswangerd: Won't this overwrite `from django.utils.translation import gettext_lazy as _`?

awcrosby (changes requested):
I added a question. Also the suggestions by @rochacbruno seem pretty good.

galaxy_ng/app/tasks/index_registry.py
newswangerd: I'm torn on this. I think I'm going to keep it out of settings for now. I can't think of a good reason why anyone would need to edit this value

galaxy_ng/app/tasks/index_registry.py
newswangerd: I think I need to pass a distribution instance here. Fabricio had a note about this in IRC.

galaxy_ng/app/tasks/index_registry.py
newswangerd: neat!

galaxy_ng/app/tasks/index_registry.py
newswangerd: when you create distributions you need to put a lot on all distributions so that they can validate that the base path doesn't overlap

galaxy_ng/app/tasks/index_registry.py
rochacbruno: You are right, so

`readme, _created = ...`

galaxy_ng/app/tasks/index_registry.py
awcrosby: > I think I need to pass a distribution instance here

I assumed this was an instance of type models.*Distribution rather than a url, I did a quick search in pulpcore and pulp_ansible for resources|exclusive_resources|shared_resources and they seem to all pass database objects. But the url may just be a pattern I don't know.

I _thought_ the only reason it allowed a string to be passed in, was in case it was a pk

awcrosby: this is correct as-is per pulp folks!

awcrosby (approved):

PR #986, opened 2021-09-24
Fix the need to redeclare viewsets using PEP0562

rochacbruno:
No-Issue

### DIFF
--- a/galaxy_ng/app/viewsets.py
+++ b/galaxy_ng/app/viewsets.py
@@ -22,7 +22,7 @@ def __getattr__(self, name):
     <plugin_name>.app.viewsets. Since galaxy_ng stores it's viewsets in a different
     module, this file is necesary for pulp tbe able to associate the ContainerRegistryRemote
     model to a viewset.
-    Without this viewset defined here, pulp get's confused when it tries to auto generate
+    Without this viewset defined here, pulp gets confused when it tries to auto generate
     the form on the repositories endpoint because it tries to provide a dropdown
     on the remote field and can't find a viewset name for galaxy's ContainerRegistryRemote model.
 
--- a/galaxy_ng/app/viewsets.py
+++ b/galaxy_ng/app/viewsets.py
@@ -19,7 +19,7 @@ def __getattr__(self, name):
     Pulp expects models it manages to come with a viewset attached to them. Since the
     ContainerRegistryRemote model is a pulp remote, pulp expects it to have a NamedModelViewSet
     viewset attached to it. Pulp associates viewsets with models by looking in
-    <plugin_name>.app.viewsets. Since galaxy_ng stores it's viewsets in a different
+    <plugin_name>.app.viewsets. Since galaxy_ng stores its viewsets in a different
     module, this file is necesary for pulp tbe able to associate the ContainerRegistryRemote
     model to a viewset.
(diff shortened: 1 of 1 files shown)


galaxy_ng/app/viewsets.py
```diff
@@ -1,35 +1,40 @@
...
+
+    This file is necessary to prevent the DRF web API browser from breaking on all of the
+    pulp/api/v3/repositories/ endpoints.
+    Pulp expects models it manages to come with a viewset attached to them. Since the
+    ContainerRegistryRemote model is a pulp remote, pulp expects it to have a pulp
+    viewset attached to it. Pulp associates viewsets with models by looking in
+    <plugin_name>.app.viewsets. Since galaxy_ng stores it's viewsets in a different
+    module, this file is necesary for pulp tbe able to associate the ContainerRegistryRemote
+    model to a viewset.
+    Without this viewset defined here, pulp get's confused when it tries to auto generate
```
daviddavis, line 25:
```suggestion
    Without this viewset defined here, pulp gets confused when it tries to auto generate
```

galaxy_ng/app/viewsets.py
```diff
@@ -1,35 +1,40 @@
...
+    the form on the repositories endpoint because it tries to provide a dropdown
+    on the remote field and can't find a viewset name for galaxy's ContainerRegistryRemote model.
+
+    """
+
+    if name in viewsets:
+        try:
+            module = __import__(viewsets[name])
+        except ImportError as e:
+            log.error("Error importing viewset: %s" % e)
```
daviddavis, line 35: Maybe worth being consistent with line 39?

galaxy_ng/app/viewsets.py
```diff
@@ -1,35 +1,40 @@
...
+
+
+def __getattr__(self, name):
+    """
+    This is a hack to get pulp to associate the model with a viewset.
+
+    This file is necessary to prevent the DRF web API browser from breaking on all of the
+    pulp/api/v3/repositories/ endpoints.
+    Pulp expects models it manages to come with a viewset attached to them. Since the
+    ContainerRegistryRemote model is a pulp remote, pulp expects it to have a pulp
```
daviddavis, line 20:
```suggestion
    ContainerRegistryRemote model is a pulp remote, pulp expects it to have a NamedModelViewSet
```

galaxy_ng/app/viewsets.py
```diff
@@ -1,35 +1,40 @@
...
+    viewset attached to it. Pulp associates viewsets with models by looking in
+    <plugin_name>.app.viewsets. Since galaxy_ng stores it's viewsets in a different
```
daviddavis, line 22:
```suggestion
    <plugin_name>.app.viewsets. Since galaxy_ng stores its viewsets in a different
```

daviddavis (approved):
LGTM. A couple small suggestions.

galaxy_ng/app/viewsets.py
rochacbruno: In this case the viewset is declared but there was an error importing it, in the other case it is really not found.

daviddavis: Oh I was just wondering about the use of %

newswangerd (approved):

PR #988, opened 2021-09-24
Add api endpoint to sync all remotes in a container registry.

newswangerd:
Issue: AAH-945

### DIFF
--- a/galaxy_ng/app/api/ui/views/sync.py
+++ b/galaxy_ng/app/api/ui/views/sync.py
@@ -6,7 +6,6 @@ from django.shortcuts import get_object_or_404
 from django.utils.translation import gettext_lazy as _
 
 from pulp_container.app import models as pulp_models
-from pulp_container.app.tasks.synchronize import synchronize as container_sync
 from pulpcore.plugin.viewsets import (
     OperationPostponedResponse,
 )
@@ -14,7 +13,7 @@ from pulpcore.plugin.tasking import dispatch
 
 from galaxy_ng.app.api import base as api_base
 from galaxy_ng.app.access_control import access_policy
-from galaxy_ng.app import models
+from galaxy_ng.app import models, tasks
 
 
 class ContainerSyncRemoteView(api_base.APIView):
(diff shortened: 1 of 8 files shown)


himdel (approved):
Works for me, creates a child task per each remote :+1: 

(updated https://github.com/ansible/ansible-hub-ui/pull/898 to use pk instead of name)

galaxy_ng/app/tasks/registry_sync.py
```diff
@@ -0,0 +1,32 @@
...
+
+def launch_container_remote_sync(remote, registry, repository):
+    for key, value in registry.get_connection_fields().items():
+        setattr(remote, key, value)
+        remote.save()
+
+    return dispatch(
+        container_sync,
+        shared_resources=[remote],
+        exclusive_resources=[repository],
```
ipanova, line 16: :+1:

galaxy_ng/app/api/ui/views/sync.py
```diff
@@ -40,18 +39,21 @@ def post(self, request: Request, *args, **kwargs) -> Response:
...
+    action = 'sync'
+
+    def post(self, request: Request, *args, **kwargs) -> Response:
+        registry = get_object_or_404(models.ContainerRegistryRemote, pk=kwargs['pk'])
 
         result = dispatch(
-            container_sync,
-            shared_resources=[remote],
-            exclusive_resources=[distro.repository],
+            tasks.sync_all_repos_in_registry,
```
ipanova, line 54: this would be a great usecase to use task groups, here is an example https://github.com/pulp/pulp_file/pull/536/files#diff-dc946cef2bc4fef289c4a2d72dde9e612e92efc300edb9273a4e01b914092401R212

newswangerd: @ipanova thanks! I'll look into this.

newswangerd: We're not totally sure if our task ui will work with task groups, but I'll give it a shot

ipanova: err-  i forgot this particular part returning task group response from viewset is unreleased but we do have task groups in pulp https://docs.pulpproject.org/pulpcore/restapi.html#tag/Task-Groups

PR #994, opened 2021-09-27
Allow deletion of ContainerRegistryRemote

rochacbruno:
Issue: AAH-931

### DIFF
--- a/CHANGES/931.feature
+++ b/CHANGES/931.feature
@@ -1,0 +1,1 @@
+Allow container registry-remote to be deleted
--- a/galaxy_ng/app/access_control/statements/standalone.py
+++ b/galaxy_ng/app/access_control/statements/standalone.py
@@ -294,8 +294,9 @@ STANDALONE_STATEMENTS = {
         # prevents deletion of registry
         {
             "action": "destroy",
-            "principal": "*",
-            "effect": "deny",
+            "principal": "authenticated",
+            "effect": "allow",
+            "condition": "has_model_perms:galaxy.delete_containerregistryremote"
         },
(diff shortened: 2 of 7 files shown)


galaxy_ng/app/api/ui/viewsets/execution_environment.py
```diff
@@ -346,6 +351,34 @@ class ContainerRegistryRemoteViewSet(api_base.ModelViewSet):
...
+        This will make it so that container remotes that are associate with a deleted registry will
+        still be available and can be re-added to another registry, but they won't be able to
+        be re-synced.
+
+        return: Response(status=204)
+        """
+
+        registry = self.get_object()
+
+        for remote_association in registry.containerregistryrepos_set.all():
```
newswangerd, line 375: This might not be necessary. The model is already set to cascade if the registry is deleted: https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/models/container.py#L84

https://stackoverflow.com/questions/38388423/what-does-on-delete-do-on-django-models

newswangerd: You may not have to create a destroy method at all for this. The one  from the parent viewset might be enough.

newswangerd (changes requested):
This needs to be updated to create the registry/remote connection if one doesn't exist: https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/api/ui/serializers/execution_environment.py#L360. Right now you can't add a new registry on a remote if the registry was deleted.

newswangerd (approved):

PR #996, opened 2021-09-29
Add auth_provider to users/ endpoint to denote an SSO user

ShaiahWren:
Issue: AAH-952
Assignment: Add auth_provider to users/ endpoint to denote an SSO user
See on Jira: https://issues.redhat.com/browse/AAH-952

Creates a new serializer method field, "auth_provider" that checks to see if the user is an SSO user. If so, method returns the SSO provider; if not it returns "django" as a string.

### DIFF
--- a/CHANGES/952.bugfix
+++ b/CHANGES/952.bugfix
@@ -1,0 +1,1 @@
+Add auth_provider to users/ endpoint to denote an SSO user
--- a/galaxy_ng/app/api/ui/serializers/user.py
+++ b/galaxy_ng/app/api/ui/serializers/user.py
@@ -21,6 +21,7 @@ class GroupSerializer(serializers.ModelSerializer):
 
 
 class UserSerializer(serializers.ModelSerializer):
+    auth_provider = serializers.SerializerMethodField()
 
     class Meta:
         model = auth_models.User
@@ -33,12 +34,22 @@ class UserSerializer(serializers.ModelSerializer):
             'groups',
             'password',
             'date_joined',
-            'is_superuser'
+            'is_superuser',
+            'auth_provider',
         )
         extra_kwargs = {
             'date_joined': {'read_only': True},
             'password': {'write_only': True, 'allow_blank': True, 'required': False}
         }
+
+    def get_auth_provider(self, user):
+        if hasattr(user, 'social_auth') and user.social_auth.all():
+            providers = []
+            for social in user.social_auth.all():
+                providers.append(social.provider)
+            return providers
+
+        return ['django']
 
     def validate_password(self, password):
         if password:
--- a/galaxy_ng/app/api/ui/views/settings.py
+++ b/galaxy_ng/app/api/ui/views/settings.py
@@ -14,6 +14,7 @@ class SettingsView(api_base.APIView):
             "GALAXY_ENABLE_UNAUTHENTICATED_COLLECTION_ACCESS",
             "GALAXY_ENABLE_UNAUTHENTICATED_COLLECTION_DOWNLOAD",
             "GALAXY_FEATURE_FLAGS",
+            "GALAXY_TOKEN_EXPIRATION",
         ]
         data = {key: settings.as_dict().get(key, None) for key in keyset}
         return Response(data)


.compose.env.example
```diff
@@ -1,7 +1,7 @@
-# Determines how the dev environment should run. 
+# Determines how the dev environment should run.
```
newswangerd, line 1: get rid of these changes to `.compose.env.example`

galaxy_ng/app/api/ui/serializers/user.py
```diff
@@ -31,13 +31,19 @@ class Meta:
...
             'date_joined',
-            'is_superuser'
+            'is_superuser',
         )
         extra_kwargs = {
             'date_joined': {'read_only': True},
             'password': {'write_only': True, 'allow_blank': True, 'required': False}
         }
 
+    def auth_provider(self, user):
```
newswangerd, line 41: This function is correct, but it's not hooked up to return anything when the API endpoint is called. Do some research on the `SerializerMethodField` field.

newswangerd (approved):

newswangerd (changes requested):
Meant to request changes.

galaxy_ng/app/models/auth.py
```diff
@@ -18,8 +22,7 @@
 
 class User(auth_models.AbstractUser):
     """Custom user model."""
-
-    pass
+    auth_provider = models.CharField(_('auth provider'), max_length=60, blank=True)
```
rochacbruno, line 25: where does this field gets set? does this data come from the UI?

galaxy_ng/app/api/ui/serializers/user.py
```diff
@@ -31,13 +32,20 @@ class Meta:
...
+            'is_superuser',
+            'auth_provider',
         )
         extra_kwargs = {
             'date_joined': {'read_only': True},
             'password': {'write_only': True, 'allow_blank': True, 'required': False}
         }
 
+    def get_auth_provider(self, user):
+        if hasattr(user, 'social_auth'):
```
rochacbruno, line 44: the model has `auth_provider` field, should that field also be exposed here?

galaxy_ng/app/models/auth.py
awcrosby: @rochacbruno I was working with @ShaiahWren yesterday, I walked through the model edit + makemigrations, but then I realized this field is never populated in the database, just computed. 

@ShaiahWren I think if `auth_provider` is removed from UserSerializer.Meta and then spin down/up the containers, it should run without errors. If that all works then you can remove the model edit and the 0026 migration.

awcrosby (changes requested):
see suggestion for `auth_provider` in UserSerializer.Meta

galaxy_ng/app/api/ui/serializers/user.py
ShaiahWren: Could you clarify what you mean here?

galaxy_ng/app/models/auth.py
ShaiahWren: @awcrosby auth_provider is removed from UserSerializer.Meta and I get this error:
Screen Shot 2021-10-04 at 12 13 16 PM

.compose.env.example
ShaiahWren: @awcrosby not sure how to reverse these changes.. seems like these lines were removed and added back

CHANGES/952.bugfix
```diff
@@ -0,0 +1 @@
+Make it clear when a user is an SSO user.
```
awcrosby, line 1: I recommend the PR title (which will likely become the commit title) and change log entry be changed to something like:
`Add auth_provider to users/ endpoint to denote an SSO user`

awcrosby:
Seems to be working, I just tested with my local keycloak setup, I had an admin user created outside keycloak, and logged in with one user that was in the test LDAP server.
```
        ...
            "id": 2,
            "username": "fry",
            "first_name": "Philip J. Fry",
            ...
            "auth_provider": [
                "keycloak"
            ]
       ...
            "id": 1,
            "username": "admin",
            ...
            "auth_provider": [
                "django"
            ]
```

awcrosby (approved):

PR #1005, opened 2021-10-05
Enable azure test at nightly latest workflow

fao89:
No-Issue


rochacbruno:
Container Registry currently doesn´t work on azure right?

If we start adding functional tests for the container registry we will need to skip those tests only on that environment.

Is there an environment variable we can read?

```py
@skip_if(os.getenv("RUNNING_ON_AZURE"))
def ContainerTestCase(....)
...
```

fao89:
@rochacbruno could you please merge it?

PR #1040, opened 2021-10-29
Localization: Update translated strings

rooftopcellist:
This PR pulls in newly translated strings. They were pushed on October 15th. I expect these will need to be backported to 4.4.

PR #1041, opened 2021-11-01
Localization: Update translated strings

Spredzy:
> This PR pulls in newly translated strings. They were pushed on October 15th. I expect these will need to be backported to 4.4.

Supersed of https://github.com/ansible/galaxy_ng/pull/1040


himdel (approved):
Confirming I see no msgid vs msgstr mismatches re format strings.

LGTM :+1:

PR #1056, opened 2021-11-15
Fix last_sync_task field on sync_all_repos_in_registry task

jerabekjiri:
Issue: [AAH-1094](https://issues.redhat.com/browse/AAH-1094)

This fixes the API part of issue AAH-1094.

Task field `reserved_resources_record` doesn't get populated on `galaxy_ng.app.tasks.registry_sync.sync_all_repos_in_registry` action. But we still can get the value from `kwargs` field and return `last_sync_field` which is needed for Registry sync status on UI.

Screenshot from 2021-11-15 16-00-25

### DIFF
--- a/CHANGES/1094.bugfix
+++ b/CHANGES/1094.bugfix
@@ -1,0 +1,1 @@
+Remote registry sync status not shown on registry page
--- a/galaxy_ng/app/api/ui/views/sync.py
+++ b/galaxy_ng/app/api/ui/views/sync.py
@@ -72,5 +72,6 @@ class ContainerSyncRegistryView(api_base.APIView):
             kwargs={
                 "registry_pk": str(registry.pk),
             },
+            exclusive_resources=[registry]
         )
         return OperationPostponedResponse(result, request)


newswangerd (changes requested):
I'm impressed with the solution you've come up with for this. I never would have considered looking for the pk in `kwargs`.

I don't think it's the right way to solve this problem though. The root of the problem is that we aren't locking (reserving) the remote when we launch the sync task, which is actually a pretty major oversight on my part. To fix that, you can add `exclusive_resources=str(registry.pk)` to `dispatch` here: https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/api/ui/views/sync.py#L70

That will lock the remote, and it should show up in the serializer for future tasks. It won't associate any previous tasks with the registry, but that's okay.

mdellweg:
I think, it should be `exclusive_resources=[registry]`. I completely agree with the rest you said.

mdellweg:
> @ newswangerd @ mdellweg thanks for hinting solution. I was actually looking at dispatch `tasks.sync_all_repos_in_registry` in pulp core documentation, but didn't figure it out. :)

I think those are the places to look:
https://docs.pulpproject.org/pulpcore/plugins/plugin-writer/concepts/subclassing/viewsets.html#kick-off-tasks
https://docs.pulpproject.org/pulpcore/plugins/api-reference/tasking.html#pulpcore.plugin.tasking.dispatch
Any suggestion to improve the docs is very much appreciated.

jerabekjiri:
@mdellweg thanks, I'll check it out, I'm new to galaxy_ng and not experienced with pulp at all, so a lot of missteps but slowly learning :)

newswangerd (approved):

PR #1057, opened 2021-11-15
[WIP] Remove django guardian

newswangerd:


galaxy_ng/app/access_control/fields.py
```diff
@@ -2,51 +2,48 @@
...
 from django.utils.translation import gettext_lazy as _
 
-from guardian.shortcuts import get_perms_for_model
 
 from rest_framework import serializers
 from rest_framework.exceptions import ValidationError
 
+from pulpcore.app.models.role import Role
+
+from pulpcore.app.role_util import get_perms_for_model
```
mdellweg, line 11: @bmbouter do you agree, we should have these in the plugin api?

galaxy_ng/app/access_control/fields.py
```diff
@@ -2,51 +2,48 @@
...
-        for perm in perms:
-            if '.' in perm:
-                app_label, codename = perm.split('.', maxsplit=1)
-                filter_q = Q(content_type__app_label=app_label) & Q(codename=codename)
-            else:
-                filter_q = Q(codename=perm)
-
+        for role in roles:
             # TODO(newswangerd): Figure out how to make this one SQL query instead of
             # performing N queries for each permission
```
mdellweg, line 35: How about
```python
missing_roles = set(roles) - set(Role.objects.filter(name__in=roles))
```
?

galaxy_ng/app/access_control/fields.py
bmbouter: I think so; if a plugin needs to use it and it's the right tool for the job then yes.

rochacbruno (approved):

awcrosby (changes requested):
We do have recently added guardian piece that will need to get updated, it is run on each deploy of insights mode: https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/management/commands/maintain-pe-group.py

galaxy_ng/app/management/commands/maintain-pe-group.py
```diff
@@ -18,34 +21,50 @@ class Command(BaseCommand):
...
         ]
+        for app_label, codename in pe_perms:
+            perm = Permission.objects.filter(
+                content_type__app_label=app_label,
+                codename=codename,
+            ).first()
+            if perm:
+                pe_role.permissions.add(perm)
+            else:
+                self.stdout.write(f"Permission {app_label}.{codename} not found.")
```
awcrosby, line 64: If a perm is not found, this will fail semi-silently with stdout. Should we have it fail more loudly with this?
```
perm = Permission.objects.get(content_type__app_label=app_label, codename=codename)
pe_role.permissions.add(perm)
```

galaxy_ng/tests/unit/api/base.py
```diff
@@ -95,51 +95,33 @@ def _create_namespace(name, groups=None):
...
-            'ansible.change_collectionremote',
+            'galaxy.synclist_owner',
         ]
         pe_group = auth_models.Group.objects.create(
             name='partner-engineers')
 
-        for perm in pe_perms:
-            assign_perm(perm, pe_group)
+        for role in pe_roles:
+            assign_role(role, pe_group)
```
awcrosby, line 125: We should have the `pe_group` in unit tests & live environments `maintain-pe-group.py` match
* Unit test `pe_group` has various pre-existing roles
* Management command `pe_group` has 1 new role with 14 perms

I assume using pre-existing roles (is the official term "locked"?) is the preferred way to do it

awcrosby (dismissed):

galaxy_ng/app/access_control/statements/roles.py
```diff
@@ -36,6 +36,18 @@
...
+                "galaxy.view_group",
+                "galaxy.delete_group",
+                "galaxy.add_group",
+                "galaxy.change_group",
+            ],
+            "galaxy.user_admin": [
+                "galaxy.view_user",
+                "galaxy.delete_user",
+                "galaxy.add_user",
+                "galaxy.change_user",
```
awcrosby, line 49: I see `LOCKED_ROLES` only exist under a specific ViewSet in the `VIEWSETS` dictionary. How do these apply to the ViewSet they are under? Once a locked role is in the database, can it be used as a system-wide role that is independent of the ViewSet?

galaxy_ng/tests/unit/api/base.py
awcrosby: I am not seeing `ansible.modify_ansible_repo_content` in any of the locked roles. If we use only locked roles for the pe_group, I guess we would need to add that to a new or existing locked role

galaxy_ng/app/access_control/statements/roles.py
mdellweg: Yes. Keeping it with the viewset is just a convenient convention.

galaxy_ng/tests/unit/api/base.py
bmclaughlin: It seems the unit test `pe_group` has been augmented over time.  I'll correct that and create new users with proper roles for those soon to be failing tests.

`maintain-pe-group.py` is now using locked roles (including the newly created `galaxy.content_admin` locked role)

galaxy_ng/app/management/commands/maintain-pe-group.py
bmclaughlin: I believe this is no longer an issue due to going with locked roles here.

awcrosby:
@bmclaughlin spinning up an ephemeral env and trying to create a namespace in the UI, I added a group to "Namespace owners" but the POST hit an error:
```
"status":"400",
"title":"Invalid input.",
"detail":"object_roles field is required",
"source":{"parameter":"groups"}
```

newswangerd:
> @bmclaughlin spinning up an ephemeral env and trying to create a namespace in the UI, I added a group to "Namespace owners" but the POST hit an error:
> 
> ```
> "status":"400",
> "title":"Invalid input.",
> "detail":"object_roles field is required",
> "source":{"parameter":"groups"}
> ```

@awcrosby the UI still needs to be updated to work with these changes. We've had to make some breaking changes to this api by removing the ability to add permissions to group/object combinations, since that's no longer supported by pulpcore.

bmclaughlin (approved):

PR #1060, opened 2021-11-18
Clowder: support IMPORTER_IMAGE_TAG to be defined in config

awcrosby:
The galaxy-importer uses a quay image to run ansible-test sanity as an
openshift job. This change allows this image tag to be set in our
config.

No-Issue


drodowic:
By setting the IMPORTER_IMAGE_TAG parameter as required but not providing a default value, it will have to be explicitly set when deploying to ephemeral, including during the pr check. e.g. https://github.com/ansible/galaxy_ng/blob/master/pr_check.sh#L33

I recommend setting "  value: latest" to the parameter definition

drodowic (approved):
/lgtm

PR #1061, opened 2021-11-18
Support IMPORTER_IMAGE_TAG to be defined in config (#1060)

awcrosby:
The galaxy-importer uses a quay image to run ansible-test sanity as an
openshift job. This change allows this image tag to be set in our
config.

No-Issue


drodowic (approved):
/lgtm

PR #1069, opened 2021-12-02
Add additional logging to custom import task

bmclaughlin:
Adds conditional logging based on if `GALAXY_ENABLE_API_ACCESS_LOG` is enabled then logs collection details to the api access log in the `galaxy_ng_worker_1` container where the uploads occur.

Log: 
```
2021-12-02 18:13:54,441: INFO: Anonymous uploaded collection: newswangerd-collection_demo-1.0.11
```

Issue: AAH-1018


galaxy_ng/app/tasks/publishing.py
```diff
@@ -61,6 +67,14 @@ def import_and_move_to_staging(temp_file_pk, **kwargs):
         call_copy_task(collection_version, inbound_repo, staging_repo)
         call_remove_task(collection_version, inbound_repo)
 
+        if settings.GALAXY_ENABLE_API_ACCESS_LOG:
+            api_access_log.info(
+                "Anonymous uploaded collection: %s-%s-%s",
```
awcrosby, line 72: Since only authenticated users can import a collection, is there a reason we are saying "Anonymous..." versus "Collection uploaded: ..."?

galaxy_ng/app/tasks/publishing.py
```diff
@@ -3,15 +3,21 @@
...
-from pulpcore.plugin.models import Task
 from pulp_ansible.app.models import AnsibleDistribution, AnsibleRepository, CollectionVersion
 from pulp_ansible.app.tasks.collections import import_collection
+from pulpcore.plugin.models import Task
 
 from .promotion import call_copy_task, call_remove_task
 
 log = logging.getLogger(__name__)
 
+api_access_log = logging.getLogger("automated_logging")
```
awcrosby, line 14: im not sure if logging config should be done in a tasking module, especially globally. This logger will always get created on module loading regardless of `GALAXY_ENABLE_API_ACCESS_LOG`,

galaxy_ng/app/tasks/publishing.py
```diff
@@ -3,15 +3,21 @@
...
+formatter = logging.Formatter("%(asctime)s %(levelname)s %(message)s")
+handler = logging.FileHandler("/var/log/galaxy_api_access.log")
+handler.setFormatter(formatter)
+api_access_log.setLevel(logging.INFO)
+api_access_log.addHandler(handler)
```
awcrosby, line 19: i assume this logger is accessible to django throughout app, can we use it? https://github.com/ansible/galaxy_ng/blob/c2eb855c2229dc20adf894e8d7887006b14f966a/galaxy_ng/app/dynaconf_hooks.py#L180-L185

galaxy_ng/app/tasks/publishing.py
bmclaughlin: I don't believe we have access to the `request` object to determine who the user is, hence the `Anonymous`

galaxy_ng/app/tasks/publishing.py
bmclaughlin: Apologies, should have been wrapped it in a conditional based on `GALAXY_ENABLE_API_ACCESS_LOG`

galaxy_ng/app/tasks/publishing.py
bmclaughlin: Ha, nothing but errors yesterday when trying to reuse the global logger.  Today, worked immediately.  :man_shrugging:

galaxy_ng/app/tasks/publishing.py
awcrosby: I think this should be `"Collection uploaded: %s-%s-%s"` - folks reading the log may think the app allows anonymous users to access the system and uploaded a collection.

awcrosby (changes requested):

galaxy_ng/app/tasks/publishing.py
```diff
@@ -94,3 +104,11 @@ def import_and_auto_approve(temp_file_pk, **kwargs):
         log.info('Imported and auto approved collection artifact %s to repository %s',
                  collection_version.relative_path,
                  golden_repo.latest_version())
+
+        if settings.GALAXY_ENABLE_API_ACCESS_LOG:
```
awcrosby, line 108: are there any unit or functional tests for api logging where this msg could be added?

galaxy_ng/tests/functional/cli/test_collection_upload.py
```diff
@@ -44,3 +45,54 @@ def test_upload_collection(self):
         self.assertEqual(collection.namespace, "pulp")
         self.assertEqual(collection.name, "squeezer")
         self.assertEqual(collection.highest_version["version"], "0.0.9")
+
+        # Cleanup
+        self.delete_collection(collection.namespace, collection.name)
+        self.delete_namespace(collection.namespace)
+
+    @skip("Requires GALAXY_ENABLE_API_ACCESS_LOG to be enabled")
```
awcrosby, line 53: I think before this PR can be merged we need to either
- Find/build way to export env var GALAXY_ENABLE_API_ACCESS_LOG so this test is not skipped
- Move this test to IQE, test against pr branch, merge test

newswangerd (approved):

awcrosby (approved):

PR #1080, opened 2021-12-14
Update to pulp-container 2.8.3

newswangerd:
Issue: AAH-1188


awcrosby (approved):

.github/workflows/ci.yml
```diff
@@ -82,7 +82,6 @@ jobs:
       matrix:
         env:
           - TEST: pulp
-          - TEST: azure
```
fao89, line 85: why removing azure tests?

newswangerd: @fao89 I just ran the plugin template. I'm not sure why it did that

fao89: I think your plugin-template is outdated

newswangerd: looks like it was. Good catch

PR #1088, opened 2021-12-20
Skip checking commit messages from dependabot (ui repo)

jctanner:
Issue: AAH-1106

Signed-off-by: James Tanner <tanner.jc@gmail.com>

### DIFF
--- a/.ci/scripts/validate_commit_message_custom.py
+++ b/.ci/scripts/validate_commit_message_custom.py
@@ -124,11 +124,15 @@ def validate_pr_commits(github_pr_commits_url):
 def main():
     logging.basicConfig(level=logging.INFO, format="%(levelname)s: %(message)s")
 
+    github_user = os.environ.get("GITHUB_USER")
     github_pr_commits_url = os.environ["GITHUB_PR_COMMITS_URL"]
     start_commit = os.environ["START_COMMIT"]
     end_commit = os.environ["END_COMMIT"]
+    skip_users = ['dependabot[bot]', 'patchback[bot]']
 
-    if github_pr_commits_url:
+    if github_user in skip_users:
+        is_valid = True
+    elif github_pr_commits_url:
         is_valid = validate_pr_commits(github_pr_commits_url)
     else:
         is_valid = validate_push_commits(start_commit, end_commit)
--- a/CHANGES/1106.misc
+++ b/CHANGES/1106.misc
@@ -1,0 +1,1 @@
+normalize commit checking to use the galaxy_ng repo as a source of truth


.github/workflows/ci.yml
```diff
@@ -22,15 +22,8 @@ jobs:
 
       - name: Run script to validate commits for both pull request and a push
         env:
-          PY_COLORS: '1'
```
rochacbruno, line 25: https://github.com/jctanner/galaxy_ng/blob/SKIP_DEPENDABOT/.github/workflows/ci.yml#L1-L6

.github/workflows/ci.yml
```diff
@@ -22,15 +22,8 @@ jobs:
...
-          ANSIBLE_FORCE_COLOR: '1'
-          GITHUB_PULL_REQUEST: ${{ github.event.number }}
-          GITHUB_PULL_REQUEST_BODY: ${{ github.event.pull_request.body }}
-          GITHUB_BRANCH: ${{ github.head_ref }}
-          GITHUB_REPO_SLUG: ${{ github.repository }}
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          GITHUB_CONTEXT: ${{ github.event.pull_request.commits_url }}
           GITHUB_PR_COMMITS_URL: ${{ github.event.pull_request.commits_url }}
+          GITHUB_USER: ${{ github.event.pull_request.user.login }}
```
rochacbruno, line 26: we can add more vars in 2 ways

1. https://github.com/ansible/galaxy_ng/blob/master/template_config.yml#L26 adding as a config param
2. add as part of template https://github.com/pulp/plugin_template/blob/master/templates/macros.j2#L1-L18

rochacbruno (changes requested):
There are some files on the CI that we can't change directly, we need to make changes on template_condig.yml or to template repo.

awcrosby (dismissed):

.ci/scripts/validate_commit_message_custom.py
```diff
@@ -124,11 +124,15 @@ def validate_pr_commits(github_pr_commits_url):
 def main():
     logging.basicConfig(level=logging.INFO, format="%(levelname)s: %(message)s")
 
+    github_user = os.environ.get("GITHUB_USER")
```
awcrosby, line 127: where will `GITHUB_USER` get into the environment?

awcrosby (commented):
where will `GITHUB_USER` from `validate_commit_message_custom.py` get into the environment?

rochacbruno:
looks like there is a GITHUB_ACTOR added by default https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables

rochacbruno (approved):
not sure about the source of GITHUB_USER var, everything else LGTM

himdel:
> not sure about the source of GITHUB_USER var

For UI, that's https://github.com/ansible/ansible-hub-ui/pull/1380/files#diff-1da7a27709d11ac16eef3de8e46ed10672b7b1d63a790ffbcbd23a94b19efdcbR22

Though maybe we should provide the env var in galaxy_ng as well? (We're probably not using patchback here, but dependabot yes.)

jctanner:
@himdel alterting the workflow file in this repo would require an edit to https://github.com/ansible/galaxy_ng/blob/master/template_config.yml#L26  and then a regeneration with https://github.com/pulp/plugin_template which would add the env var to *every* task in the workflow file. I'd rather not have to deal with that right now.

himdel:
> I'd rather not have to deal with that right now.

Understood, yeah it's a bit too complicated now :)
I'm fine with galaxy_ng not providing any usenames, seems like we don't *need* to for now :+1:

PR #1098, opened 2022-01-12
API Refactor stage 1

newswangerd:
Issue: [AAH-765](https://issues.redhat.com/browse/AAH-765)

This PR requires https://github.com/pulp/pulp_ansible/pull/849 to run.

This PR does the following:
- Makes access policies for pulp ansible views configurable in galaxy ng
- Imports the pulp ansible APIs directly into galaxy ng

Blocked by: https://github.com/pulp/pulp_ansible/pull/888


newswangerd:
Converting this to a draft PR because I'm working on this in tandem with https://issues.redhat.com/browse/AAH-753, and I want to make sure the two issues work together before merging anything.

galaxy_ng/app/api/ui/urls.py
```diff
@@ -156,12 +156,12 @@
     path('controllers/', views.ControllerListView.as_view(), name='controllers'),
     path('groups/', include(group_paths)),
     path(
-        'repo/<str:path>/',
+        'repo/<str:distro_base_path>/',
```
mdellweg, line 159: I see this change reflected in the pulp_ansible pr with no further hints there.
What is the final goal? Should it eventually be called `distro_base_path` in all places?

galaxy_ng/app/dynaconf_hooks.py
```diff
@@ -19,7 +19,8 @@ def post(settings: Dynaconf) -> Dict[str, Any]:
     data.update(configure_logging(settings))
     data.update(configure_keycloak(settings))
     data.update(configure_cors(settings))
-    data.update(configure_feature_flags(settings))
+    data.update(configure_pulp_ansible(settings))
+    data.update(configure_pulp_ansible(settings))
```
rochacbruno, line 23: You were not supposed to remove `data.update(configure_feature_flags(settings))` from here, this call to configure pulp_ansible is duplicated.

galaxy_ng/app/dynaconf_hooks.py
```diff
@@ -279,3 +280,10 @@ def configure_feature_flags(settings: Dynaconf) -> Dict[str, Any]:
     data["GALAXY_FEATURE_FLAGS__collection_auto_sign"] = settings.get(
         "GALAXY_AUTO_SIGN_COLLECTIONS", False)
     return data
+
+
+def configure_pulp_ansible(settings: Dynaconf) -> Dict[str, Any]:
+    return {
+        "ANSIBLE_URL_NAMESPACE": "galaxy:api:v3:",
+        "ANSIBLE_DEFAULT_DISTRIBUTION_PATH": "published"
+    }
```
rochacbruno, line 289: The main reason to use dynaconf_hooks is to be able to add conditional settings, in this case you don't have any kind of conditional. can't this be just variables on `settings.py` or you need it to be set on the late stage of settings loading?

rochacbruno: The problem with this approach is that this is no more overridable via environment variables, as hooks runs after envvars are loaded the values set here are the final one.

You can at least do:

```suggestion
import os

def configure_pulp_ansible(settings: Dynaconf) -> Dict[str, Any]:
    return {
        "ANSIBLE_URL_NAMESPACE": os.getenv("PULP_ANSIBLE_URL_NAMESPACE", "galaxy:api:v3:"),
        "ANSIBLE_DEFAULT_DISTRIBUTION_PATH": os.getenv("PULP_ANSIBLE_DEFAULT_DISTRIBUTION_PATH", "published")
    }
```

galaxy_ng/app/settings.py
```diff
@@ -55,11 +55,21 @@
...
+
+# TODO: remove GALAXY_AUTHENTICATION_CLASSES classes and use DEFAULT_AUTHENTICATION_CLASSES instead
 GALAXY_AUTHENTICATION_CLASSES = [
     "rest_framework.authentication.SessionAuthentication",
     "rest_framework.authentication.TokenAuthentication",
     "rest_framework.authentication.BasicAuthentication",
 ]
+
+REST_FRAMEWORK__DEFAULT_PERMISSION_CLASSES = \
+    ("galaxy_ng.app.access_control.access_policy.AccessPolicyBase", )
```
rochacbruno, line 67:
```suggestion
REST_FRAMEWORK__DEFAULT_PERMISSION_CLASSES = (
    "galaxy_ng.app.access_control.access_policy.AccessPolicyBase", 
)
```

galaxy_ng/app/api/v3/viewsets/collection.py
```diff
@@ -271,8 +78,14 @@ def _get_path(kwargs, filename_ns):
...
            then use an inbound distribution based on filename namespace.
         """
-        path = kwargs['path']
-        if kwargs.get('no_path_specified', None):
+
+        # the legacy collection upload views don't get redirected and still have to use the
+        # old path arg
+        path = kwargs.get(
+            'distro_base_path',
+            kwargs.get('path', settings.ANSIBLE_DEFAULT_DISTRIBUTION_PATH))
```
rochacbruno, line 86:
```suggestion
        path = kwargs.get(
            'distro_base_path',
            kwargs.get('path', settings.ANSIBLE_DEFAULT_DISTRIBUTION_PATH)
        )
```

rochacbruno (changes requested):

galaxy_ng/app/dynaconf_hooks.py
newswangerd: @rochacbruno `ANSIBLE_URL_NAMESPACE` can't really be set to anything else or it will break all of our APIs, so that doesn't bother me.

I would like to set these in settings.py, but the pulp ansible settings file takes prescendence over ours, and I couldn't think of any other way to easily override these.

rochacbruno: in that case  that is ok

galaxy_ng/app/dynaconf_hooks.py
```diff
@@ -279,3 +275,35 @@ def configure_feature_flags(settings: Dynaconf) -> Dict[str, Any]:
...
+    # the pulp APIs use the same authentication as the galaxy APIs. Rather than setting
+    # the galaxy auth classes and the DRF classes in all those environments just set the
+    # default rest framework auth classes to the galaxy auth classes. Ideally we should
+    # switch everything to use the default DRF auth classes, but given how many
+    # environments would have to be reconfigured, this is a lot easier.
+    galaxy_auth_classes = settings.get("GALAXY_AUTHENTICATION_CLASSES", None)
+
+    if galaxy_auth_classes:
+        return {
+            "REST_FRAMEWORK__DEFAULT_AUTHENTICATION_CLASSES": galaxy_auth_classes
```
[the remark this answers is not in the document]
newswangerd: Thanks!

setup.py
```diff
@@ -90,7 +90,7 @@ def run(self):
 requirements = [
     "galaxy-importer==0.4.2",
     "pulpcore>=3.18.1,<3.19.0",
-    "pulp-ansible>=0.12.0,<0.13.0",
+    "pulp-ansible@git+https://github.com/pulp/pulp_ansible@main#egg=pulp-ansible",  # noqa E501
```
awcrosby, line 93: if our commit needs to include a non-released pulp-ansible, this should be a commit hash rather than `main` so we can have reproducible builds

awcrosby (changes requested):

setup.py
newswangerd: @awcrosby I couldn't get a commit to work with the plugin template. Is it okay to set setup.py to a commit and the plugin template to main?

awcrosby: if the `requirements/` files have the commit fed from setup.py, that should be good enough for reproducible builds. for now i think that is ok.

galaxy_ng/app/access_control/statements/insights.py
```diff
@@ -1,4 +1,87 @@
...
+        "action": "create",
+        "principal": "authenticated",
+        "effect": "allow",
+        "condition": ["can_create_collection", "has_rh_entitlements"],
+    },
+    {
+        "action": "update",
+        "principal": "authenticated",
+        "effect": "allow",
+        "condition": "can_update_collection",
```
awcrosby, line 38: it seems we have been missing `has_rh_entitlements` on the update action, we should be able to add it

awcrosby (approved):

jctanner (approved):
Just waiting for tests to pass and report back.

PR #1106, opened 2022-01-26
Update CI

fao89:
I noticed CI update workflow is broken: https://github.com/ansible/galaxy_ng/actions/workflows/update_ci.yml
So I manually update it from plugin-template


template_config.yml
```diff
@@ -85,11 +86,13 @@ tasking_allow_async_unsafe: true
 test_azure: true
 test_bindings: false
 test_cli: false
+test_deprecations: true
```
fao89, line 89: PR fails because we are testing pulpcore deprecations

.ci/ansible/Containerfile.j2
```diff
@@ -1,4 +1,4 @@
-FROM {{ ci_base | default("pulp/pulp-ci-centos:" + pulp_container_tag) }}
+FROM {{ ci_base | default("ghcr.io/pulp/pulp-ci-centos:" + pulp_container_tag) }}
```
fao89, line 1: this was my main motivator for this PR, use github container registry instead of docker

template_config.yml
rochacbruno: @fao89 should we keep testing it? or change to false in this PR?

fao89: I don't know, it is up to galaxy
I personally think this deprecation test could be moved to your nightly latest

rochacbruno (approved):

fao89:
could you please merge?

fao89:
could you please merge?

PR #1076, opened 2021-12-09
Add Keys, Script and Signing Service to dev environment

rochacbruno:
This PR adds the following features

- [x] Dev environment GPG keys
- [x] Dev environment Signing Script
- [x] Dev environment Signing Service creation
- [x] (the same as above for Github CI test runs)
- [x] Settings variables for signing configuration
- [x] Auto sign on collection approval
- [x] endpoints to perform on-demand signing
- [x] signatures exposed on v3/collection/ endpoint (for CLI)
- [x] signatures and sign states exposed on ui/v1/ endpoints (for UI)
- [x] settings and feature flags exposed on /settings/ and /feature-flags/ (for UI)
- [x] permissions exposed on /me/ (for UI)
- [x] Filtering by sign state (for UI endpoint)
- [x] Logging for collection signatures
- [x] Setup Signing Service on ephemeral (by @drodowic ) 
- [ ] Functional testing for collection signatures
- [x] Sync `signed_only` content field on Remote configuration

## metadata

Issue: AAH-1181
env:LOCK_REQUIREMENTS=0
env:PULP_ANSIBLE_REVISION=main

PR #1111, opened 2022-01-27
[WIP] enable collection signing for ephemeral and c.rh.c

drodowic:
Issue: AAH-1247


docker/entrypoint.sh
```diff
@@ -100,6 +102,17 @@ run_manage() {
     exec django-admin "$@"
 }
 
+setup_signing_service() {
```
rochacbruno, line 105: This file is the same used by dev environment https://github.com/ansible/galaxy_ng/pull/1076/files#diff-4f5cabe26761257a4d685a6edc7a43e0fe0f78762f50eeb48530f2bd3b3ee7ca we need to make it in a way it works the same code for both environments, or conditionally select a different function for cloud.

drodowic: The function you wrote should work for the most part in the cloud environments. Just the KEY_ID will still need to be parameterized, instead of requiring galaxy3.
Other than that, a new ENV var would also be required instead of re-using $DEV_SOURCE_PATH. e.g. https://github.com/ansible/galaxy_ng/pull/1111/commits/766e16ab2d8ab53005b648f1364942075ec126aa

rochacbruno:
@drodowic I merged your changes to https://github.com/ansible/galaxy_ng/pull/1076

PR #1130, opened 2022-02-14
Add landing page endpoint for the console redhat

brumik:
Issue: [AAH-1278](https://issues.redhat.com/browse/AAH-1278)

This commit adds the endpoints for the landing page on the console.redhat.com. It is going to be a self service so all
the dynamic data should be prefilled: collection and partner  in the current case.

Next after this is deployed to STAGE by this [document](https://issues.redhat.com/browse/RHCLOUD-17405)
* I'll contact the platform to add this endpoint to chrome
* After they add the endpoint QE can verify the looks and data displayed on STAGE 

---

The currently used schema is here: https://github.com/RedHatInsights/landing-page-frontend/blob/b1f358352916f3f6b4813e9cfe0b4a7765005109/src/contentApi/ansible-api.js#L500

The expected prefilled format example: https://github.com/RedHatInsights/landing-page-frontend/blob/master/docs/exmple-schema.md

### DIFF
--- a/galaxy_ng/app/api/ui/views/landing_page.py
+++ b/galaxy_ng/app/api/ui/views/landing_page.py
@@ -1,0 +1,216 @@
+from galaxy_ng.app.access_control import access_policy
+from random import sample
+from rest_framework.response import Response
+from pulp_ansible.app.models import CollectionVersion, AnsibleDistribution
+from galaxy_ng.app.models import Namespace
+from galaxy_ng.app import settings
+from galaxy_ng.app.api import base as api_base
+
+
+class LandingPageView(api_base.APIView):
+    permission_classes = [access_policy.LandingPageAccessPolicy]
+    action = "retrieve"
+
+    def get(self, request, *args, **kwargs):
+        golden_name = settings.GALAXY_API_DEFAULT_DISTRIBUTION_BASE_PATH
+
+        distro = AnsibleDistribution.objects.get(base_path=golden_name)
+        repository_version = distro.repository.latest_version()
+        collection_count = CollectionVersion.objects.filter(
+            pk__in=repository_version.content, is_highest=True
+        ).count()
+
+        partner_count = Namespace.objects.count()
+
+        # If there are no partners dont show the recommendation for it
+        recommendations = {}
+        if partner_count > 0:
+            namespace = sample(list(Namespace.objects.all()), 1)[0]
+            recommendations = {
+                "recs": [
+                    {
+                        "id": "ansible-partner",
+                        "icon": "bulb",
+                        "action": {
+                            "title": "Check out our partner %s" % namespace.company,
+                            "href": "./ansible/automation-hub/partners/%s" % namespace.name,
+                        },
+                        "description": "Discover automation from our partners.",
+                        "permissions": [
+                            {
+                                "method": "isEntitled",
+                                "args": ["ansible"],
+                            }
+                        ]
+                    }
+                ]
+            }
+
+        data = {
+            "estate": {
+                "items": [
+                    {
+                        "id": "ansible-collections",
+                        "count": collection_count,
+                        "shape": {
+                            "title": "Collections",
+                            "href": "./ansible/automation-hub/",
+                        },
+                        "permissions": [
+                            {
+                                "method": "isEntitled",
+                                "args": ["ansible"],
+                            }
+                        ],
+                    },
+                    {
+                        "id": "ansible-partners",
+                        "count": partner_count,
+                        "shape": {
+                            "title": "Partners",
+                            "href": "./ansible/automation-hub/partners/",
+                        },
+                        "permissions": [
+                            {
+                                "method": "isEntitled",
+                                "args": ["ansible"],
+                            }
+                        ],
+                    },
+                ],
+            },
+            "recommendations": recommendations,
+            "configTryLearn": {
+                "configure": [
+                    {
+                        "permissions": [
+                            {
+                                "method": "isEntitled",
+                                "args": ["ansible"],
+                            }
+                        ],
+                        "shape": {
+                            "title": "Sync Red Hat certified collections",
+                            "description": (
+                                "Configure access to sync collections ",
+                                "to Private Automation Hub.",
+                            ),
+                            "link": {
+                                "title": "Get started",
+                                "href": "./ansible/automation-hub/token",
+                            },
+                        },
+                    }
+                ],
+                "try": [
+                    {
+                        "permissions": [
+                            {
+                                "method": "isEntitled",
+                                "args": ["ansible"],
+                            }
+                        ],
+                        "shape": {
+                            "title": "Install Private Automation Hub",
+                            "link": {
+                                "title": "Get started",
+                                "external": True,
+                                "href": (
+                                    "https://access.redhat.com/documentation/en-us/"
+                                    "red_hat_ansible_automation_platform/2.1/html/"
+                                    "red_hat_ansible_automation_platform_installation_guide/index"
+                                ),
+                            },
+                        },
+                    },
+                    {
+                        "permissions": [
+                            {
+                                "method": "isEntitled",
+                                "args": ["ansible"],
+                            }
+                        ],
+                        "shape": {
+                            "title": "Manage repositories in Private Automation Hub",
+                            "description": (
+                                "Add community and privately developed collections "
+                                "to your Private Automation Hub."
+                            ),
+                            "link": {
+                                "title": "Get started",
+                                "external": True,
+                                "href": (
+                                    "https://access.redhat.com/documentation/en-us/"
+                                    "red_hat_ansible_automation_platform/2.1/html/"
+                                    "publishing_proprietary_content_collections_in_"
+                                    "automation_hub/index"
+                                ),
+                            },
+                        },
+                    },
+                ],
+                "learn": [
+                    {
+                        "permissions": [
+                            {
+                                "method": "isEntitled",
+                                "args": ["ansible"],
+                            }
+                        ],
+                        "shape": {
+                            "title": "Connect Automation Hub to your automation infrastructure",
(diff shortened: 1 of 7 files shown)


galaxy_ng/app/api/ui/views/landing_page.py
```diff
@@ -0,0 +1,153 @@
...
+from galaxy_ng.app.models import Namespace
+
+from galaxy_ng.app.api import base as api_base
+
+
+class LandingPageView(api_base.APIView):
+    permission_classes = [IsAuthenticatedOrReadOnly]
+
+    def get(self, request, *args, **kwargs):
+        collection_count = Collection.objects.count()
```
rochacbruno, line 13: This needs more filtering, not all `Collections` should be counted, we want to show the count only on `published` collections.

This queryset needs to filter only the collections that are on `/published` repo. (we have a constant GOLDEN_REPO on constants file)

galaxy_ng/app/api/ui/views/landing_page.py
```diff
@@ -0,0 +1,153 @@
...
+        partner_count = Namespace.objects.count()
```
rochacbruno, line 14: I am not sure all namespaces are partners, @awcrosby and @newswangerd can confirm, but maybe we need filtering also here.

rochacbruno (changes requested):

galaxy_ng/app/api/ui/views/landing_page.py
brumik: The current one is using the `v3/namespace/` endpoint with `meta.count` accessor.

galaxy_ng/app/api/ui/views/landing_page.py
brumik: @rochacbruno can you tell me which file it is? I cannot find anything repo related constant.

brumik: And then the code would look like this?

```py
distro = AnsibleDistribution.objects.get(base_path=golden_name)
repository_version = distro.repository.latest_version()
collection_count = Collection.objects.filter(pk__in=repository_version.content).count()
```

rochacbruno: @brumik yes, sorry the proper one is that https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/settings.py#L37

galaxy_ng/app/api/ui/views/landing_page.py
```diff
@@ -0,0 +1,161 @@
...
+from rest_framework.response import Response
+from pulp_ansible.app.models import (
+    CollectionVersion,
+    AnsibleDistribution
+)
+from galaxy_ng.app.models import Namespace
+from django.conf import settings
+from galaxy_ng.app.api import base as api_base
+
+golden_name = settings.GALAXY_API_DEFAULT_DISTRIBUTION_BASE_PATH
```
newswangerd, line 11: move this into `get`. You're not using it in any other methods, so it's best not to pollute the global namespace.

galaxy_ng/app/api/ui/views/landing_page.py
```diff
@@ -0,0 +1,161 @@
...
+
+
+class LandingPageView(api_base.APIView):
+    permission_classes = [IsAuthenticatedOrReadOnly]
```
newswangerd, line 15: You should use an access policy for this. Docs on RBAC can be found here: https://github.com/ansible/galaxy_ng/wiki/RBAC-in-Galaxy-NG

Access policies are declared here: https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/access_control/access_policy.py

Policy statements are declared here: https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/access_control/statements/insights.py

The statement for this can look like this. This will grant permissions on users that have the correct Red Hat entitlements.

```
        {
            "action": ["retrieve"],
            "principal": "authenticated",
            "effect": "allow",
            "condition": "has_rh_entitlements",
        },
```

newswangerd (changes requested):
Python code looks pretty good overall. Added some comments about the data that's being returned as well as some improvements to the view code.

rochacbruno (approved):

galaxy_ng/app/api/ui/views/landing_page.py
```diff
@@ -0,0 +1,161 @@
...
+                        'permissions': [{
+                            'method': 'isEntitled',
+                            'args': ['ansible'],
+                        }],
+                        'shape': {
+                            'title': 'Connect Automation Hub to your automation infrastructure',
+                            'link': {
+                                'title': 'Get started',
+                                'external': True,
+                                'href': 'https://docs.ansible.com/ansible-tower/latest/html/userguide/\
```
newswangerd, line 118: I'm not sure `\` works correctly for escaping new lines here. I think you'll end up with all the spaces in the url.

brumik: I gonna rest it today, i had the same thoughts, thanks for the reminder

galaxy_ng/app/api/ui/views/landing_page.py
brumik: @newswangerd I am not sure how to do this, this is a view and not a viewset, and as it is currently it has to have the `permission_classes` which does not accepts this.

galaxy_ng/app/api/ui/views/landing_page.py
brumik: Fixed, thanks

galaxy_ng/app/api/ui/views/landing_page.py
```diff
@@ -0,0 +1,161 @@
...
+                            'href': './ansible/automation-hub/partners/',
+                        },
+                        'permissions': [{
+                            'method': 'isEntitled',
+                            'args': ['ansible'],
+                        }],
+                    },
+                ],
+            },
+            'recommendations': {},  # Hub currently does not have recommendation items.
```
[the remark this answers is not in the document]
brumik: The daynum has the chance not showing up some of the namespaces, so to avoid discrimination I used a random generator to do this (if we mind that on every reload there could be a different one I can go with the daynum as you wrote above). I also did my best to do the title and description it, but I probably need to talk to somebody what should it actually say.

galaxy_ng/app/api/ui/views/landing_page.py
```diff
@@ -0,0 +1,161 @@
...
+                            },
+                        },
+                    },
+                    {
+                        'permissions': [{
+                            'method': 'isEntitled',
+                            'args': ['ansible'],
+                        }],
+                        'shape': {
+                            'title': 'Explore Red Hat certified collections',
```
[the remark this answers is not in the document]
brumik: I looked into it, but then it would be the same url then the other one. Or did you meant a different page than the `collections` page with the list of collections and the toolbar for search?

galaxy_ng/app/api/ui/views/landing_page.py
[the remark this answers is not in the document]
brumik: Taufique and Tina approved the current looks (icon + title + description)

galaxy_ng/app/api/ui/views/landing_page.py
newswangerd: @brumik you set the access policy class as the `permission_classes`. Here's an example on one of the views for execution environments: https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/api/ui/views/index_execution_environments.py#L22

That should be a good example to follow

galaxy_ng/app/api/ui/views/landing_page.py
```diff
@@ -0,0 +1,210 @@
...
+            "recommendations": {
+                "recs": [
+                    {
+                        "id": "ansible-partner",
+                        "icon": "bulb",
+                        "action": {
+                            "title": "Check out our partner %s" % namespace.name,
+                            "href": "./ansible/automation-hub/partners/%s" % namespace.name,
+                        },
+                        "description": "Explore a recommended partner from Automation Hub.",
```
newswangerd, line 67: "Explore a recommended partner from Automation Hub" makes me a little nervous. We don't want to sound like we're playing favorites.

"Discover automation from our partners" might be better. Check in with @jonnyfiveiq about this.

galaxy_ng/app/api/ui/views/landing_page.py
```diff
@@ -0,0 +1,210 @@
...
+                    },
+                ],
+            },
+            "recommendations": {
+                "recs": [
+                    {
+                        "id": "ansible-partner",
+                        "icon": "bulb",
+                        "action": {
+                            "title": "Check out our partner %s" % namespace.name,
```
newswangerd, line 64: I'd recommend `namespace.company` here. That will show the more human readable name instead of the internal namespace value.

newswangerd (changes requested):

galaxy_ng/app/api/ui/views/landing_page.py
```diff
@@ -0,0 +1,210 @@
...
+
+class LandingPageView(api_base.APIView):
+    permission_classes = [access_policy.LandingPageAccessPolicy]
+
+    def get(self, request, *args, **kwargs):
+        golden_name = settings.GALAXY_API_DEFAULT_DISTRIBUTION_BASE_PATH
+
+        distro = AnsibleDistribution.objects.get(base_path=golden_name)
+        repository_version = distro.repository.latest_version()
+        versions = CollectionVersion.objects.filter(pk__in=repository_version.content)
```
newswangerd, line 18: this is gonna return the number of collection versions, not the number of collections. `versions = CollectionVersion.objects.filter(pk__in=repository_version.content, is_highest=True)` should do the trick

newswangerd (approved):

PR #1176, opened 2022-03-15
Reorder lint_po to be before longer pulp checks

awcrosby:
Move `lint_po` ci check before checks such as `test (pulp)`, since linting issues fail `test (pulp)` before `lint_po` can run.


himdel (approved):
LGTM :+1: 

I think the original idea behind waiting for tests was just not to prevent their running if the linter fails, but this aligns it more closely with the other linters, makes sense :)

PR #1179, opened 2022-03-18
Allowing unauth access when the settings is set

brumik:
Issue: AAH-1386

# Description 🛠
The settings flag `GALAXY_ENABLE_UNAUTHENTICATED_COLLECTION_DOWNLOAD` only affected the UI until now. So the `ansible-galaxy` installs were failing as the could not access the `/content/...` path without authentication. This PR aims to fix that.

# Reviewer Checklists  👀
Developer reviewer:

### DIFF
--- a/galaxy_ng/app/access_control/statements/standalone.py
+++ b/galaxy_ng/app/access_control/statements/standalone.py
@@ -98,7 +98,28 @@ STANDALONE_STATEMENTS = {
             "action": "*",
             "principal": "authenticated",
             "effect": "allow",
+        },
+
+        # we need the redirect in order to support anonymous downloads when the options are enabled
+        {
+            "action": "*",
+            "principal": "anonymous",
+            "effect": "allow",
         }
+    ],
+
+    "AppRootViewSet": [
+        {
+            "action": ["retrieve"],
+            "principal": "authenticated",
+            "effect": "allow",
+        },
+        {
+            "action": ["retrieve"],
+            "principal": "anonymous",
+            "effect": "allow",
+            "condition": "unauthenticated_collection_download_enabled",
+        },
     ],
 
     'NamespaceViewSet': [
--- a/galaxy_ng/app/api/views.py
+++ b/galaxy_ng/app/api/views.py
@@ -7,10 +7,10 @@ from django.http import HttpResponseRedirect
 from django.shortcuts import get_object_or_404
 from pulp_ansible.app.models import AnsibleDistribution
 
-from rest_framework.permissions import IsAuthenticated
 from rest_framework.response import Response
 from rest_framework.reverse import reverse
 
+from galaxy_ng.app.access_control import access_policy
 from galaxy_ng.app.api import base as api_base
 
 
@@ -31,7 +31,8 @@ VERSIONS = {
 
 
 class ApiRootView(api_base.APIView):
-    permission_classes = [IsAuthenticated]
+    permission_classes = [access_policy.AppRootAccessPolicy]
+    action = "retrieve"
 
     def get(self, request, *args, **kwargs):
         """
@@ -50,7 +51,8 @@ class ApiRootView(api_base.APIView):
 
 
 class ApiRedirectView(api_base.APIView):
-    permission_classes = [IsAuthenticated]
+    permission_classes = [access_policy.AppRootAccessPolicy]
+    action = "retrieve"
 
     """Redirect requests to /api/automation-hub/api/ to /api/automation-hub/
 
--- a/CHANGES/1386.bugfix
+++ b/CHANGES/1386.bugfix
@@ -1,0 +1,1 @@
+Fixes forbidden message when installing from ansible-galaxy a public collection and the settings has enable unautheticated download.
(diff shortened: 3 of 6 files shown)


brumik:
@newswangerd I am pretty sure this is a bad approach. How should I go around this issue? I was thinking duplicating the view and one for the `content/` endpoint would be with the conditional, the other would remain the same, but that's again a copy-paste issue which I don't like either. I have no experience yet in django access policy.

galaxy_ng/app/api/views.py
```diff
@@ -25,7 +26,10 @@
 
 
 class ApiRootView(api_base.APIView):
-    permission_classes = [IsAuthenticated]
+    if settings.GALAXY_ENABLE_UNAUTHENTICATED_COLLECTION_DOWNLOAD:
+        permission_classes = []
+    else:
+        permission_classes = [IsAuthenticated]
```
rochacbruno, line 32: This cannot be done at the class level, because this `if` will be evaluated at the class definition time and it is better for permissions like this to be evaluated at runtime, so it is also easier for testing.

You have to create a new Access Policy `ApiRootAccessPolicy` and inherit and/or implement the same idea of this other one here https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/access_control/access_policy.py#L75-L76

You can see the mixin here https://github.com/ansible/galaxy_ng/blob/6ce0f25b4fd5b93d778099dfe23bbf3f9f26d44d/galaxy_ng/app/access_control/mixins.py#L41

then on the permissions you can set like this https://github.com/ansible/galaxy_ng/blob/6ce0f25b4fd5b93d778099dfe23bbf3f9f26d44d/galaxy_ng/app/access_control/statements/standalone.py#L12

@bmclaughlin implemented all of this and might be able to give you some hints.

rochacbruno (changes requested):

galaxy_ng/app/api/views.py
bmclaughlin: This is the approach I'd go with as well.

brumik: I tried to implement this, but I am wondering, is this okey for the other AppRootViewSet using endpoints too? I feel like I should create a new ViewSet just for the `content/` and `content/.../api` endpoint (and redirect endpoint)

galaxy_ng/app/api/views.py
```diff
@@ -18,14 +18,15 @@
...
+    "pulp_core_version": apps.get_app_config("core").version,
+    "pulp_ansible_version": apps.get_app_config("ansible").version,
+    "pulp_container_version": apps.get_app_config("container").version,
 }
 
 
 class ApiRootView(api_base.APIView):
-    permission_classes = [IsAuthenticated]
+    permission_classes = [access_policy.AppRootAccessPolicy]
+    action = "list"
```
newswangerd, line 29: This should be `retrieve`

galaxy_ng/app/api/views.py
```diff
@@ -34,17 +35,15 @@ def get(self, request, *args, **kwargs):
...
+            distro = get_object_or_404(AnsibleDistribution, base_path=self.kwargs["path"])
             data["distro_base_path"] = distro.base_path
 
         return Response(data)
 
 
 class ApiRedirectView(api_base.APIView):
-    permission_classes = [IsAuthenticated]
+    permission_classes = [access_policy.AppRootAccessPolicy]
+    action = "list"
```
newswangerd, line 46: Same here

galaxy_ng/app/access_control/statements/standalone.py
```diff
@@ -1,4 +1,17 @@
 STANDALONE_STATEMENTS = {
+    'AppRootViewSet': [
+        {
+            "action": ["list", "retrieve"],
```
newswangerd, line 4: if this is only being used on the api root view, then you can set this to `"action": ["retrieve"]`

newswangerd (changes requested):
These really need tests, but we don't have a way to write integration or functional where we can change application settings, so I'm not totally sure how to test this. You may be able to get away with unit tests, but there isn't a way to upload collections or run the `ansible-galaxy` cli there. Perhaps @ironfroggy can help

brumik:
@ironfroggy I seen tests with ansible-galaxy cli (`tests/integration/cli/test_cli_flow`), though I am not sure how to change settings and how to make sure the user is **not** logged in.

bmclaughlin:
@brumik, I ran into the same issue when originally trying to test unauthenticated access, I ended up using unit tests:
https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/tests/unit/api/test_view_only_access.py

rochacbruno (changes requested):
this needs rebase and changes

newswangerd (changes requested):
CI is still failing.

rochacbruno (changes requested):
I think it needs a rebase?

awcrosby:
The integration tests got fixes yesterday for intermittent failures, if you rebase now `ci.int.devshift.net` and `Standalone / integration` will likely pass.

rochacbruno (changes requested):
CI still failing, I guess it needs a rebase

PR #1199, opened 2022-04-05
Migrate permissions from Groups to Roles

bmclaughlin:
# Description 🛠
Migrates Permissions from Groups to a custom Role which is assigned to the Group.  If a Galaxy locked Role can be applied, the Permissions that make up that Role are filtered out prior to creating the custom Role.

Script used to populate my database for testing:
https://gist.github.com/bmclaughlin/3a9d61d7310a285dd00627bfdff4ee69

Issue:  AAH-1128

# Reviewer Checklists  👀
Developer reviewer:


galaxy_ng/app/migrations/0026_move_perms_to_roles.py
```diff
@@ -0,0 +1,136 @@
...
+        ("ansible", "delete_collection"),
+    ), "galaxy.collection_admin"),
+    ((
+        ("container", "delete_containerrepository"),
+        ("container", "namespace_change_containerdistribution"),
+        ("container", "namespace_modify_content_containerpushrepository"),
+        ("container", "namespace_push_containerdistribution"),
+        ("container", "add_containernamespace"),
+        ("container", "change_containernamespace"),
+    ), "execution_environment_admin"),
```
newswangerd, line 19: should be `galaxy.execution_environment_admin`

galaxy_ng/app/migrations/0026_move_perms_to_roles.py
```diff
@@ -0,0 +1,136 @@
...
+
+def move_permissions_to_roles(apps, schema_editor):
+    Group = apps.get_model("galaxy", "Group")
+    GroupRole = apps.get_model("core", "GroupRole")
+    Permission = apps.get_model("auth", "Permission")
+    Role = apps.get_model("core", "Role")
+
+    group_roles = []
+
+    for i, group in enumerate(Group.objects.all()):
```
mdellweg, line 82:
```suggestion
    for group in Group.objects.all():
```
I don't see `i` being used (may be blind...).

galaxy_ng/app/migrations/0026_move_perms_to_roles.py
```diff
@@ -0,0 +1,136 @@
...
+        permissions = group.permissions.all()
+        perms_to_remove = []
+
+        # Use Galaxy locked Roles where possible
+        for locked_perm_names, locked_rolename in GALAXY_LOCKED_ROLE_TRANSLATOR:
+            locked_perms = [Permission.objects.filter(
+                content_type__app_label=app_label, codename=codename
+            ).first() for app_label, codename in locked_perm_names]
+
+            if all(locked_perms):
```
mdellweg, line 94: This looks good to me. Reasoning (maybe add as a comment). Maybe any of these permissions did not even exist by the time the migration runs. But at that time we can safely assume no group to have all of them (because they would have needed to exist).

galaxy_ng/app/migrations/0026_move_perms_to_roles.py
```diff
@@ -0,0 +1,136 @@
...
+                    else:
+                        break
+                # remove permissions covered by locked_roles from the groups permission list
+                permissions = remove_locked_perms_from_permissions(perms_to_remove, permissions)
+
+        # Create custom role for current Group with permissions not in Galaxy locked roles
+        if len(permissions) > 0:
+            role, _ = Role.objects.get_or_create(name=f"{group.name}_role")
+            for permission in permissions:
+                role.permissions.add(permission)
```
mdellweg, line 120: Maybe this can work:
```suggestion
            role.permissions.set(permissions)
```

galaxy_ng/app/migrations/0026_move_perms_to_roles.py
```diff
@@ -0,0 +1,191 @@
...
+        ("galaxy", "change_synclist"),
+        ("galaxy", "delete_synclist"),
+        ("galaxy", "view_synclist"),
+        ("ansible", "change_collectionremote"),
+    ), "galaxy.synclist_owner"),
+]
+
+GROUP_OBJECT_PERMISSION_TRANSLATOR = [
+    ("galaxy", "upload_to_namespace"),
+]
```
mdellweg, line 58: What is this translating to? Is the role to be assigned missing?

galaxy_ng/app/migrations/0026_move_perms_to_roles.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+def move_permissions_to_roles(apps, schema_editor):
+    Group = apps.get_model("galaxy", "Group")
+    GroupRole = apps.get_model("core", "GroupRole")
+    UserRole = apps.get_model("core", "UserRole")
+    Permission = apps.get_model("auth", "Permission")
+    Role = apps.get_model("core", "Role")
+    UserObjectPermission = apps.get_model("guardian", "UserObjectPermission")
+    GroupObjectPermission = apps.get_model("guardian", "GroupObjectPermission")
```
mdellweg, line 90: This reminds me that we need to come up with an exit strategy eventually when we do not even have guardian installed anymore. Maybe we can just assume that there is nothing to do at that point as it must have been a fresh install.

galaxy_ng/app/migrations/0026_move_perms_to_roles.py
```diff
@@ -0,0 +1,191 @@
...
+                                name=locked_rolename,
+                                defaults={"locked": True}
+                            )
+                            group_roles.append(GroupRole(group=group, role=locked_role))
+
+                            # add permissions from locked_perms to list to be removed
+                            perms_to_remove.extend(locked_perms)
+                        continue
+                    else:
+                        break
```
mdellweg, line 125:
```suggestion
                if all((locked_perm in permissions for locked_perm in locked_perms)):
                    # add current locked_role to list to add to current group
                    locked_role, _ = Role.objects.get_or_create(
                        name=locked_rolename,
                        defaults={"locked": True}
                    )
                    group_roles.append(GroupRole(group=group, role=locked_role))
                        
                    # add permissions from locked_perms to list to be removed
                    perms_to_remove.extend(locked_perms)
```

mdellweg: I'm a fan of list comprehensions. ;) But really i think it's more readable.

galaxy_ng/app/migrations/0026_move_perms_to_roles.py
```diff
@@ -0,0 +1,191 @@
...
+        if len(permissions) > 0:
+            role, _ = Role.objects.get_or_create(name=f"{group.name}_role")
+            role.permissions.set(permissions)
+            group_roles.append(GroupRole(group=group, role=role))
+
+    # Group Object Permissions
+    for gop in GroupObjectPermission.objects.all():
+        # create GroupRole objects, assign to group from gop
+        role, _ = Role.objects.get_or_create(name=f"{gop.group.name}_role")
+        role.permissions.add(gop.permission)
```
mdellweg, line 145: That sounds dangerous. You may hit the same group you created before without or for another object, so now you are adding the new permissions to that role making it effective for that other object or even globally.
I think you need to extend the role name with the objects pk to make it unique.

newswangerd (changes requested):
This still needs to attempt to assign locked roles when creating object permissions.

galaxy_ng/app/migrations/0026_move_perms_to_roles.py
```diff
@@ -0,0 +1,205 @@
...
+        for locked_perm_names, locked_rolename in MODEL_PERMISSION_TRANSLATOR:
+            locked_perms = [Permission.objects.filter(
+                content_type__app_label=app_label, codename=codename
+            ).first() for app_label, codename in locked_perm_names]
+
+            if all(locked_perms):
+                # compare locked role perms to perms
+                if all((locked_perm in permissions for locked_perm in locked_perms)):
+                    # add current locked_role to list to add to current group
+                    locked_role, _ = Role.objects.get_or_create(
```
newswangerd, line 92: all new group names need to be prefixed with `galaxy.`

mdellweg:
Should this PR change it's base branch now?

galaxy_ng/app/migrations/0028_move_perms_to_roles.py
```diff
@@ -0,0 +1,286 @@
...
+    ), "galaxy.synclist_owner"),
+]
+
+
+def batch_process_group_roles(apps, group_roles):
+    GroupRole = apps.get_model("core", "GroupRole")
+    if len(group_roles) > 1000:
+        # print('group_roles > 1000')
+        GroupRole.objects.bulk_create(group_roles)
+        group_roles.clear()
```
mdellweg, line 62: I'm not sure how expensive the `GroupRole = apps.get_model("core", "GroupRole")` call is. How about the following:

```suggestion
def batch_process(model, objects, flush=False):
    if len(objects) > 1000 or flush:
        # print(f'#{model.__name__} = {len(objects)}')
        model.objects.bulk_create(objects)
        objects.clear()
```

And you would call it with `batch_process(GroupRole, group_roles)` in the loop and once again with `flush=True` at the end. As far as i see, you are currently missing the last unfinished batch.

galaxy_ng/app/signals/handlers.py
```diff
@@ -56,12 +56,8 @@ def set_role_definitions(sender, **kwargs):
...
         from django.apps import apps
     role_prefix = f"{sender.label}."
-    desired_roles = {}
-    for viewset in GALAXY_VIEWSETS:
-        locked_roles = GALAXY_VIEWSETS[viewset]['LOCKED_ROLES']
-        if locked_roles is not None:
-            desired_roles.update(locked_roles or {})
-    adjust_roles(apps, role_prefix, desired_roles, verbosity=kwargs.get("verbosity", 1))
+    if GALAXY_LOCKED_ROLES is not None:
+        adjust_roles(apps, role_prefix, GALAXY_LOCKED_ROLES, verbosity=kwargs.get("verbosity", 1))
```
mdellweg, line 60: Hmm, this might be a problem. `adjust_roles` is already called for every plugin by pulpcore. And it will enforce whatever it finds on the plugins viewsets (including deleting all roles if nothing was found).
I think, you can either define the locked roles on viewsets and just completely remove this hook, or use a different prefix.
A third option may be to adjust the plugins `AppConfig.ready`.

newswangerd (changes requested):
I stress tested the system with 500 containers and 500 collection namespaces and this generated 7000 new roles, which is unmangeable. It looks like none of the locked roles are getting applied to object permissions because we didn't define any locked roles with same permission sets that the UI exposes.

I'll get back to you with a new set of locked roles, as well as some other possible strategies we can employ for object permissions.

galaxy_ng/tests/unit/migrations/test_0028_move_perms_to_roles.py
```diff
@@ -0,0 +1,199 @@
...
+from pulp_ansible.app.models import Collection
+
+from pulpcore.app.models.role import GroupRole
+from pulpcore.app.models.role import UserRole
+from pulpcore.app.models.role import Role
+
+from django.contrib.auth.models import Permission
+from guardian.models import GroupObjectPermission
+from guardian.models import UserObjectPermission
+from guardian.shortcuts import assign_perm as guardian_assign_perm
```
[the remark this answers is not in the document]
newswangerd: I think we can leave this test here and possibly updated it or remove it when we switch to pulpcore 3.20

galaxy_ng/app/migrations/0028_move_perms_to_roles.py
```diff
@@ -0,0 +1,352 @@
...
+            ))
+
+    return group_roles
+
+
+def add_object_role_for_users_with_permission(role, permission, UserRole, ContentType, User):
+    user_roles = []
+
+    # Use raw sql because guardian won't be available
+    cursor = connection.cursor()
```
ipanova, line 239: You want to close the cursor as soon as you're done with it

galaxy_ng/app/access_control/statements/roles.py
```diff
@@ -1,67 +1,112 @@
...
+LOCKED_ROLES = {
+    "galaxy.collection_admin": {
+        "permissions": [
+            "galaxy.add_namespace",
+            "galaxy.change_namespace",
+            "galaxy.delete_namespace",
+            "galaxy.upload_to_namespace",
+            "ansible.delete_collection",
+            "ansible.change_collectionremote",
+            "ansible.view_collectionremote",
```
awcrosby, line 10: Today the `system:partner-engineering` group has these perms https://github.com/ansible/galaxy_ng/blob/b6eb764d54c6e7a57bb391d77aee4da06e339b5a/galaxy_ng/app/management/commands/maintain-pe-group.py#L27

The updated management command in this feature branch keeps those perms via these roles https://github.com/ansible/galaxy_ng/blob/6bead6e86732cdff7bd8329003bec4a3c6916707/galaxy_ng/app/management/commands/maintain-pe-group.py#L28-L33

This PR will give this group `ansible.change_collectionremote` and `ansible.view_collectionremote` which is probably ok since those viewsets would be unavailable.

galaxy_ng/app/access_control/statements/roles.py
```diff
@@ -1,67 +1,112 @@
...
+            "ansible.modify_ansible_repo_content",
```
awcrosby, line 11: This PR also adds `ansible.modify_ansible_repo_content` to `galaxy.collection_admin`. I assume we still want the role `galaxy.content_admin` (but it wouldnt be needed for CRC). We can remove `galaxy.content_admin` from `maintain-pe-group.py` or keep it there as it just overlaps a perm with `galaxy.collection_admin`

I also see we could remove `galaxy.namespace_owner` from `maintain-pe-group.py` since it overlaps perms with `galaxy.collection_admin`.

Let me know if you want me to do a PR to remove these on this branch.

newswangerd (approved):

galaxy_ng/app/access_control/statements/roles.py
newswangerd: `galaxy.content_admin` seems like it should include EEs and collections maybe. @awcrosby what do you think?

awcrosby: I'm trying to think where to draw the line between `galaxy.collection_admin` and `galaxy.content_admin`. Would it be more clear to move the 4 namespace perms (add/change/delete/upload) to a namespace_admin role?

Then if kinda feels like we have 2 roles left (maybe could be renamed?):
* Role `galaxy.collection_delete`
  * `ansible.delete_collection`
* Role `galaxy.collection_edit`
  * `ansible.modify_ansible_repo_content` - Move a collection already in the system across repos
  * `ansible.view|change_collectionremote` - Bring in collections from remotes

Do we want to mix collections and non-collection perms in the locked roles?, im not sure we know enough yet about how they might be used to create a good mix right now.

galaxy_ng/app/access_control/statements/roles.py
```diff
@@ -1,67 +1,110 @@
...
+    "galaxy.content_admin": {
+        "permissions": [
+            "ansible.modify_ansible_repo_content",
+        ],
+        "description": "Approve and reject collections."
+    },
+    "galaxy.namespace_owner": {
+        "permissions": [
+            "galaxy.change_namespace",
+            "galaxy.upload_to_namespace",
```
awcrosby, line 64: This seems like a good change to limit the namespace owner to uploads and editing the namespace object (not adding/delete namespaces nor deleting collections).

What is the clearest way to see that the role `galaxy.namespace_owner` and it's permissions will apply to a namespace object, and not to a namespace model (this role previously had `galaxy.add_namespace` permission which only makes sense at the model level)

newswangerd: @awcrosby there isn't a distinction between object level roles and model level roles. Both can be applied in either setting, which sort of complicates a lot of things. This one is intended to be used for objects, but there's not really a way to denote that.

awcrosby: So when this migration looks at an existing group and its permissions, what happens when a group has a mix of object and group level permissions... I like the docstring example in `migrate_group_permissions_to_roles` maybe it could get a couple more examples?

    Example:
    If a group has permissions:
        - galaxy.change_namespace (at object level)
        - galaxy.upload_to_namespace (at object level)
        - galaxy.delete_collection (at model level)
        - galaxy.view_group (at model level)
        - galaxy.view_user (at object level)
        
    The following roles would get applied:
        - galaxy.namespace_owner (does it have the same perms each at its previous object/model level?)
        - _permission:galaxy.view_group (at model level?)
        - _permission:galaxy.view_user (at object level?)

newswangerd: `OBJECT_PERMISSION_TRANSLATOR` gets applied to object perms and `GLOBAL_PERMISSION_TRANSLATOR` gets applied to model perms

newswangerd: For future reference: As long as a role has any permissions for an object you can assign it to that object. The extra permissions are ignored. So if I have a role with galaxy.change_namespace and galaxy.view_user and apply it to a collection namespace, view_user won't do anything.

galaxy_ng/app/access_control/statements/roles.py
```diff
@@ -1,67 +1,153 @@
...
-            "galaxy.collection_admin": [
-                "galaxy.change_namespace",
-                "galaxy.delete_namespace",
-                "galaxy.view_namespace",
-                "galaxy.upload_to_namespace",
-                "ansible.delete_collection",
-            ],
-        }
+LOCKED_ROLES = {
+    "galaxy.content_admin": {
```
awcrosby, line 2: @newswangerd did you intent to add a `signature_admin` with `can_sign_collections` perm that also exists in `content_admin`?

galaxy_ng/app/management/commands/maintain-pe-group.py
```diff
@@ -29,7 +29,7 @@ def handle(self, *args, **options):
             'galaxy.group_admin',
             'galaxy.user_admin',
             'galaxy.collection_admin',
-            'galaxy.namespace_owner',
+            'galaxy.collection_namespace_owner',
             'galaxy.content_admin',
```
awcrosby, line 33: `galaxy.collection_namespace_owner` can be removed, it's perms are covered in `galaxy.collection_admin`

`galaxy.content_admin` can be removed, it has perms that currently don't apply to the users in this group.

awcrosby: if `signature_admin` gets added as a locked role, that should also be added here

galaxy_ng/app/migrations/0028_move_perms_to_roles.py
```diff
@@ -0,0 +1,399 @@
...
+    ("galaxy", "upload_to_namespace"),
+    ), "galaxy.collection_publisher"),
+    ((
+        ("ansible", "change_collectionremote"),
+        ("ansible", "view_collectionremote"),
+        ("ansible", "modify_ansible_repo_content"),
+    ), "galaxy.collection_curator"),
+    ((
+        ("galaxy", "change_namespace"),
+        ("galaxy", "upload_to_namespace"),
```
awcrosby, line 75: `galaxy.collection_curator` in LOCKED_ROLES is different

galaxy_ng/app/migrations/0028_move_perms_to_roles.py
```diff
@@ -0,0 +1,399 @@
...
+        # ("container", "namespace_add_containerdistribution"),
+    ), "galaxy.execution_environment_namespace_owner"),
+    ((
+        ("container", "namespace_push_containerdistribution"),
+        ("container", "namespace_change_containerdistribution"),
+        ("container", "namespace_modify_content_containerpushrepository"),
+    ), "galaxy.execution_environment_collaborator"),
+    ((
+        ("ansible", "modify_ansible_repo_content"),
+    ), "galaxy.content_admin"),
```
awcrosby, line 114: should this be removed? `galaxy.content_admin` is already in `GLOBAL_PERMISSION_TRANSLATOR`

newswangerd: oh, good catch

awcrosby (approved):

PR #1205, opened 2022-04-11
Revert RBAC Roles

newswangerd:

PR #1212, opened 2022-04-13
Reinstate RBAC Roles into 4.6

newswangerd:
Reverts ansible/galaxy_ng#1205

### DIFF
--- a/CHANGES/1092.misc
+++ b/CHANGES/1092.misc
@@ -1,1 +1,1 @@
-Update to pulpcore 3.17.3
+Create galaxy_ng specific Roles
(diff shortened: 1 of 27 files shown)


awcrosby (commented):
Since multiple PRs will be needed to make RBAC work, I recommend having a feature branch on `galaxy_ng` for rbac roles. PRs can be raised and merged there, and when the feature branch is complete, that can be merged into `master`.

We want those PRs to kick off the same CI, including the ephemeral environment ci.int.devshift.net pr_check.sh using images on quay. If that is missing on the feature branch PR, we can look at config to try to resolve it.

awcrosby (approved):

PR #1222, opened 2022-04-21
Add makefile command to run integration tests in a container

newswangerd:
Adds `make docker/test-integration` command that runs the integration tests. Pytest flags can be passed to the command with `make docker/test-integration FLAGS="-k sometest"`

No-Issue


Makefile
```diff
@@ -82,6 +82,12 @@ docker/build:     ## Build all development images.
 docker/test:      ## Run unit tests.
 	$(call exec_or_run, api, $(DJ_MANAGER), test, galaxy_ng.tests.unit$(TEST))
 
+# e.g: make docker/test FLAGS="my pytest flags"
+.PHONY: docker/test-integration
+docker/test-integration:      ## Run integration tests.
+	docker build . -f dev/standalone/integration-test-dockerfile -t galaxy-integration-runner
```
rochacbruno, line 88: do we need to rebuilt this every time the tests run?

jctanner (approved):

rochacbruno (approved):

Makefile
newswangerd: @rochacbruno yeah, the files aren't mounted into the container. As long as docker is caching things properly, this should be pretty fast. We don't need to do it this way. We could mount the `galaxy` folder and provide a `make build/integration-container` command to build the image and a `make docker/test-integration` command to run it, but this seems like it works well enough.

brumik:
When running it I got an error that I need to use buildkit to modify permissions (chmod).

```
Step 8/10 : COPY --chmod=777 dev/common/RUN_INTEGRATION.sh /home/RUN_INTEGRATION.sh
the --chmod option requires BuildKit. Refer to https://docs.docker.com/go/buildkit/ to learn how to build images with BuildKit enabled
Unable to find image 'galaxy-integration-runner:latest' locally
docker: Error response from daemon: pull access denied for galaxy-integration-runner, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.
```

It seems like we can add it easily to the build file or as an env varable?

newswangerd:
Closing this because the integration tests require docker, which makes running these tests in a container a lot more complicated.

jctanner (commented):
which virtualenv || pip install --user virtualenv

brumik (approved):

dev/standalone/integration-test-dockerfile
```diff
@@ -0,0 +1,19 @@
...
+
+WORKDIR /app/
+
+# Install test requirements first to make running the tests faster
+SHELL ["/bin/bash", "-c"] 
+COPY integration_requirements.txt /app/integration_requirements.txt
+RUN pip install virtualenv && \
+    virtualenv /tmp/gng_testing && \
+    source /tmp/gng_testing/bin/activate &&\
+    pip install -r integration_requirements.txt
```
rochacbruno, line 11: This can be simplified

1. `venv` is built-in in Python 3+
2. You can run `bin/python` directly without the need to activate the venv

```suggestion
RUN python -m venv /tmp/gng_testing && \
    /tmp/gng_testing/bin/python -m pip install -r integration_requirements.txt
```

rochacbruno, line 11: Thinking a bit more about it, you don't need a virtualenv inside a docker container, the container is already an isolated env.

```suggestion
RUN pip install -r integration_requirements.txt
```

awcrosby:
I think the intermittent CI failures are fixed now, it looks to be failing `ci.int.devshift.net` only because of merge conflict

dev/standalone/integration-test-dockerfile
newswangerd: the test script creates a virtual env, so that's why this is here.

PR #1236, opened 2022-04-28
It's time for TDD!!!

jctanner:
# Description 🛠
It's time for TDD!

### DIFF
--- a/.github/workflows/ci_tdd.yml
+++ b/.github/workflows/ci_tdd.yml
@@ -22,6 +22,7 @@ jobs:
       - name: Ensure changes were made to tests
         env:
           PY_COLORS: '1'
+          GITHUB_BASE_REF: ${{ github.base_ref }}
           GITHUB_BRANCH: ${{ github.head_ref }}
           GITHUB_USER: ${{ github.event.pull_request.user.login }}
           GITHUB_CONTEXT: ${{ toJson(github) }}
--- a/.github/workflows/ci_tdd.yml
+++ b/.github/workflows/ci_tdd.yml
@@ -17,7 +17,7 @@ jobs:
           python-version: "3.8"
 
       - name: make the checkout in a sane way.
-        run: git clone https://github.com/${{ github.event.pull_request.user.login }}/galaxy_ng; cd galaxy_ng ; git checkout ${{ github.head_ref }}
+        run: git clone https://github.com/${{ github.event.pull_request.user.login }}/galaxy_ng; cd galaxy_ng ; git checkout "${{ github.head_ref }}"
 
       - name: Ensure changes were made to tests
         env:
--- a/.github/workflows/ci_tdd.yml
+++ b/.github/workflows/ci_tdd.yml
@@ -1,0 +1,28 @@
+---
+name: TDD
+on: 
+  pull_request:
+    branches:
+      - '*'
+  workflow_dispatch:
+
+jobs:
+
+  validation:
+    runs-on: ubuntu-latest
+    steps:
+
+      - uses: actions/setup-python@v2
+        with:
+          python-version: "3.8"
+
+      - name: make the checkout in a sane way.
+        run: git clone https://github.com/${{ github.event.pull_request.user.login }}/galaxy_ng; cd galaxy_ng ; git checkout ${{ github.head_ref }}
+
+      - name: Ensure changes were made to tests
+        env:
+          PY_COLORS: '1'
+          GITHUB_BRANCH: ${{ github.head_ref }}
+          GITHUB_USER: ${{ github.event.pull_request.user.login }}
+          GITHUB_CONTEXT: ${{ toJson(github) }}
+        run: cd galaxy_ng; python dev/common/tdd.py
--- a/dev/common/tdd.py
+++ b/dev/common/tdd.py
@@ -6,11 +6,14 @@ import os
 import subprocess
 import sys
 
-from pprint import pprint
-
 
 def get_current_branch():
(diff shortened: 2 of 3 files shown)


jctanner:
image

dev/common/tdd.py
```diff
@@ -0,0 +1,82 @@
...
+    target_branch = os.environ.get("GITHUB_BASE_REF", "master")
+    print(f'TARGET-BRANCH: {target_branch}')
+    if target_branch != 'master':
+        print('TDD is only enforced on the master branch')
+        sys.exit(0)
+
+    changed_files = get_changed_files(pr_branch, target_branch=target_branch)
+    for cf in changed_files:
+        print(f'modified file: {cf}')
+    verify_test_files_changed(changed_files)
```
awcrosby, line 78: We might want a easily edited set of exceptions. We could only call `verify_test_files_changed(changed_files)` if changed_files are in galaxy_ng/ (that would skip for edits to .ci/ .github/ and template_config.yml, requirements/, setup.py, pr_check.sh)

Not sure if want or need any exceptions within galaxy_ng/?

jctanner: Makes sense. I'll add that.

dev/common/tdd.py
```diff
@@ -0,0 +1,111 @@
...
+    for cf in changed_files:
+        for tp in test_paths:
+            if cf.startswith(tp):
+                tests_found = True
+                break
+        if tests_found:
+            break
+
+    if not tests_found:
+        raise Exception('Tests should be added or modified with -every- PR!!!')
```
himdel, line 74: This bit may be overly ambitious :)

Should the policy be that we ignore this check for those kinds of PRs or should we enable this only for ..feature PRs? (Like, only run if `CHANGES/*.feature` is being updated as well.)

(Looking at the last 16 merged PRs, 12 of those should not be touching tests.. CI changes, dev env changes, ephemeral, dependency changes (updating pulp), release PRs .. there's probably more?)

.github/workflows/ci_tdd.yml
```diff
@@ -0,0 +1,29 @@
...
+  validation:
+    runs-on: ubuntu-latest
+    steps:
+
+      - uses: actions/setup-python@v2
+        with:
+          python-version: "3.8"
+
+      - name: make the checkout in a sane way.
+        run: git clone https://github.com/${{ github.event.pull_request.user.login }}/galaxy_ng; cd galaxy_ng ; git checkout ${{ github.head_ref }}
```
himdel, line 20:
```suggestion
        run: git clone https://github.com/${{ github.event.pull_request.user.login }}/galaxy_ng; cd galaxy_ng ; git checkout '${{ github.head_ref }}'
```

(branch names can contain parentheses)

dev/common/tdd.py
jctanner: The code is already filtering out any changes to files outside the application source path.

If tests aren't written for bugfixes, you will inevitably have to fix the same bug again. There should be no excuse for not writing tests for every PR that changes the application's behavior.

https://evolveum.com/test-driven-bugfixing/#:~:text=You%20write%20an%20(automated)%20test,that%20the%20test%20is%20passing.

jctanner:
@newswangerd it requires some change, any change, no matter how big or small to be made to the galaxy_ng/tests directory -if- the PR changes anything in the galaxy_ng directory.

The check is skipped if the target branch is not master or from dependabot/backportbot.

newswangerd:
It looks like this boils down to:

```
if code is changed:
    verify that tests are changed
```

Is that right? It doesn't do anything fancy like checking code coverage?

newswangerd (approved):
I'm really happy to see that we're doing this. More tests = better sleep at night for me.

jctanner:
@newswangerd correct, it doesn't actually care what the changes are. Implementing coverage with our setup seems difficult, so I can't really make something that insists on increasing coverage [yet].

hendersonreed:
I see this as a significant improvement of the project workflow more generally - having developers produce most of the tests means that we'll be able to pretty drastically streamline the release process I think (and the testing effort will be more symmetric with the development effort - as it is QE is heavily outnumbered and can't keep up with release activities AND writing tests for all new features.) 

Putting more testing responsibilities on devs means we need to ensure that tests are written and contributed regularly, and this is the simplest way to approach that.

Maybe it's not the ideal solution, but I think it's worth doing *something* now and refining it down the line, versus waiting for a better process to do *anything*.

dev/common/tdd.py
ironfroggy: I agree with @jctanner about 90% and the other 10% is covered by @mdellweg there. Features and bugfixes? Absolutely should get new or updated tests. Refactoring I'd almost say SHOULDN'T have test changes, almost by definition of "refactoring". As long as people are responsible about it, maybe we could add a flag you can put in PR or commit descriptions to mark it as refactoring and accepting the existing test state.

jctanner: Without coverage reports or any sense of what functions get tested, i'd argue that a refactor absolutely needs to add tests to -prove- that the author actually took the time to understand and learn what the code was meant to do.

jctanner:
To be discussed at 05/02/2022 team meeting.

himdel:
From the 4.5.0 release, these should probably be added to exceptions?

* `galaxy_ng/app/settings.py` - needed to adjust the list of supported LANGUAGES
* `galaxy_ng/__init__.py` - version
* `galaxy_ng/app/__init__.py` - version

PR #1246, opened 2022-05-03
dev/: use UID, GID instead of hardcoding 1000; drop USER_ID

himdel:
the problem: running as a user with uid=1001 and owning src/,
the dev environment doesn't work, failing with

    error: could not create 'galaxy_ng.egg-info': Permission denied

or

    running egg_info
    error: [Errno 13] Permission denied

because the docker files assume uid=1000 and gid=1000 for the galaxy user inside the container.

Changing to use the current user's uid & gid.

### DIFF
--- a/dev/Dockerfile.base
+++ b/dev/Dockerfile.base
@@ -1,6 +1,7 @@
 FROM registry.access.redhat.com/ubi8
 
-ARG USER_ID=1000
+ARG UID=1000
+ARG GID=1000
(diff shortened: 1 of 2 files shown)


himdel:
turns out UID and GID are not set by default in ..some environments?
(and no idea why Standalone / integration runs the dev environment?)

newswangerd:
Closing stale PR

PR #1247, opened 2022-05-03
Serve the pulp api at `/api/automation-hub/pulp/api/v3/`

awcrosby:
# Description 🛠
Serve the pulp api at `/api/automation-hub/pulp/api/v3/`
Issue: [AAH-1544](https://issues.redhat.com/browse/AAH-1544)

# Reviewer Checklists  👀
Developer reviewer:


awcrosby:
@rochacbruno Hmm, what is the criteria for adding a config option to docs... That the config option is one the users might likely change? Or describing the settings in general?

I'm not sure the user would edit this or not, but I can add it just in case.

.github/workflows/scripts/install_python_client.sh
```diff
@@ -16,7 +16,7 @@ cd "$(dirname "$(realpath -e "$0")")"/../../..
 
 pip install twine wheel
 
-export REPORTED_VERSION=$(http $PULP_URL/pulp/api/v3/status/ | jq --arg plugin galaxy --arg legacy_plugin galaxy_ng -r '.versions[] | select(.component == $plugin or .component == $legacy_plugin) | .version')
+export REPORTED_VERSION=$(http {$PULP_URL}{$PULP_API_ROOT}api/v3/status/ | jq --arg plugin galaxy --arg legacy_plugin galaxy_ng -r '.versions[] | select(.component == $plugin or .component == $legacy_plugin) | .version')
```
awcrosby, line 19: I am trying to resolve CI issues with 404s on `/pulp/api/v3/status/`. @fao89 do we need to update the calls to the status api in these search results to include `settings.API_ROOT`? https://github.com/pulp/plugin_template/search?q=%22api%2Fv3%2Fstatus%22&type=

fao89: the plugin-template already brings the necessary changes
https://github.com/pulp/plugin_template/blob/f49381dbd6aa45d2a510235a285a7bc3e5274eff/templates/github/.ci/ansible/start_container.yaml#L78-L90

fao89: FYI, S3 job already runs rerouted:
https://github.com/ansible/galaxy_ng/blob/master/.github/workflows/scripts/install.sh#L126-L127
https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/tests/functional/utils.py#L245-L246

awcrosby: @fao89 I am using https://github.com/pulp/plugin_template/commit/36696fcf9dfe3c5b418ed498c4f97d3168361e5e but in the `Galaxy CI / test (pulp)` job logs we still see
```
2022-05-11T13:07:02.3709391Z http: error: SSLError: HTTPSConnectionPool(host='pulp', port=443): Max retries exceeded with url: /pulp/api/v3/status/ (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self signed certificate (_ssl.c:1131)'))) while doing a GET request to URL: https://pulp/pulp/api/v3/status/
```

fao89: actually plugin-template needs more changes for your use case, I'll start a PR

fao89: you need this: https://github.com/pulp/plugin_template/pull/641

template_config.yml
```diff
@@ -17,16 +17,16 @@ additional_repos:
   name: galaxy-importer
   org: ansible
 aiohttp_fixtures_origin: 172.18.0.1
+api_root: /pulp/
```
fao89, line 20:
```suggestion
api_root: "/api/automation-hub/pulp/"
```

rochacbruno (approved):

fao89 (approved):

dev/insights/galaxy_ng.env
```diff
@@ -1,6 +1,7 @@
 PULP_CONTENT_PATH_PREFIX=/api/automation-hub/v3/artifacts/collections/
 
 PULP_GALAXY_API_PATH_PREFIX=/api/automation-hub/
+PULP_API_ROOT=/api/automation-hub/pulp/
```
newswangerd, line 4: could we find a way to configure this a little more easily? Maybe put it in dynaconf hooks so that it doesn't have to be configured in so many different places? Something like this could work:

``` python
# /galaxy_ng/galaxy_ng/app/dynaconf_hooks.py
def configure_api_base_path(settings: Dynaconf) -> Dict[str, Any]:
    galaxy_base = settings.get("GALAXY_API_PATH_PREFIX", None)

    if galaxy_base:
        return {
            "API_ROOT": f"/{galaxy_base.strip('/')}/pulp/"
        }
    else:
        return {}
```

This won't allow anyone to set `API_ROOT` by itself, but perhaps we could add some logic to only alter the value if it's set to a default value. @rochacbruno do you have any thoughts?

galaxy_ng/app/settings.py
```diff
@@ -30,6 +30,10 @@
 # FIXME(cutwater): 1. Rename GALAXY_API_ROOT in pulp-ansible to ANSIBLE_API_ROOT
 #                  2. Rename API_PATH_PREFIX to GALAXY_API_ROOT
 GALAXY_API_PATH_PREFIX = "/api/galaxy"
+
+# Location where the pulp api will be served
+API_ROOT = "/api/galaxy/pulp/"
```
newswangerd, line 35: I'm a little concerned about pulling the rug out from under the existing `/pulp/` endpoints. We have no idea if on prem customers are using these endpoints or not and it seems like breaking the existing behavior of the API could be a bad idea. It might be a good idea to return a `301 moved permanently` on the old `/pulp/api/v3/**` endpoints or something like that.

newswangerd (changes requested):
I'd like us to have some kind of plan for deprecating/ the `/pulp/api/v3/` endpoints for on prem before we merge this. This could be something along the lines of:
- Communicating the change clearly to customers
- Redirecting the APIs
- Some kind of indication that we're certain that we won't break customers if they are using `/pulp/api/v3/`

galaxy_ng/app/settings.py
awcrosby: Agreed we should add a redirect

dev/insights/galaxy_ng.env
newswangerd: @awcrosby one strategy you could use is create a new setting called something like `CORE_API_ROOT` that lets users set `API_ROOT`. Something like:

```
if core_api_root:
    API_ROOT=core_api_root
else:
    API_ROOT=galaxy_base + "/pulp/"
```

awcrosby: thanks, i will give this a try

awcrosby: @newswangerd PTAL - I went with the more simple approach of not having the pulp API_ROOT be user-configurable in our settings.py

newswangerd (approved):
Looks good. You're gonna look into redirects for the old `/pulp/` endpoints after this is merged, right?

awcrosby:
@newswangerd that is correct, will add redirect with AAH-1646

fao89 (commented):
you need:

```diff

diff --git a/galaxy_ng/openapi/__init__.py b/galaxy_ng/openapi/__init__.py
index 9557b5f2..a8353f3f 100644
--- a/galaxy_ng/openapi/__init__.py
+++ b/galaxy_ng/openapi/__init__.py
@@ -16,7 +16,7 @@ class GalaxySchemaGenerator(PulpSchemaGenerator):
         """Munge pulp's get_schema result"""
         self._input_request = request
         schema = super().get_schema(request=request, public=public)
-        if self._input_request.path.startswith(settings.GALAXY_API_PATH_PREFIX):
+        if self._input_request.path.startswith(settings.GALAXY_API_PATH_PREFIX) and "bindings" not in self._input_request.query_params:
             return self.dedupe_operationIds(schema)
         return schema
         
```

awcrosby:
@fao89 i made a couple other adjustments and got all tests passing except for the pulp ci s3 test, where "Waiting for pulp" 404s on `http://pulp/rerouted/djnd/api/v3/status/` - are there any other template_config params we need to adjust for this?

fao89 (commented):
I believe you need to disable test_reroute:
https://github.com/ansible/galaxy_ng/blob/master/template_config.yml#L98
and reapply plugin-template

As your entire CI is rerouted, this test doesn't make sense, and it is probably the thing that is currently breaking the CI

awcrosby:
@fao89 thank you!

PR #1244, opened 2022-05-03
Localization stable-4.5 string update

rooftopcellist:
# Description 🛠

Localization Update from 3-28 push.  Soon to follow, the translated strings from the 4-19 push.  

# Reviewer Checklists  👀
Developer reviewer:
- [ ] Code looks sound, good architectural decisions, no [code smells](https://www.codegrip.tech/productivity/everything-you-need-to-know-about-code-smells/)
- [ ] There is a Jira issue associated (note that "No-Issue" should be rarely used)
- [ ] Tests are included in `galaxy_ng/tests/integration` or `galaxy_ng/tests/functional`, and they fully cover necessary test scenarios… or tests not needed

PR #1248, opened 2022-05-04
Localization stable-4.5 string update

rooftopcellist:
# Description 🛠

Localization Update from 4-19 push.  Soon to follow, the translated strings from the 4-19 push.

### DIFF
--- a/galaxy_ng/locale/fr/LC_MESSAGES/django.po
+++ b/galaxy_ng/locale/fr/LC_MESSAGES/django.po
@@ -7,22 +7,25 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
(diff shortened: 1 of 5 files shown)


rooftopcellist:
This PR has all of the changes https://github.com/ansible/galaxy_ng/pull/1244 does and more, so the old PR can be closed now.

rooftopcellist:
cc @himdel for review

newswangerd (changes requested):
Can you open this against master instead?

rooftopcellist:
@newswangerd In the future, to catch any string changes between stable-4.5 and master, I will push strings for both master and the stable branch.  Memsource pre-fills translations when new strings are pushed, so these changes will be picked up for the master branch on the next push/pull cycle (and make it into stable-4.6).  

^^ just for context.  I will target the master branch, that way we get these strings merged there earlier.

newswangerd (approved):

himdel:
(maybe close and open will wake github up...)

PR #1261, opened 2022-05-11
Update dev env base image to use ubi8

awcrosby:
Fixes ./compose build

No-Issue

### DIFF
--- a/dev/Dockerfile.base
+++ b/dev/Dockerfile.base
@@ -1,4 +1,4 @@
-FROM centos:8
+FROM registry.access.redhat.com/ubi8
 
 ARG USER_ID=1000
 ARG USER_NAME=galaxy


newswangerd (commented):
@awcrosby is this a good idea? The prod dockerfile uses UBI 8. Shouldn't we keep the dev image compatible with the prod image?

awcrosby:
@newswangerd I opened this because I was getting errors with `./compose build` on `dnf install` from the centos 8 repos… but it seems to be building ok right now.

> The prod dockerfile uses UBI 8

This PR makes the dev dockerfile also use UBI 8.

I didn’t realize we have 2 Dockerfiles in `stable-4.2`. `Dockerfile` uses centos:8, while `Dockerfile.rhel8` uses ubi8. I'm not sure how `stable-4.2` uses these during a release... in subsequent release branches `Dockerfile` symlinks to `Dockerfile.rhel8`

https://github.com/ansible/galaxy_ng/blob/630ebc2b710af1fe99238f3cb92fb3f32a0dee24/Dockerfile#L1-L4

https://github.com/ansible/galaxy_ng/blob/630ebc2b710af1fe99238f3cb92fb3f32a0dee24/Dockerfile.rhel8#L1

newswangerd (approved):

PR #1268, opened 2022-05-16
Removing hardcoded environment variables from dev

brumik:
No-Issue

#### What is this PR doing:
Removing the two hardcoded env variables from the dev setup for standalone so [cypress test](https://github.com/ansible/ansible-hub-ui/pull/2069/files) can pass locally. If needed they can be added back in the `.compose.env`

#### Notes: 

### DIFF
--- a/.compose.env.example
+++ b/.compose.env.example
@@ -3,6 +3,13 @@
 # Set tp standalone-keycloak to enable the Social Auth Login
 COMPOSE_PROFILE=standalone
 
+# The following two variables were initally specified in the galaxy_ng.env file. 
+# When wanna reach the same environment for cypress testing as in CI
+# you should set PULP_GALAXY_REQUIRE_CONTENT_APPROVAL=False in galaxy_ng/app/settings.py
+# otherwise cypress cannot change it. 
+# PULP_GALAXY_REQUIRE_CONTENT_APPROVAL=False
+# PULP_GALAXY_AUTO_SIGN_COLLECTIONS=True
+
 # Add extra paths to run project dependencies from local filesystem in editable mode
 # The order of repositories in the list is important, it is `:` separated.
 # DEV_SOURCE_PATH='pulpcore:pulp_ansible:galaxy_ng'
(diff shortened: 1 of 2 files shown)


brumik:
Yes. Also I am not authored to merge :)

himdel:
Do we also need to change `galaxy_ng/app/setttings.py` so that we don't actually change the defaults for dev env?

brumik:
@himdel I think we can mention it in the docs, since in dev env it is easier and more transparent for the developers if we keep all the variables in `.compose.env`. Or at least the sensible ones. So when trying to run for example cypress tests, all you need to do is comment out all your set variables in `compose.env` and it should pass. 

Hmm, on second thought we have actually one variable `GALAXY_REQUIRE_CONTENT_APPROVAL: false` in the CI, but that is not environment variable so cypress can overwrite it. I am not sure how we could consolidate that. Maybe add to the docs saying that some tests require you to have in you `.compose.env` `GALAXY_REQUIRE_CONTENT_APPROVAL: false` and some does not? (It does not seem like an ideal solution. Ideally we would have some local settings file where we could edit the environment variables without adding it to the `env` and preventing cypress to overwrite it.

**Summary**
Yes, we should add this info to the docs, as it is little more transparent and it add one time few seconds overhead to development. (I would maybe remove the `PULP_GALAXY_COLLECTION_SIGNING_SERVICE=ansible-default` line too).

brumik:
Changing the defaults in `settings.py` I don't think is too viable. Default values are for the end user, not developers, and they are sensible when somebody installs the system. But I am not an expert on this case so maybe others have different opinions on it.

himdel:
Yeah, the only way all tests will work locally is if devs change their settings.py, not compose.env.

That will change once we're pulling settings from the db, but probably not before that.
(unless we change from allowing tests to change settings to not running tests written for the wrong settings)

So .. if this PR is to make it possible to run all cypress tests locally, it can't rely on devs setting env vars.

(I'm fine with just documenting it though)

brumik:
Where can I document it? Happy to do it

brumik:
Added it to the compose.env. The galaxy_ng docs refer to this file too.

PR #1271, opened 2022-05-16
Fix CI on feature branch

newswangerd:


.github/template_gitref
```diff
@@ -1 +1 @@
-2021.08.26-106-g556224f
+2021.08.26-129-gf780fda-dirty
```
awcrosby, line 1: May need to rebase against master, we are already at `f780fda`

https://github.com/ansible/galaxy_ng/blob/8bbb7a58aa9f64ad3afaf6c4ba9cb46611d3ff5b/.github/template_gitref#L1

awcrosby (changes requested):
May need to rebase

.github/template_gitref
awcrosby: oh nevermind, you are updating this feature branch... why does this say `-dirty`? ... i have not seen that

newswangerd: I think i screwed up my plugin template checkout

newswangerd:
>

newswangerd:
> May need to rebase

I'd have to rebase the whole feature branch against master

PR #1276, opened 2022-05-23
compose.env.example - fix pythonisms

himdel:
The comment was added in #1268, but the example is using python True & False instad of true & false as the rest of the options.

And the note about changing settings.py mentions the name as it would be used in .compose.env, but not in settings.py. (PULP_ prefixed)

Cc @brumik @rochacbruno


brumik:
Good catch for the `PULP_` prefix. For the `True/true` it works with both in my experience.

brumik (approved):

PR #1279, opened 2022-05-25
Upgrade to pulpcore 3.20

newswangerd:
Blocked by:
- https://github.com/pulp/pulpcore/issues/2798
- https://github.com/ansible/galaxy_ng/pull/1280
- https://github.com/pulp/plugin_template/issues/654
- https://github.com/ansible/galaxy_ng/pull/1308
- https://github.com/pulp/plugin_template/pull/656

Todo:
- [x] fix unit tests for migration
- [x] replace `get_groups_with_perms_attached_roles`
- [x] finalize namespace pulp_href
- [x] clean up viewsets.py
- [x] switch to full pulp release
- [x] update plugin template


Makefile
```diff
@@ -159,14 +159,6 @@ docker/all: 	                                ## Build, migrate, loaddata, transl
...
 	make docker/translations
 
-.PHONY: docker/db_snapshot
-docker/db_snapshot:   ## Snapshot database
-	docker exec galaxy_ng_postgres_1 pg_dump -U galaxy_ng -F c -b -f "/galaxy.backup" galaxy_ng
-
-.PHONY: docker/db_restore
-docker/db_restore:   ## Restore database from last snapshot
-	docker exec galaxy_ng_postgres_1 pg_restore --clean -U galaxy_ng -d galaxy_ng "/galaxy.backup"
-
```
alikins, line 169: Commit message says "Fix Makefile". Is there any elaboration? Curious what it broke and how.  I don't think we need this in the Makefile, so no issue with removing it, just curious.

newswangerd: I created these commands to help debug the migration and accidentally merged them with the migration PR. I added them separately to master with some better support here (https://github.com/ansible/galaxy_ng/pull/1281), and they ones from master were causing conflicts.

alikins (approved):
It's a little scary changing migrations once deployed, but it looks like they take any potential problems into account.

Lots of tricky bits to get right here, but looks good to me.  (aside from the failing intg tests)

newswangerd:
> It's a little scary changing migrations once deployed, but it looks like they take any potential problems into account.
> 
> Lots of tricky bits to get right here, but looks good to me. (aside from the failing intg tests)

Yeah, no kidding D:

requirements/requirements.insights.txt
```diff
@@ -90,60 +95,57 @@ diff-match-patch==20200713
...
+django-lifecycle==1.0.0
     # via pulpcore
 django-picklefield==3.0.1
     # via django-automated-logging
 django-prometheus==2.2.0
     # via galaxy-ng (setup.py)
 django-storages[boto3]==1.12.3
     # via -r requirements/requirements.insights.in
-djangorestframework==3.13.1
+django==3.2.13
```
awcrosby, line 118: Can you confirm you are using the latest pip-tools? I see has reordered where some reqs go. I assume if we update to latest then this will be last reorder.

galaxy_ng/app/migrations/0028_move_perms_to_roles.py
```diff
@@ -337,7 +343,11 @@ def migrate_group_permissions_to_roles(apps, schema_editor):
     # Group Permissions
     for group in Group.objects.filter(name__ne="system:partner-engineers"):
         group_roles.extend(get_global_group_permissions(group, Role, GroupRole, Permission))
-        group_roles.extend(get_object_group_permissions(group, Role, GroupRole, ContentType, Permission))
+
+        # Skip migrating object permissions if guardian is not installed.
+        if is_guardian_table_avaiable:
+            group_roles.extend(
+                get_object_group_permissions(group, Role, GroupRole, ContentType, Permission))
```
awcrosby, line 350: When would this table not exist? is that to cover in case the migration is run again, or something else?

galaxy_ng/app/viewsets.py
```diff
@@ -22,41 +22,48 @@
...
+# Added to keep the DRF forms from breaking
+class ContainerRegistryRemoteViewSet(
+    pulp_viewsets.NamedModelViewSet,
+    mixins.RetrieveModelMixin,
+):
     queryset = models.ContainerRegistryRemote.objects.all()
     serializer_class = serializers.ContainerRegistryRemoteSerializer
     permission_classes = [access_policy.ContainerRegistryRemoteAccessPolicy]
-    endpoint_name = "execution-environments-registry-detail"
+    endpoint_name = "galaxy/registry-remote"
```
awcrosby, line 33: should the `endpoint_name` prefixes match on this module... either galaxy or galaxy_ng?

awcrosby (commented):
Added a couple comments

galaxy_ng/app/migrations/0028_move_perms_to_roles.py
newswangerd: that table doesn't exist if you're installing a new instance of hub after guardian was removed (ie not upgrading). I set it up so it only skips the object roles and not the model roles because this lets us run the unit tests for the model roles migrations, which is useful

bmclaughlin: Would this cover fresh installs after guardian has been removed?

bmclaughlin (approved):
Spotted a typo in a function name, but I think it looks good once passing otherwise.

galaxy_ng/app/migrations/0028_move_perms_to_roles.py
awcrosby: The skipping after guardian removal makes sense, but why is it useful to keep running unit tests for the model roles migrations?

template_config.yml
```diff
@@ -27,6 +27,7 @@ check_stray_pulpcore_imports: true
 cherry_pick_automation: false
 ci_env:
   GITHUB_USER: ${{ github.event.pull_request.user.login }}
+  LOCK_REQUIREMENTS: 0
```
newswangerd, line 30: Just an FYI, this will get reverted before the `feature/rbac-roles` branch gets merged into master.

newswangerd:
@mdellweg @bmbouter can the two of you take a look at this as well?

galaxy_ng/app/migrations/0004_rbac.py
```diff
@@ -69,16 +33,16 @@ class Migration(migrations.Migration):
         migrations.RunPython(
             code=initialize_permissions,
         ),
-        migrations.RunPython(
-            code=convert_namespace_groups_to_permissions,
-        ),
+        # migrations.RunPython(
```
bmbouter, line 36: Do you want to leave these commented out but here? Maybe?

bmbouter:
Overall this is looking good. I definitely want to see @mdellweg look at it too. I don't have any additional suggestions.

galaxy_ng/app/migrations/0028_move_perms_to_roles.py
newswangerd: The unit for the migration helped me catch a few bugs and I kinda don't want to get rid of them just yet in case we need to make other changes before the feature branch is merged. I we could remove them in a followup pr. WDYT?

newswangerd:
Well, my `LOCK_REQUIREMENTS: 0` hack doesn't work anymore and using the pulp template git revision doesn't help, so we may be out of luck in terms of CI.

template_config.yml
```diff
@@ -1,15 +1,15 @@
 # This config represents the latest values used when running the plugin-template. Any settings that
 # were not present before running plugin-template have been added with their default values.
 
-# generated with plugin_template@2021.08.26-122-g37dc7b5
+# generated with plugin_template@2021.08.26-136-ga3c43d7
 
 additional_repos:
 - bindings: true
-  branch: 0.13
+  branch: main
```
fao89, line 8:
```suggestion
  revision: e0df10e19782e1d207c179283d4de5a890efda50
```

template_config.yml
```diff
@@ -1,15 +1,15 @@
...
   name: pulp_ansible
   org: pulp
 - bindings: true
-  branch: 2.12.1
+  branch: main
```
fao89, line 12:
```suggestion
  revision: 43f7987176148bf85785b436f7c65a04c802b5d6
```

fao89:
I believe this will help with your PR:
https://github.com/ansible/galaxy_ng/pull/1308

setup.py
```diff
@@ -88,11 +88,11 @@ def run(self):
 
 requirements = [
     "galaxy-importer==0.4.5",
-    "pulpcore>=3.18.1,<3.20.0",
-    "pulp-ansible>=0.13.0,<0.14.0",
+    "pulpcore@git+https://git@github.com/pulp/pulpcore@6e44fb2fe609f92dc1f502b19c67abd08879148f#egg=pulpcore",  # noqa E501
```
fao89, line 91: it doesn't have `.git`

.github/workflows/scripts/install.sh
```diff
@@ -28,24 +28,14 @@ pip install -r functest_requirements.txt
...
-fi
-
+PULP_ANSIBLE=git+https://github.com/pulp/pulp_ansible.git@e0df10e19782e1d207c179283d4de5a890efda50#egg=pulp-ansible
+PULP_CONTAINER=git+https://github.com/pulp/pulp_container.git@43f7987176148bf85785b436f7c65a04c802b5d6#egg=pulp-container
 if [ -e $REPO_ROOT/../galaxy-importer ]; then
   GALAXY_IMPORTER=./galaxy-importer
 else
   GALAXY_IMPORTER=git+https://github.com/ansible/galaxy-importer.git@v0.4.5
 fi
+PULPCORE=git+https://github.com/pulp/pulpcore.git@6e44fb2fe609f92dc1f502b19c67abd08879148f#egg=pulpcore
```
fao89, line 38: it does have `.git`

fao89 (commented):
I believe the `.git` is why it still failing

galaxy_ng/app/access_control/mixins.py
```diff
@@ -60,15 +33,23 @@ def _set_groups(self, groups):
         if self._state.adding:
             self._groups = groups
         else:
-            current_groups = get_groups_with_perms_attached_roles(self)
+            obj = self
+
+            # If the model is a proxy model, get the original model since pulp
+            # doesn't allow us to assign permissions to proxied models.
```
mdellweg, line 39: That comment is confusing to me. I know, Pulp allows to assign roles via the proxy model, but yes they need to contain permissions of that proxy model (the ones of the base model would have no effect). This was needed in order to define new permissions for the proxy model.

I believe this to be in line with: https://docs.djangoproject.com/en/3.2/topics/auth/default/#proxy-models

newswangerd: @mdellweg I was getting "this role doesn't have permissions on the object" error from pulps `assign_role` function.

The role check uses `for_concrete_model=False`. https://github.com/pulp/pulpcore/blob/main/pulpcore/app/role_util.py#L34

I can add a patch for this in pulpcore if you want.

mdellweg: This parameter is confusing me all the time too; `for_concrete_model=False` means the content_type fetched in the database is the content type of the proxy model [0] and not the one of the (concrete) one with the database table. That means, pulp checks permissions against the proxy model. The problem is, you cannot use "auth.view_group", but you must use the "core.view_group" permission (`pulpcore.models.Group` is the actual example i ran into trouble with). All in all, i think, this means you should use the proxy model as if it were a real one, and i think this is how django auth interprets permissions, too.

[0] https://docs.djangoproject.com/en/3.2/ref/contrib/contenttypes/#django.contrib.contenttypes.models.ContentTypeManager.get_for_model

newswangerd: If we got https://github.com/pulp/pulpcore/issues/2769 fixed, then we wouldn't need to proxy the models to add our helpers for getting and setting permissions.

awcrosby (approved):

PR #1284, opened 2022-05-31
[dev env] Reload API when .compose.env is changed

rochacbruno:
When developer makes changes on `.compose.env` or `dev/common/settings.py`
the API container reloads the gunicorn process and settings are reflected immediately.

Affects: Docker based dev environment only

No-Issue


jctanner (approved):

himdel (approved):
Changed `PULP_GALAXY_REQUIRE_CONTENT_APPROVAL` in `.compose.env`, the workers restarted, seems to work :+1: 

(Can't be used for `PULP_DB_NAME` though, that gets either ignored or overwritten in `dev/common/galaxy_ng.env`)

PR #1285, opened 2022-05-31
[dev env] reload pulpcore-worker when file changes

rochacbruno:
When any `*.py` or `*.env` file changes on galaxy_ng or any other directory added to `DEV_SOURCE_PATH` then worker reloads inside the container.

Affects: docker based dev environment only

No-Issue

### DIFF
--- a/dev/Dockerfile.base
+++ b/dev/Dockerfile.base
@@ -39,6 +39,7 @@ RUN set -ex; \
         libpq \
         libpq-devel \
         pinentry \
+        make \
     && dnf clean all \
     && rm -rf /var/cache/dnf/ \
     && rm -f /var/lib/rpm/__db.* \
@@ -46,6 +47,13 @@ RUN set -ex; \
     && chown ${USER_NAME}:${USER_GROUP} /venv \
     && mkdir --mode=2775 /app \
     && chown ${USER_NAME}:${USER_GROUP} /app
+
+# Install entr so we can have reload on worker processes
+RUN git clone https://github.com/eradman/entr && \
+    cd entr && \
+    cp Makefile.linux Makefile && \
+    make && \
+    make install
 
 USER ${USER_NAME}:${USER_GROUP}
 RUN set -ex; \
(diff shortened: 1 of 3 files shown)


dev/Dockerfile.base
```diff
@@ -47,6 +48,13 @@ RUN set -ex; \
     && mkdir --mode=2775 /app \
     && chown ${USER_NAME}:${USER_GROUP} /app
 
+# Install entr so we can have reload on worker processes
+RUN git clone https://github.com/eradman/entr && \
+    cd entr && \
+    cp Makefile.linux Makefile && \
+    make && \
+    make install
```
rochacbruno, line 56: `entr` is available on fedore repos but not for our redhat base image, so decided to build it from git.

jctanner (approved):

docker/bin/start-worker-reload
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+set -o errexit
+set -o nounset
+
+WATCHED_FILES=$(python -c "print(' '.join([f'/src/{item}' for item in '$DEV_SOURCE_PATH'.split(':')]))")
+exec find $WATCHED_FILES \( -path /src/galaxy_ng/.venv -o -path /src/galaxy_ng/build -o -path /src/galaxy_ng/.eggs \) -prune -o -name '*.py' -o -name '*.env' | entr -n -r pulpcore-worker
```
alikins, line 7: why the exec here?

rochacbruno: a successful exec never returns and that is how we make this to be the docker container command that never ends, otherwise it runs and stops the container soon.

alikins, line 7: Could this make use of `git ls-files` in lieu of walking the file tree and excluding stuff?

For me at least, I tend to to prefer if tools anything that isn't explicitly in the project. But that is because my source checkouts get littered with all kind of garbage (log files, test-of-something.py, somehelpertool.py, etc)

I know opinion is divided on that expectation though.

PR #1286, opened 2022-05-31
Add a rediirect for the /pulp/ api endpoints when they are not set to the default.

newswangerd:
… 
Issue: AAH-1646


galaxy_ng/app/urls.py
```diff
@@ -47,6 +47,14 @@
     path("healthz", views.health_view),
 ]
 
+if settings.get("API_ROOT") != "/pulp/":
```
awcrosby, line 50: do we need this conditional since we made this not user-configurable?

newswangerd: I like to cover the bases, but probably not

PR #1289, opened 2022-05-31
Update README's openapi spec url.

jctanner:
#### What is this PR doing:
```
speaking of galaxy_ng, I tried to look at the open API spec listed in the readme and the link is broken
seems like it was removed in some PR because it was out of date and the app auto-generates one at runtime
but for someone not running it, there's now no way to view it, and the readme still has that broken link
would've opened an issue, but that customized jira instance hosting the issues is um... super hard to interact with, sorry
```
#### Notes: 

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ To learn more about Pulp, [view the Pulp project page](https://pulpproject.org/)
 
 ## OpenAPI Spec
 
-View the latest version of the spec by [clicking here](https://petstore.swagger.io/?url=https://raw.githubusercontent.com/ansible/galaxy_ng/master/openapi/openapi.yaml).
+View the latest version of the spec by [clicking here](https://console.redhat.com/api/automation-hub/v3/openapi.yaml).
 
 ## Contributing
 


README.md
```diff
@@ -14,7 +14,7 @@ To learn more about Pulp, [view the Pulp project page](https://pulpproject.org/)
 
 ## OpenAPI Spec
 
-View the latest version of the spec by [clicking here](https://petstore.swagger.io/?url=https://raw.githubusercontent.com/ansible/galaxy_ng/master/openapi/openapi.yaml).
+View the latest version of the spec by [clicking here](https://console.redhat.com/api/automation-hub/v3/openapi.yaml).
```
rochacbruno, line 17: Not sure if we want to mention `CRC` in the upstream as a docs resource.

We have talked about replacing all the `automation-hub` references from upstream repo and use only `galaxy_ng`

So looks like we need to publish the open api spec somewhere else.

Ideally we would generate the openapi-spec as a static site and publish with the docs we have.

awcrosby (changes requested):
In the readme we can be explicit saying it is only generated at runtime, so at least is accurate, since we do not have a publicly published one. We should have an issue to generate an openapi spec and publish in docs or swagger.

newswangerd (commented):
+1 to not linking to c.rh.c

PR #1290, opened 2022-06-01
Rm synclist code AAH-1556

alikins:
#### What is this PR doing:

Remove all usage of SyncList.repository and SyncList.upstream_repository  and remove leftover synclist code

Issue: AAH-1556

#### Notes: 

### DIFF
--- a/galaxy_ng/app/api/v3/viewsets/collection.py
+++ b/galaxy_ng/app/api/v3/viewsets/collection.py
@@ -267,7 +267,8 @@ class CollectionVersionMoveViewSet(api_base.ViewSet):
         response_data = {
             "copy_task_id": None,
             "remove_task_id": None,
-            # Preserve after synclist removal for client compat -akl
+            # Can be removed once all synclist stuff is remove
+            # and client compat isnt a concern -akl
             "curate_all_synclist_repository_task_id": None,
         }
         golden_repo = settings.get("GALAXY_API_DEFAULT_DISTRIBUTION_BASE_PATH", "published")
--- a/galaxy_ng/tests/integration/utils.py
+++ b/galaxy_ng/tests/integration/utils.py
@@ -383,7 +383,7 @@ def set_certification(client, collection):
         assert 'remove_task_id' in job_tasks
 
         # wait for each unique task to finish ...
-        for key in ['copy_task_id', 'curate_all_synclist_repository_task_id', 'remove_task_id']:
+        for key in ['copy_task_id', 'remove_task_id']:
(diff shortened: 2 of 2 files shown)


galaxy_ng/app/auth/auth.py
```diff
@@ -65,58 +60,15 @@ def authenticate(self, request):
...
         return group, created
 
-    def _ensure_synclists(self, group):
-        with transaction.atomic():
-            # check for existing synclists
-            perms = ['galaxy.view_synclist']
-
-            synclists_owned_by_group = \
-                shortcuts.get_objects_for_group(group, perms, klass=SyncList,
-                                                any_perm=False, accept_global_perms=True)
```
awcrosby, line 86: The SyncList object is still being used on CRC. We are only removing the SyncList.repository and the curation of those repos.

galaxy_ng/tests/integration/api/test_synclist.py
```diff
@@ -1,165 +0,0 @@
-"""test_synclist.py - Test related to synclists."""
```
awcrosby, line 1: We'll want to keep all of these tests

awcrosby (changes requested):

galaxy_ng/app/settings.py
```diff
@@ -40,10 +40,6 @@
 GALAXY_API_STAGING_DISTRIBUTION_BASE_PATH = "staging"
 GALAXY_API_REJECTED_DISTRIBUTION_BASE_PATH = "rejected"
 
-# The format for the name of the per account synclist, and the
-# associated repository, distribution, and distribution base_paths
```
awcrosby, line 44: Still needed, so think just need to remove `repository` from the code comment

galaxy_ng/app/settings.py
```diff
@@ -40,10 +40,6 @@
...
-GALAXY_API_SYNCLIST_NAME_FORMAT = "{account_name}-synclist"
```
awcrosby, line 45: This is still used in auth.py, so we want to keep this setting. We do need to make sure auth.py `_ensure_synclists` does not call `_ensure_repository` which creates a synclist repo

awcrosby (changes requested):
Added some comments.

galaxy_ng/app/settings.py
awcrosby: We also need to remove `_ensure_repository` from auth/auth.py so we do not create synclist repos anymore

It would also be good to do a general sweep of the code to make sure we are not looking up synclist repos anywhere

PR #1291, opened 2022-06-02
Add make commands for functional tests.

newswangerd:
No-Issue

## Instructions

Create a new python virtual environment and activate it.

```
python3 -m venv venv/

source venv/bin/activate
```

Clone [pulp open api generator](https://github.com/pulp/pulp-openapi-generator) into the directory above galaxy_ng as well as any other pulp plugins you want to test (such as pulp_ansible or galaxy_ng). The parent dir should look something like this:

```
..
├── ansible-hub-ui
├── galaxy-importer
├── galaxy_ng
├── pulp-openapi-generator
├── pulp_ansible
├── pulp_container
└── pulpcore
```

With the `./compose up` environment running, run the following make commands from the galaxy_ng project:

NOTE: Pulp smash will ask you for your password to inspect the containers.

```
# Install the test dependencies and bindings.
make test/functional/install_requirements PLUGIN=pulp_ansible

# Run the tests
make test/functional/run PLUGIN=pulp_ansible FLAGS="-k test_download_with_content_guard"
```

After making any changes to the API (this includes query parameters, serializers, routes etc) you'll need to rebuild the pulp bindings:

```
make test/functional/generate_client PLUGIN=pulp_ansible
```

Notes:
- `PLUGIN` is required on each of these make commands and should be the name of the git checkout for the plugin. `pulpcore`, `galaxy_ng`, `pulp_ansible` and `pulp_container` should work.
- `test/functional/run` takes and optional `FLAGS` argument that allows the user to pass pytest flags to pytest. A handy one is `-k name_of_my_test` which allows the user to run any test that match the given name.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -243,3 +243,17 @@ docs/build:
 
 docs/serve:
 	@mkdocs serve
+
+.PHONY: test/functional/generate_client
+test/functional/generate_client:         ## Opens bash session in the api container
+	./dev/pulp_smash/generate_client.sh $(PLUGIN)
+
+.PHONY: test/functional/install_requirements
+test/functional/install_requirements:         ## Opens bash session in the api container
+	./dev/pulp_smash/generate_client.sh pulpcore
+	./dev/pulp_smash/generate_client.sh $(PLUGIN)
+	./dev/pulp_smash/install_requirements.sh $(PLUGIN)
+
+.PHONY: test/functional/run
+test/functional/run:         ## Opens bash session in the api container
+	./dev/pulp_smash/run_functional_tests.sh $(PLUGIN) $(FLAGS)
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -243,3 +243,17 @@ docs/build:
 
 docs/serve:
 	@mkdocs serve
+
+.PHONY: test/functional/generate_client
+test/functional/generate_client:         ## Opens bash session in the api container
```
newswangerd, line 248: Note to self to fix the comments here

brumik:
After installing some missing packages on my Ubuntu this ran just fine (some test with signing fail but I think you need to add more setup for those?)

brumik:
Actually as I was going trough, all tests in the `cli` are failing for `pulp_ansible`. It seems like the cause of the error is `Unauthorized`.

newswangerd:
No longer needed now that we have OCI Env

PR #1310, opened 2022-06-08
Add upload statements to insights

rochacbruno:
Issue: AAH-1700


rochacbruno:
@awcrosby we have to test manually now, and later enable the tests, this tests needs the adjusts to use different user profiles and auth, also the tests are using requests to make api calls directly and that is implemented only for local/standalone urls.

awcrosby (approved):

PR #1313, opened 2022-06-10
Adds python package name to AppConfig

bmbouter:
This is a pulpcore 3.20 compatibility requirement.

[noissue]


bmbouter:
@newswangerd you could also add this into your 3.20 PR, which would be easier maybe?

newswangerd:
Closing this and reopening it against the feature/rbac-roles branch (where we're doing the upgrade work)

https://github.com/ansible/galaxy_ng/pull/1314

PR #889, opened 2021-08-10
Update settings.py with Social Auth values when environment is configured

chambridge:
* Update standalone requirements
* Update settings conditional on environment
* Update urls is settings are present

Partial-Implements: AAH-846

PR #1323, opened 2022-06-20
dev standalone-keycloak: use keycloak/keycloak:legacy, not latest

himdel:
when #889 was merged, the current keycloak version was 15.0.2,
since 17.0.0, `docker run keycloak/keycloak` just shows help instead of running the server
and there's a new `*-legacy` version which still runs the server

Updating to latest legacy, tested with 18.0.1-legacy

(this seems related to https://github.com/keycloak/keycloak/commit/30d2dcb7b3e6f283e4bdd463e051dd8fb80af836 and making https://github.com/keycloak/keycloak-containers legacy)

No-Issue

---

Somebody should test this actually works, I've seen it work and I've seen it fail, either could have been caused by some extra data from running keycloak 15.0.2 which definitely works.

### DIFF
--- a/dev/standalone-keycloak/docker-compose.yml
+++ b/dev/standalone-keycloak/docker-compose.yml
@@ -15,7 +15,7 @@ services:
       - './standalone-keycloak/galaxy_ng.env'
       - '../.compose.env'
   keycloak:
-    image: quay.io/keycloak/keycloak:latest
+    image: quay.io/keycloak/keycloak:legacy
     env_file:
       - './standalone-keycloak/keycloak.env'
       - '../.compose.env'


himdel:
Testing...

`.compose.env`:

```
COMPOSE_PROFILE=standalone-keycloak
DEV_SOURCE_PATH='galaxy_ng'
LOCK_REQUIREMENTS=1
ANSIBLE_HUB_UI_PATH='/home/himdel/ansible-hub-ui'
DEV_IMAGE_SUFFIX='standalone-keycloak'
WAIT_FOR_MIGRATIONS=1
ENABLE_SIGNING=1
PULP_GALAXY_REQUIRE_CONTENT_APPROVAL=false
PULP_GALAXY_REQUIRE_SIGNATURE_FOR_APPROVAL=false
PULP_GALAXY_AUTO_SIGN_COLLECTIONS=true
#PULP_SOCIAL_AUTH_KEYCLOAK_PUBLIC_KEY="..."
```

```
./compose build
./compose up -d postgres redis
./compose up -d keycloak kc-postgres ldap

ansible-galaxy collection install community.general --upgrade
ansible-playbook ./dev/standalone-keycloak/keycloak-playbook.yaml
```

keep retrying the last one if it fails, until it outputs a long token.

Before: 

```
TASK [Create or update AAP Keycloak realm] *********************************************************************************************
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Could not obtain access token from http://localhost:8080/auth/realms/master/protocol/openid-connect/token: timed out"}
```

and `./compose ps` indicated state `Exit 0` for the `keycloak` container.


After: the script should pass (eventually), the container should not die, and there should be a token on stdout.

... then update `.compose.env` with `PULP_SOCIAL_AUTH_KEYCLOAK_PUBLIC_KEY="that long token"`

```
./compose run --rm api manage migrate
./compose run --rm -e PULP_FIXTURE_DIRS='["/src/galaxy_ng/dev/automation-hub"]' api manage loaddata initial_data.json
./compose down

# actually run
./compose up
```

(Keycloak login screen should load when accessing localhost:8002)

jctanner (approved):

PR #1325, opened 2022-06-21
Poll the move task(s) during integration

jctanner:
#### What is this PR doing:
Move tasks weren't pollable before? Now they are? ... anyways, poll them and check state.

#### Notes: 

### DIFF
--- a/galaxy_ng/tests/integration/utils.py
+++ b/galaxy_ng/tests/integration/utils.py
@@ -375,7 +375,21 @@ def set_certification(client, collection):
             f"{collection.version}/move/staging/published/"
         )
 
-        client(url, method="POST", args=b"{}")
+        job_tasks = client(url, method="POST", args=b"{}")
+        assert 'copy_task_id' in job_tasks
+        assert 'curate_all_synclist_repository_task_id' in job_tasks
+        assert 'remove_task_id' in job_tasks
+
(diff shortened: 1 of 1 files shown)


galaxy_ng/tests/integration/utils.py
```diff
@@ -375,7 +375,20 @@ def set_certification(client, collection):
...
         )
 
-        client(url, method="POST", args=b"{}")
+        job_tasks = client(url, method="POST", args=b"{}")
+        assert 'copy_task_id' in job_tasks
+        assert 'curate_all_synclist_repository_task_id' in job_tasks
+        assert 'remove_task_id' in job_tasks
+
+        task_ids = list(job_tasks.values())
+        task_ids = [x for x in task_ids if x]
```
awcrosby, line 384: will this be of arbitrary order? not sure if we care to try to wait on copy first... i guess if we wait on remove first and copy fails then remove task will eventually timeout i assume

jctanner: I don't know that the order really matters. Does an end user or client know which order those tasks should start and complete in?

awcrosby: no it doesn't. but for testing, if copy fails, remove will likely not be `complete` it might be `canceled` ... nice for a test to fail as close as possible to the failing code.

PR #1346, opened 2022-07-06
Update filter for indexing EEs from catalog.redhat.com

newswangerd:
Issue: AAH-1737


newswangerd:
@awcrosby I'm not adding integration tests, because this has to be backported to 4.4

awcrosby (approved):

PR #1367, opened 2022-07-13
remove conditional 'view_group'

ShaiahWren:
Issue: AAH-1805

With the current `view_group` conditional, users without this permission cannot view the roles list nor a group detail view. This should not be the case. This pr simply removes that conditional.


CHANGES/1805.bug
```diff
@@ -0,0 +1 @@
+Remove conditional `view_task`.
```
newswangerd, line 1: .bugfix

galaxy_ng/app/access_control/statements/standalone.py
```diff
@@ -400,7 +400,6 @@
             "action": ["list", "retrieve"],
             "principal": "authenticated",
             "effect": "allow",
-            "condition": "has_model_perms:galaxy.view_group"
```
newswangerd, line 403: this needs to go in insights.py too

newswangerd (commented):
Can you write a quick integration test to verify that any user without permissions can read data from the roles api, but not change anything?

ShaiahWren:
@newswangerd Can you clarify 1. which specific permissions should/not be assigned in order to manipulate roles and 1. what actions on roles should/not be allowed. Edit, delete?

newswangerd (approved):

PR #1377, opened 2022-07-18
standalone-community social auth

jctanner:
#### What is this PR doing:

Adds github social auth.

Issue: AAH-2526


galaxy_ng/app/settings.py
```diff
@@ -21,6 +21,7 @@
 
 INSTALLED_APPS = [
     'rest_framework.authtoken',
+    'social_django',
```
rochacbruno, line 24: Maybe this must be added on dev/standalone-community env only?
adding here will require this app to be available for all intallations

jctanner: I agree, I'm just now sure how to do that. How did you inject that with the ldap mode? I notice it also uses the same library, which is why I picked it.

jctanner: Oh, I found it ... https://github.com/ansible/galaxy_ng/blob/facddc251698d1ac9cbd0dbb926956b6dbb428ca/galaxy_ng/app/dynaconf_hooks.py#L122-L123

I'll get that patched in ASAP.

dev/standalone-community/docker-compose-ui.yml
```diff
@@ -5,4 +5,4 @@ services:
     environment:
       - "API_PROXY_HOST=api"
       - "API_PROXY_PORT=8000"
-      - "DEPLOYMENT_MODE=${COMPOSE_PROFILE}"
```
rochacbruno, line 8: shouldn't this be `DEPLOYMENT_MODE=standalone-community` ?

rochacbruno: Actually, we have only 2 deployment modes AFAIK: standalone and insights, I think we don't want to add a new one, is that useful? or can we stick with `standalone` here?

jctanner: Maybe, but all the the other UI files are using a variable:

```
[jtanner@p1 galaxy_ng]$ find dev -type f -name "*ui.yaml" | xargs fgrep -H DEPLOYMENT_MODE=
dev/insights/docker-compose-ui.yaml:      - "DEPLOYMENT_MODE=${COMPOSE_PROFILE}"
dev/standalone-keycloak/docker-compose-ui.yaml:      - "DEPLOYMENT_MODE=${COMPOSE_PROFILE}"
dev/standalone-ldap/docker-compose-ui.yaml:      - "DEPLOYMENT_MODE=${COMPOSE_PROFILE}"
dev/standalone/docker-compose-ui.yaml:      - "DEPLOYMENT_MODE=${COMPOSE_PROFILE}"
dev/standalone-community/docker-compose-ui.yaml:      - "DEPLOYMENT_MODE=${COMPOSE_PROFILE}"
```

jctanner: I hardcoded it to standalone. We may need to revise later when v1 support is added depending on the capability of feature flags.

PR #1381, opened 2022-07-20
Update pulp dependencies from commit hashes to pinned versions

awcrosby:
#### What is this PR doing:
Update pulp dependencies from commit hashes to pinned versions

No-Issue


requirements/requirements.standalone.txt
```diff
@@ -36,48 +56,97 @@ bracex==2.3.post1
...
 django-automated-logging==6.1.3
     # via galaxy-ng (setup.py)
+django-currentuser==0.5.3
+    # via pulpcore
+django-filter==22.1
+    # via pulpcore
+django-guid==3.3.0
+    # via pulpcore
+django-import-export==2.8.0
+    # via pulpcore
```
awcrosby, line 109: @newswangerd these and other dependencies generate with `pip-compile` when using standard requirements format, but not with `_format_pulp_requirement()`

awcrosby: In the `feature/rbac-roles` branch while `requirements.standalone.txt` does not have for example `django-currentuser` and `django-filter` ... after doing a `./compose up` in the developer environment a `pip freeze` shows these in fact are installed, so this is probably fine for development, but we will need to correct this if we want to use `_format_pulp_requirement()` to pin to revisions on `master` for CRC deploys

newswangerd: hmmm, that's really disappointing :/

I guess we'll just have to document how to run pulp from a commit and hope people get it right

setup.py
```diff
@@ -110,11 +110,11 @@ def _format_pulp_requirement(plugin, specifier=None, ref=None, gh_namespace="pul
 
 requirements = [
     "galaxy-importer==0.4.5",
-    _format_pulp_requirement("pulpcore", ref="6e44fb2fe609f92dc1f502b19c67abd08879148f"),
```
newswangerd, line 113: @awcrosby if `_format_pulp_requirement` doesn't work, can you add a comment with an example for pinning pulp  to a git commit and some notes about how to pin to a commit?

newswangerd (changes requested):

setup.py
awcrosby: @newswangerd I don't know how to pin pulp deps to a commit and get it to work with both pip-tools and pulp ci. I feel like the current `_format_pulp_requirement` is going in the right direction. I can put a fixme on that method noting its limitations, how does that sound?

newswangerd: alright, that sounds good

newswangerd (approved):

PR #1396, opened 2022-08-02
Fix sign_state query for faster collection signing performance

ShaiahWren:
Issue: AAH-1794

Remove counters for partially signed, all signed, and all unsigned collection versions and instead return the sign state of only the latest version of a collection.

### DIFF
--- a/CHANGES/1794.bugfix
+++ b/CHANGES/1794.bugfix
@@ -1,0 +1,1 @@
+Return only the sign state of the latest version of a collection.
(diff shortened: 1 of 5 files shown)


rochacbruno:
The CI linter error is unrelated to this PR, looks like it is a style error that was not being catch by linter but started to be a problem with a linter update.

https://github.com/ansible/galaxy_ng/blob/bf3503edc4098ed3a89a5f051c2d10c6ceb6347f/galaxy_ng/app/api/ui/serializers/execution_environment.py#L400 

That `if(..` needs a space `if (..`

rochacbruno (approved):

newswangerd (approved):

PR #1411, opened 2022-08-11
update annotation method

ShaiahWren:
No-Issue

Connected to https://github.com/ansible/galaxy_ng/pull/1396 / AAH-1794 for fixing the annotation method call.

### DIFF
--- a/galaxy_ng/app/api/ui/viewsets/collection.py
+++ b/galaxy_ng/app/api/ui/viewsets/collection.py
@@ -141,7 +141,12 @@ class CollectionViewSet(
             name=self.kwargs["name"],
         )
         return get_object_or_404(
-            base_qs.annotate(**self.build_signing_annotations(base_qs)),
+            base_qs.annotate(
+            sign_state=Case(
+                When(signatures__isnull=False, then=Value("signed")),
+                When(signatures__isnull=True, then=Value("unsigned")),
+    )
+),
             version=version,
         )
 
--- a/galaxy_ng/app/api/ui/viewsets/collection.py
+++ b/galaxy_ng/app/api/ui/viewsets/collection.py
@@ -142,11 +142,11 @@ class CollectionViewSet(
         )
         return get_object_or_404(
             base_qs.annotate(
-            sign_state=Case(
-                When(signatures__isnull=False, then=Value("signed")),
-                When(signatures__isnull=True, then=Value("unsigned")),
-    )
-),
+                sign_state=Case(
+                    When(signatures__isnull=False, then=Value("signed")),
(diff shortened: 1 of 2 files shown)


galaxy_ng/app/api/ui/viewsets/collection.py
```diff
@@ -141,7 +141,12 @@ def get_object(self):
             name=self.kwargs["name"],
         )
         return get_object_or_404(
-            base_qs.annotate(**self.build_signing_annotations(base_qs)),
+            base_qs.annotate(
+            sign_state=Case(
+                When(signatures__isnull=False, then=Value("signed")),
+                When(signatures__isnull=True, then=Value("unsigned")),
+    )
+),
```
rochacbruno, line 149: Indentation problem here

```suggestion
            base_qs.annotate(
                sign_state=Case(
                    When(signatures__isnull=False, then=Value("signed")),
                    When(signatures__isnull=True, then=Value("unsigned")),
                )
            ),
```

rochacbruno (changes requested):
There is an indentation problem caught by CI linter

Also we don't have tests for that specific part (that is the reason we missed it on previous PR)

Would be nice to add a simple test to check that endpoint, a test similar to this one https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/tests/integration/api/test_collection_signing.py#L139-L150 but checking the detail endpoint instead of the list endpoint.

List endpoint is: /_ui/v1/repo/published/  (being tested)

Detail endpoint is: /api/automation-hub/_ui/v1/repo/published/{namespace}/{collection}/?version=x.y.z

galaxy_ng/tests/integration/api/test_collection_signing.py
```diff
@@ -149,6 +149,19 @@ def test_collection_auto_sign_on_approval(api_client, config, settings, flags, u
     assert metadata["signatures"][0]["signature"].startswith("-----BEGIN PGP SIGNATURE-----")
     assert metadata["signatures"][0]["pubkey_fingerprint"] is not None
 
+    # Assert that the collection is signed on UI API (detail endpoint)
+    collection_on_ui = api_client(
+        "/api/automation-hub/_ui/v1/repo/published/{namespace}/{collection}/?version=x.y.z"
```
rochacbruno, line 154: It is needed to interpolate the URL str with the variables
```suggestion
        f"/api/automation-hub/_ui/v1/repo/published/{NAMESPACE}/{artifact.name}/?version={artifact.version}"
```

rochacbruno (changes requested):

rochacbruno (approved):

PR #1430, opened 2022-08-25
UI permissions refactor

ShaiahWren:
[WIP]
Issue: AAH-1714

First bit of the UI Permissions refactor. 
All relevant permissions are added to the `galaxy_ng/app/constants.py` file. 

Reviewers, please check for appropriate phrasing for each permission.

Thank you!

### DIFF
--- a/galaxy_ng/app/api/ui/serializers/user.py
+++ b/galaxy_ng/app/api/ui/serializers/user.py
@@ -1,3 +1,4 @@
+from math import perm
 from django.conf import settings
 from django.contrib.auth import password_validation
 from django.utils.translation import gettext_lazy as _
@@ -171,68 +172,13 @@ class CurrentUserSerializer(UserSerializer):
 
     @extend_schema_field(OpenApiTypes.OBJECT)
     def get_model_permissions(self, obj):
-
-        permissions = {
-            # Collection Namespace
-            "add_namespace": obj.has_perm('galaxy.add_namespace'),
-            "upload_to_namespace": obj.has_perm('galaxy.upload_to_namespace'),
-            "change_namespace": obj.has_perm('galaxy.change_namespace'),
-            "delete_namespace": obj.has_perm('galaxy.delete_namespace'),
-
-            # Collection Repository
-            "move_collection": obj.has_perm('ansible.modify_ansible_repo_content'),
-            "delete_collection": obj.has_perm('ansible.delete_collection'),
-
-            # Collection Remote
-            "add_remote": obj.has_perm('ansible.add_collectionremote'),
-            "change_remote": obj.has_perm('ansible.change_collectionremote'),
-            "delete_remote": obj.has_perm('ansible.delete_collectionremote'),
-
-            # Collection Distribution
-            "view_distribution": obj.has_perm('ansible.view_ansibledistribution'),
-            "add_distribution": obj.has_perm('ansible.add_ansibledistribution'),
-            "change_distribution": obj.has_perm('ansible.change_ansibledistribution'),
-            "delete_distribution": obj.has_perm('ansible.delete_ansibledistribution'),
-
-            # Container Namespace
-            "add_containernamespace": obj.has_perm('container.add_containernamespace'),
-            "change_containernamespace": obj.has_perm('container.change_containernamespace'),
-            "delete_containernamespace": obj.has_perm('container.delete_containernamespace'),
-
-            # Container Repository
-            "add_containerrepository": obj.has_perm('container.add_containerrepository'),
-            "change_containerrepository": obj.has_perm('container.change_containerrepository'),
-            "delete_containerrepository": obj.has_perm('container.delete_containerrepository'),
-
-            # Container Remote
-            "add_containerregistry": obj.has_perm('galaxy.add_containerregistryremote'),
-            "change_containerregistry": obj.has_perm('galaxy.change_containerregistryremote'),
-            "delete_containerregistry": obj.has_perm('galaxy.delete_containerregistryremote'),
-
-            # Container Distribution
-            "add_containerdistribution": obj.has_perm('container.add_containerdistribution'),
-            "change_containerdistribution": obj.has_perm('container.change_containerdistribution'),
-            "delete_containerdistribution": obj.has_perm('container.delete_containerdistribution'),
-
-            # Tasks
-            "view_task": obj.has_perm('core.view_task'),
-
-            # Auth
-            "view_user": obj.has_perm('galaxy.view_user'),
-            "change_group": obj.has_perm('galaxy.change_group'),
-            "view_group": obj.has_perm('galaxy.view_group'),
-            "delete_user": False,
-            "change_user": False,
-            "add_user": False,
-            "add_group": False,
-            "delete_group": False,
-        }
+        permissions = { "permissions": PERMISSIONS}
 
         if not settings.get("SOCIAL_AUTH_KEYCLOAK_KEY"):
-            permissions["delete_user"] = obj.has_perm('galaxy.delete_user')
-            permissions["change_user"] = obj.has_perm('galaxy.change_user')
-            permissions["add_user"] = obj.has_perm('galaxy.add_user')
-            permissions["add_group"] = obj.has_perm('galaxy.add_group')
-            permissions["delete_group"] = obj.has_perm('galaxy.delete_group')
+            permissions["permissions"]["galaxy.delete_user"] = obj.has_perm('galaxy.delete_user')
+            permissions["permissions"]["galaxy.change_user"] = obj.has_perm('galaxy.change_user')
+            permissions["permissions"]["galaxy.add_user"] = obj.has_perm('galaxy.add_user')
+            permissions["permissions"]["galaxy.add_group"] = obj.has_perm('galaxy.add_group')
+            permissions["permissions"]["galaxy.delete_group"] = obj.has_perm('galaxy.delete_group')
 
         return permissions
--- a/galaxy_ng/app/api/ui/serializers/user.py
+++ b/galaxy_ng/app/api/ui/serializers/user.py
@@ -172,13 +172,13 @@ class CurrentUserSerializer(UserSerializer):
 
     @extend_schema_field(OpenApiTypes.OBJECT)
     def get_model_permissions(self, obj):
-        permissions = { "permissions": PERMISSIONS}
+        permissions = { PERMISSIONS}
 
         if not settings.get("SOCIAL_AUTH_KEYCLOAK_KEY"):
-            permissions["permissions"]["galaxy.delete_user"] = obj.has_perm('galaxy.delete_user')
-            permissions["permissions"]["galaxy.change_user"] = obj.has_perm('galaxy.change_user')
-            permissions["permissions"]["galaxy.add_user"] = obj.has_perm('galaxy.add_user')
-            permissions["permissions"]["galaxy.add_group"] = obj.has_perm('galaxy.add_group')
-            permissions["permissions"]["galaxy.delete_group"] = obj.has_perm('galaxy.delete_group')
+            permissions["galaxy.delete_user"] = obj.has_perm('galaxy.delete_user')
+            permissions["galaxy.change_user"] = obj.has_perm('galaxy.change_user')
+            permissions["galaxy.add_user"] = obj.has_perm('galaxy.add_user')
+            permissions["galaxy.add_group"] = obj.has_perm('galaxy.add_group')
+            permissions["galaxy.delete_group"] = obj.has_perm('galaxy.delete_group')
 
         return permissions
--- a/galaxy_ng/app/api/ui/serializers/user.py
+++ b/galaxy_ng/app/api/ui/serializers/user.py
@@ -10,6 +10,7 @@ from rest_framework.exceptions import ValidationError
 from galaxy_ng.app.models import auth as auth_models
 from galaxy_ng.app.constants import PERMISSIONS
 
+
 class GroupSerializer(serializers.ModelSerializer):
 
     class Meta:
--- a/galaxy_ng/app/api/ui/serializers/user.py
+++ b/galaxy_ng/app/api/ui/serializers/user.py
@@ -173,15 +173,13 @@ class CurrentUserSerializer(UserSerializer):
     def get_model_permissions(self, obj):
         permissions = {}
         for i, j in PERMISSIONS.items():
-            print(f'i & j: {i, j}')
             permissions[i] = j
-        for i in permissions:
             permissions[i]["has_model_permission"] = obj.has_perm(i)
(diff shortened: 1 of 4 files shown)


galaxy_ng/app/constants.py
```diff
@@ -16,3 +16,370 @@ class DeploymentMode(enum.Enum):
 )
 
 INBOUND_REPO_NAME_FORMAT = "inbound-{namespace_name}"
+
+PERMISSIONS = {
+    "galaxy.add_namespace": {
+        # Short name to display in the UI
```
newswangerd, line 22: you don't have to copy these comments to every entry here.

galaxy_ng/app/constants.py
```diff
@@ -16,3 +16,319 @@ class DeploymentMode(enum.Enum):
...
+        "object_description": None,
+
+        # Description to use when the permission is being assigned globally
+        "global_description": "Create a new namespace.",
+
+        # Category to group the permission in the UI.
+        "ui_category": _("Collection Namespaces"),
+
+        # Model permissions
+        "has_model_permission": True,
```
newswangerd, line 35: don't store `has_model_permission` here. This will get computed when the user requests the `me/` api

galaxy_ng/app/constants.py
```diff
@@ -16,3 +16,290 @@ class DeploymentMode(enum.Enum):
...
+        "global_description": "Create a new namespace.",
+
+        # Category to group the permission in the UI.
+        "ui_category": _("Collection Namespaces"),
+
+    },
+    "galaxy.change_namespace": {
+        "name": _("Change namespace"),
+
+        "object_description": "Edit this namespace.",
```
himdel, line 38: note all of these have to be translated (including object_description and global_description), not only name and ui_category

galaxy_ng/app/api/ui/serializers/user.py
```diff
@@ -171,68 +171,17 @@ class Meta(UserSerializer.Meta):
...
-
-        if not settings.get("SOCIAL_AUTH_KEYCLOAK_KEY"):
-            permissions["delete_user"] = obj.has_perm('galaxy.delete_user')
-            permissions["change_user"] = obj.has_perm('galaxy.change_user')
-            permissions["add_user"] = obj.has_perm('galaxy.add_user')
-            permissions["add_group"] = obj.has_perm('galaxy.add_group')
-            permissions["delete_group"] = obj.has_perm('galaxy.delete_group')
+        permissions = {}
+        for i, j in PERMISSIONS.items():
+            print(f'i & j: {i, j}')
```
newswangerd, line 176: remove the print

galaxy_ng/app/api/ui/serializers/user.py
```diff
@@ -171,68 +171,17 @@ class Meta(UserSerializer.Meta):
...
+            permissions[i] = j
+        for i in permissions:
```
newswangerd, line 178: you don't need to iterate through the permissions a second time to set these values. You can do that in the other loop while you're creating the `permissions` object.

galaxy_ng/app/api/ui/serializers/user.py
```diff
@@ -171,68 +171,17 @@ class Meta(UserSerializer.Meta):
...
+            permissions[i]["has_model_permission"] = obj.has_perm(i)
+        if settings.get("SOCIAL_AUTH_KEYCLOAK_KEY"):
+            permissions["galaxy.delete_user"] = False
```
newswangerd, line 181: This is wrong. You don't want to set `permissions[NAME]` to False. You want to set `permissions[NAME]['has_model_permission']` to False. Remember, this will overwrite the whole permission object, not just the has_model_permission boolean.

galaxy_ng/app/constants.py
```diff
@@ -16,3 +16,290 @@ class DeploymentMode(enum.Enum):
...
+        # Description to use when the permission is being assigned globally
+        "global_description": _("Create a new namespace."),
+
+        # Category to group the permission in the UI.
+        "ui_category": _("Collection Namespaces"),
+
+    },
+    "galaxy.change_namespace": {
+        "name": _("Change namespace"),
+
```
newswangerd, line 37: this is a nitpick, but can you remove the spaces between each of these? It makes the file a little hard to read.

galaxy_ng/app/constants.py
```diff
@@ -16,3 +16,290 @@ class DeploymentMode(enum.Enum):
...
+        "global_description": _("Delete any existing collection."),
+
+        "ui_category": _("Collections"),
+
+    },
+    "ansible.modify_ansible_repo_content": {
+        "name": _("Modify Ansible repo content"),
+
+        "object_description": _("Modify content of an Ansible repository."),
```
newswangerd, line 78: "this ansible repository"

galaxy_ng/app/constants.py
```diff
@@ -16,3 +16,290 @@ class DeploymentMode(enum.Enum):
...
+
+        "ui_category": _("Users"),
+
+    },
+    "galaxy.add_group": {
+        "name": _("Add group"),
+
+        "object_description": None,
+
+        "global_description": _("Create new groups to the system."),
```
newswangerd, line 130: "in the system"

galaxy_ng/app/constants.py
```diff
@@ -16,3 +16,290 @@ class DeploymentMode(enum.Enum):
...
+
+    },
+    "container.change_containernamespace": {
+        "name": _("Change container namespace permissions"),
+
+        "object_description": _("Edit permissions on this namespace."),
+
+        "global_description": _("Edit permissions on any existing namespace."),
+
+        "ui_category": _("Containers"),
```
newswangerd, line 192: Lets change the "Containers" category to "Execution Environments"

galaxy_ng/app/constants.py
```diff
@@ -16,3 +16,290 @@ class DeploymentMode(enum.Enum):
...
+
+        "global_description": _("Edit any container existing in the system."),
+
+        "ui_category": _("Containers"),
+
+    },
+      "container.namespace_modify_content_containerpushrepository" : {
+        "name": _("Change image tags"),
+
+        "object_description": _("Edit this image tag"),
```
newswangerd, line 208: "this image's"

galaxy_ng/app/constants.py
```diff
@@ -16,3 +16,290 @@ class DeploymentMode(enum.Enum):
...
+
+        "global_description": _("Edit permissions on any existing namespace."),
+
+        "ui_category": _("Containers"),
+
+    },
+    "container.namespace_change_containerdistribution": {
+        "name": _("Change containers"),
+
+        "object_description": _("Edit this container."),
```
newswangerd, line 198: All these `container.namespace_` permissions grant users the ability to modify all objects in the container namespace. This (and all the other permissions in this category) should say use "{ACTION} in this container namespace" for the object_description.

galaxy_ng/app/constants.py
```diff
@@ -16,3 +16,290 @@ class DeploymentMode(enum.Enum):
...
+
+    },
+    "galaxy.add_containerregistryremote": {
+        "name": _("Add remote registry"),
+
+        "object_description": None,
+
+        "global_description": _("Add remote registry to the system."),
+
+        "ui_category": _("Remote Registries"),
```
newswangerd, line 252: Change `Remote Registries` to `Container Registry Remotes`

newswangerd (changes requested):
Next step is to add some tests. I recommend adding unit tests here. Here's an example of some other tests that test the me endpoint: https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/tests/unit/api/test_api_ui_user_viewsets.py#L276

You can go ahead and add your tests to that file too. I'd recommend creating a user with the `galaxy.content_admin` role and then iterating through the permissions returned by the api and verifying that they match what we expect to see on the role (https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/access_control/statements/roles.py#L2)

galaxy_ng/app/constants.py
ShaiahWren: Do you have a preference for wording - "...this/any container namespace in the system" vs "..any existing container namespace" vs "...any existing container ns in the system" ? Seems like we could pick one for consistency.

newswangerd (commented):
let me know when this has tests and I'll take another look

galaxy_ng/tests/unit/api/test_api_ui_user_viewsets.py
```diff
@@ -372,6 +372,159 @@ def test_me_delete(self):
...
+        with self.settings(GALAXY_DEPLOYMENT_MODE=DeploymentMode.STANDALONE.value):
+            self._test_create_or_update(
+            self.client.put, self.me_url, new_user_data, status.HTTP_200_OK, user
+        )
+            client = APIClient(raise_request_exception=True)
+            client.force_authenticate(user=user)
+            response = self.client.get(self.me_url)
+            content_admin_permissions = LOCKED_ROLES["galaxy.content_admin"]["permissions"]
+            for i in content_admin_permissions:
+                self.assertEqual({response.data["model_permissions"][i]["has_model_permission"]}, {True})
```
newswangerd, line 404: These asserts also need to check that the responses we expect to be false are also false. Here's you're just checking the ones we expect to be true, which means that this would also pass if the API just returned true for all the permissions all the time.

newswangerd: also, there's no reason to put each condition in {}. You can either do `self.assertEqual(response.data["model_permissions"][i]["has_model_permission"], True)` or `self.asserTrue(response.data["model_permissions"][i]["has_model_permission"])`

newswangerd (changes requested):

galaxy_ng/tests/unit/api/test_api_ui_user_viewsets.py
ShaiahWren: The trouble is that `self.assertEqual(response.data["model_permissions"][i]["has_model_permission"], True)` never works - it gives the AssertionError: {True} != True, hence the brackets. `self.assertTrue(response.data["model_permissions"][i]["has_model_permission"])` does work okay.

SImilarly, `self.assertFalse({response.data["model_permissions"][i]["has_model_permission"]})` has the AssertionError: {False} is not false, so I had to go back to `self.assertEqual - {False}`.

galaxy_ng/tests/unit/api/test_api_ui_user_viewsets.py
```diff
@@ -372,6 +372,61 @@ def test_me_delete(self):
...
+            self.client.put, self.me_url, new_user_data, status.HTTP_200_OK, user
+        )
+            client = APIClient(raise_request_exception=True)
+            client.force_authenticate(user=user)
+            response = self.client.get(self.me_url)
+            content_admin_permissions = LOCKED_ROLES["galaxy.content_admin"]["permissions"]
+            for i in content_admin_permissions:
+                self.assertTrue({response.data["model_permissions"][i]["has_model_permission"]})
+            for i in excluded_model_permissions:
+                self.assertEqual({response.data["model_permissions"][i]["has_model_permission"]}, {False})
```
himdel, line 419:
```suggestion
                self.assertFalse(response.data["model_permissions"][i]["has_model_permission"])
```

galaxy_ng/tests/unit/api/test_api_ui_user_viewsets.py
```diff
@@ -372,6 +372,61 @@ def test_me_delete(self):
...
+        with self.settings(GALAXY_DEPLOYMENT_MODE=DeploymentMode.STANDALONE.value):
+            self._test_create_or_update(
+            self.client.put, self.me_url, new_user_data, status.HTTP_200_OK, user
+        )
+            client = APIClient(raise_request_exception=True)
+            client.force_authenticate(user=user)
+            response = self.client.get(self.me_url)
+            content_admin_permissions = LOCKED_ROLES["galaxy.content_admin"]["permissions"]
+            for i in content_admin_permissions:
+                self.assertTrue({response.data["model_permissions"][i]["has_model_permission"]})
```
himdel, line 417:
```suggestion
                self.assertTrue(response.data["model_permissions"][i]["has_model_permission"])
```

galaxy_ng/tests/unit/api/test_api_ui_user_viewsets.py
```diff
@@ -372,6 +372,61 @@ def test_me_delete(self):
...
+                self.assertEqual({response.data["model_permissions"][i]["has_model_permission"]}, {False})
+
+        with self.settings(GALAXY_DEPLOYMENT_MODE=DeploymentMode.INSIGHTS.value):
+            self.client.force_authenticate(user=user)
+            response = self.client.get(self.me_url)
+            content_admin_permissions = LOCKED_ROLES["galaxy.content_admin"]["permissions"]
+            for i in content_admin_permissions:
+                self.assertTrue({response.data["model_permissions"][i]["has_model_permission"]})
+            for i in excluded_model_permissions:
+                self.assertEqual({response.data["model_permissions"][i]["has_model_permission"]},  {False})
```
himdel, line 428:
```suggestion
                self.assertTrue(response.data["model_permissions"][i]["has_model_permission"])
            for i in excluded_model_permissions:
                self.assertFalse(response.data["model_permissions"][i]["has_model_permission"])
```

galaxy_ng/tests/unit/api/test_api_ui_user_viewsets.py
```diff
@@ -372,6 +372,61 @@ def test_me_delete(self):
...
+            self.client.put, self.me_url, new_user_data, status.HTTP_200_OK, user
+        )
+            client = APIClient(raise_request_exception=True)
+            client.force_authenticate(user=user)
+            response = self.client.get(self.me_url)
+            content_admin_permissions = LOCKED_ROLES["galaxy.content_admin"]["permissions"]
+            for i in content_admin_permissions:
+                self.assertTrue(response.data["model_permissions"][i]["has_model_permission"])
+            for i in excluded_model_permissions:
+                self.assertFalse(response.data["model_permissions"][i]["has_model_permission"])
```
newswangerd, line 419: Hardcoding the `excluded_model_permissions` is a bad idea. What if the permission change? You're also not checking all the permissions that the api returns.

This approach here iterates through all of the permissions that the API returns (so when new ones are added those will get tested too) and checks that we're only returning true for permissions that are in the content_admin role. `permission in content_admin_permissions` should return True if the permission is in the role and false if it's not.

```suggestion
            my_permissions = response.data["model_permissions"]
            for permission in my_permissions:
                self.assertEqual(my_permissions[permission]["has_model_permission'], permission in content_admin_permissions)
```

newswangerd (changes requested):

newswangerd (approved):

PR #1438, opened 2022-09-09
Fix a bug preventing SSO users from logging in with podman

newswangerd:
Issue: AAH-1921

Note: The reason we could never reproduce this in our environment was because we were setting the correct auth classes in our container, rather than relying on the default ones in settings.py. I've removed all of these settings from our .env files where they shouldn't be changed.

We don't have keycloak CI tests, so suggestions are welcome for how to add tests.

### DIFF
--- a/CHANGES/1921.bugfix
+++ b/CHANGES/1921.bugfix
@@ -1,0 +1,1 @@
+Fix a bug preventing keycloak SSO users from logging in to the container registry with podman/docker login.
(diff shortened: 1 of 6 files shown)


bmclaughlin:
Should probably be removed here https://github.com/ansible/galaxy_ng/blob/master/dev/standalone-rbac-roles/galaxy_ng.env#L4 as well.

awcrosby (commented):
Any benefit from adding a test to ensure `REST_FRAMEWORK__DEFAULT_AUTHENTICATION_CLASSES` gets populated from `settings.GALAXY_AUTHENTICATION_CLASSES` unless `data.GALAXY_AUTHENTICATION_CLASSES` is edited prior by a dynaconf hook? (automate the check that "This should go last...")

rochacbruno (approved):

bmclaughlin (approved):

PR #1453, opened 2022-09-16
Upgrade to pulp_ansible 0.15

newswangerd:
Issue: AAH-1780
Issue: AAH-1821

#### What is this PR doing:

- Upgrade to pulp ansible 0.15
- Update the set-repo-keyring management command to load public keys into the database instead of the gpg keyring filepath


galaxy_ng/app/api/ui/serializers/collection.py
```diff
@@ -65,7 +65,7 @@ def get_signatures(self, obj):
         data = []
         for signature in obj.signatures.all():
             sig = {}
-            sig["signature"] = bytes(signature.data).decode("utf-8")
+            sig["signature"] = signature.data
```
newswangerd, line 68: @mdellweg why did I have to make this change? Looking at the data coming from console dot, the signatures still show up as bytes, but when i sign a collection locally it shows up as a string.

newswangerd: Figured it out: https://github.com/pulp/pulp_ansible/pull/845

galaxy_ng/app/management/commands/set-repo-keyring.py
```diff
@@ -55,7 +56,7 @@ def handle(self, *args, **options):
 
         if not options["yes"]:
             confirm = input(
-                f"This will set keyring to {keyring} for {repository} repository, " "Proceed? (Y/n)"
+                f"This will set keyring to {keyring_path} for {repository} repository, " "Proceed? (Y/n)"
```
ShaiahWren: It's my understanding that `keyring` and `keyring_path` both indicate the string `/etc/pulp/certs/galaxy.kbx`

awcrosby (approved):

galaxy_ng/app/management/commands/set-repo-keyring.py
```diff
@@ -67,9 +68,15 @@ def handle(self, *args, **options):
...
+            "gpg", "--keyring", keyring_path, "--export", "-a"
+        ], capture_output=True)
+
+        pubkey = proc.stdout.decode().strip()
+
         task = dispatch(
-            set_repo_keyring,
-            kwargs={"repo_id": repo.pk, "keyring": keyring},
+            set_repo_gpgkey,
+            kwargs={"repo_id": repo.pk, "gpgkey": pubkey},
```
ShaiahWren, line 80: Do I understand correctly that this is setting the field, renamed "gpgkey," to the pubkey captured from the subprocess command?

galaxy_ng/app/management/commands/set-repo-keyring.py
```diff
@@ -83,14 +90,14 @@ def handle(self, *args, **options):
             self.echo(f"Task failed with error: {task.error}", self.style.ERROR)
             sys.exit(1)
 
-        if AnsibleRepository.objects.get(pk=repo.pk).keyring == keyring:
+        if AnsibleRepository.objects.get(pk=repo.pk).gpgkey == pubkey:
```
ShaiahWren, line 94: The field, now labeled "gpgkey," has an entry of the plain text public key, the output of filepath submitted to the subprocess command.

galaxy_ng/app/management/commands/set-repo-keyring.py
```diff
@@ -83,14 +90,14 @@ def handle(self, *args, **options):
...
             self.echo(f"Keyring {keyring} set successfully to {repository} repository")
 
 
-def set_repo_keyring(repo_id, keyring):
+def set_repo_gpgkey(repo_id, gpgkey):
```
ShaiahWren, line 98: (I assume renaming keyring to gpgkey here is just for clarity / semantics) - taking in that public key rather than a the filepath, then assigning the gpgkey field in the repo table to that public key.

ShaiahWren (commented):
Thanks for taking this over - I really appreciate it, and the opportunity to review and learn from your code.

galaxy_ng/app/api/ui/serializers/collection.py
mdellweg: We save the signature as ascii armored text. It was a pain to handle it as bytes. Also import-export couldn't cope with it at all.

galaxy_ng/app/management/commands/set-repo-keyring.py
```diff
@@ -67,9 +68,15 @@ def handle(self, *args, **options):
                     self.echo("Process canceled.")
                     return
 
+        proc = subprocess.run([
+            "gpg", "--keyring", keyring_path, "--export", "-a"
```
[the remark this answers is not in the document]
mdellweg: I'd suggest setting `--homedir` to a tmpdir to avoid any sideeffects from usersettings.

Also there might be a solution using `import gnupg; gnupg.GPG(gnupghome=tmpdir)`.

galaxy_ng/app/management/commands/set-repo-keyring.py
mdellweg: Exactly. However i would call the field replaced, not renamed. It expects the key instead of a path to the key.

galaxy_ng/app/management/commands/set-repo-keyring.py
mdellweg: Yes, it's a different string that is handled the same way.

mdellweg (commented):
Looks good to me so far.

PR #1455, opened 2022-09-19
Disable signatures in the v3 collection detail serializer

awcrosby:
#### What is this PR doing:
Temporary fix for AAH-1932.... disable signatures in the v3 serializer in order for signatures to not be synced from CRC to PAH. For any sync'd signatures, they will be removed after the next sync. This should resolve the PAH UI display issue. This will be the case until the AAH-1932 PAH fix is available to customers.

Issue: AAH-1937

#### Reviewers must know:

### DIFF
--- a/galaxy_ng/app/api/v3/urls.py
+++ b/galaxy_ng/app/api/v3/urls.py
@@ -81,6 +81,14 @@ urlpatterns = [
         name="collection-artifact-upload",
     ),
 
+    # This endpoint is defined here so we can temporarily override the serializer
+    # to avoid displaying signatures due to AAH-1932
+    path(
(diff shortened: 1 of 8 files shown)


galaxy_ng/app/api/v3/urls.py
```diff
@@ -81,6 +81,15 @@
         name="collection-artifact-upload",
     ),
 
+    # This endpoint is defined here so we can temporarily override the serializer
+    # to avoid displaying signatures due to AAH-1932
+    path(
```
newswangerd, line 86: Do you need to override the list serializer for collection versions too?

newswangerd (approved):

galaxy_ng/app/api/v3/urls.py
awcrosby: The list collection version list serializer does not include signatures, so i left that one alone

PR #1457, opened 2022-09-22
Replace the old insights proxy with a new one

newswangerd:
No-Issue

#### What is this PR doing:

Replaces the old insights proxy with a new one that's written in Go. There's not a good way to redirect a request in python, so the old proxy had a lot of brittle logic to handle lots of edge cases for proxying various methods and urls on the API, all of which can be removed in Go.


rochacbruno:
> There's not a good way to redirect a request in python, so the old proxy had a lot of brittle logic to handle lots of edge cases for proxying various methods and urls on the API, all of which can be removed in Go.

Can you provide more specific explanation about this? I am curious about what Go can do Better/Different than Python on that matter

newswangerd:
> Can you provide more specific explanation about this? I am curious about what Go can do Better/Different than Python on that matter

I couldn't find a good way to modify the original HTTP request and send it to a new URL with python. This basically what I want to do:

```python
def proxy(request)):
    reqeust = set_auth_header(request)
    response = send_request(request, new_url)
    return response
```

I couldn't find any libraries that have the same interfaces with for the `request` object coming from the server and the `request` object that's sent to the `send_request()` function. All of the cases require you to take the original request, parse the data, headers, etc. and create a new request object. Recreating a new request object and passing that along seems to cause a lot of incompatibility issues, which is why the original proxy had all of those `if` statements to handle the various edge cases created by the different methods. Ideally the new request should be as close to the original as possible, so that we can test the requests exactly as the client expects to be sent, rather than introducing differences at the proxy layer aside from those that are absolutely necessary (like sending the identity header).

It might be possible to do this using python's socket interface (https://www.codementor.io/@joaojonesventura/building-a-basic-http-server-from-scratch-in-python-1cedkg0842), but it's just hard to beat the convenience of Go's HTTP handler.

newswangerd:
The core of the go proxy does pretty much exactly what my python example says:

(this is truncated for ease of explanation)

```go
	reverseProxy := http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
		// request = set_auth_header(request)
		setRHIdentityHeader(req)

		// response = send_request(request, new_url)
		req.Host = urlToProxyTo.Host
		req.URL.Host = urlToProxyTo.Host
		req.URL.Scheme = urlToProxyTo.Scheme
		req.RequestURI = ""
		upstreamServerResponse, err := http.DefaultClient.Do(req)

		// return response
		data, _ := ioutil.ReadAll(upstreamServerResponse.Body)
		rw.Write(data)
	})
```

rochacbruno:
Maybe what you are doing with the Go proxy can be done with SimpleHTTPServer that is builtin

```py
import SocketServer
import SimpleHTTPServer
import urllib

PORT = 9097

class MyProxy(SimpleHTTPServer.SimpleHTTPRequestHandler):
    def do_GET(self):
        url=self.path[1:]  # remove the / begin slash
        self.send_response(200)
        self.end_headers()
        self.copyfile(urllib.urlopen(url), self.wfile)

httpd = SocketServer.ForkingTCPServer(('', PORT), MyProxy)
print ("Now serving at "    str(PORT))
httpd.serve_forever()
```

newswangerd:
I looked into `SimpleHTTPServer`. Do you have to have separate functions for `GET`, `POST`, etc?

newswangerd:
I don't think you can handle all the request types with one handler: https://docs.python.org/3/library/http.server.html#http.server.CGIHTTPRequestHandler.do_POST

rochacbruno:
There are the methods `handle` and `handle_one_request` where you can hook all request methods.

jctanner (approved):
I'm fine with this, as long as the rest of the team understands and agrees they will all be responsible for maintaining it.

PR #1470, opened 2022-10-03
Set CacheControl hdrs directives no-cache on login

alikins:
#### What is this PR doing:
The default CacheControl value is vague and unset, but set it specifically to no-cache;must-revalidate;no-store for responses from login view.

Issue: AAH-1323

#### Reviewers must know:
Original test case was the jira issue https://issues.redhat.com/browse/AAH-1277 , of which https://issues.redhat.com/browse/AAH-1323 is a child of.  There is some limited details
about the scanning tool and a link to the output.

Open questions: Does this break our ui login?

#### Notes: 

To test, get a galaxy_ng up and running, and hit the https://GALAXY_NG_SERVER/ui/ and attempt to log in. Record the response headers. For old versions, this should include a very limited CacheControl header.

With new version, repeat, but this time the response headers should look something like:
```
Cache-Control: max-age=0, no-cache, no-store, must-revalidate
```

Additional testing would be verifying that the user stays logged into the session, and nothing session/auth related happens.

### DIFF
--- a/galaxy_ng/app/api/ui/views/auth.py
+++ b/galaxy_ng/app/api/ui/views/auth.py
@@ -4,6 +4,7 @@ from django.conf import settings
 from django.contrib import auth as django_auth
 from django.utils.decorators import method_decorator
 from django.views.decorators.csrf import ensure_csrf_cookie
+from django.views.decorators.cache import patch_cache_control
 from rest_framework.authentication import SessionAuthentication
 
 from rest_framework.response import Response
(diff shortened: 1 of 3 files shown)


newswangerd (approved):

newswangerd:
lgtm. Please fix the tests.

alikins:
@newswangerd hmm, the failing tests make me wonder if this is going to break ui logout. But it may just be the way to tests are setup, I'll check.

alikins:
Hmm, seems odd that patch_cache_control would change the status code of the result in the test. I'll digg into it more.

newswangerd:
@alikins can you test this with the UI and double check that it still works there?

alikins:
Adding some tests for cache specific stuff to see if this behaves as intended.

galaxy_ng/app/api/ui/views/auth.py
```diff
@@ -40,7 +41,11 @@ class LoginView(api_base.GenericAPIView):
 
     @method_decorator(ensure_csrf_cookie)
     def get(self, request, *args, **kwargs):
-        return Response(status=http_code.HTTP_204_NO_CONTENT)
+        response = Response(status=http_code.HTTP_204_NO_CONTENT)
+        patch_cache_control(response, no_cache=True, no_store=True,
+                            must_revalidate=True)
+
+        return Response
```
alikins, line 48: gah... I'm dumb. Clearly this should be 'return response'

jctanner:
@alikins did you check the standalone UI logins with this?

jctanner (approved):

PR #1471, opened 2022-10-03
Set django CSRF_COOKIE_SECURE in settings.py

alikins:
### What is this PR doing:

This ensures cookies are only sent on secure / https connections.

Issue: AAH-1322
Related: AAH-1277

The django settings CSRF_COOKIE_SECURE is used by the django method 
`ensure_csrf_cookie` which is used by the login process. The CSRF_COOKIE_SECURE
will add an additional attribute to the SetCookie header 'secure'

#### Reviewers must know:
Tests are mostly comparing the response from logging into https://GALAXY_NG_SERVER/ui/ 

Before: No 'secure' attribute in SetCookie
After: There will be a 'secure' attribute in SetCookie

### DIFF
--- a/CHANGES/1322.bugfix
+++ b/CHANGES/1322.bugfix
@@ -1,0 +1,1 @@
+Add CSRF_COOKIE_SECURE setting to only set cookies on https connections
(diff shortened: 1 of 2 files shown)


awcrosby (changes requested):
It looks like there is a failure with the UIClient in the Standalone integration tests

newswangerd:
I'll defer to andrew to review this.

jctanner:
If cookies are only going to passed via https, the dev stack will need to support https.

alikins:
Some more info about this vulnerability type:  https://cwe.mitre.org/data/definitions/614.html

alikins:
Based on https://cwe.mitre.org/data/definitions/614.html I'm not sure this is needed. In use, connections to galaxy are always https (with the exception of development only setups that surface http)

awcrosby:
@alikins with galaxy only using https except for dev, this also supports that conclusion https://owasp.org/www-project-web-security-testing-guide/v41/4-Web_Application_Security_Testing/06-Session_Management_Testing/02-Testing_for_Cookies_Attributes.html
> If the application can be accessed over both HTTP and HTTPS, an attacker could be able to redirect the user to send their cookie as part of non-protected requests.

It sounds like this is not needed. The issue should be updated as Closed+Won't Do with a description of why that includes these external references.

alikins:
Closing this based on https://github.com/ansible/galaxy_ng/pull/1471#issuecomment-1275036299

PR #1483, opened 2022-10-05
Update tests to use ~/.ansible.cfg rather than /ansible.cfg

mikedep333:
This fixes the tests when used in the dev env.

No-Issue

### DIFF
--- a/.github/workflows/scripts/post_before_script.sh
+++ b/.github/workflows/scripts/post_before_script.sh
@@ -10,7 +10,3 @@ password password
 " > ~/.netrc
 
 chmod og-rw ~/.netrc
-
-# Needed for the tests that write this out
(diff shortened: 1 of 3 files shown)


mikedep333:
@dkliban fyi, if you were developing a workaround for this in oci_env, you no longer need to.

alikins:
Presumably the before case, the tests were looking for ansible.cfg in CWD.

As far as I can tell, that is normally https://github.com/ansible/galaxy_ng/blob/master/ansible.cfg

It looks like that only exists to setup ansible tmp_dirs for use with ansible-importer. If importer tests work I guess it doesn't matter ?

jctanner:
deprecated by https://github.com/ansible/galaxy_ng/pull/2305

PR #1492, opened 2022-10-11
Changelog entry for v1 addition

awcrosby:
Issue: AAH-1812

### DIFF
--- a/CHANGES/1812.feature
+++ b/CHANGES/1812.feature
@@ -1,1 +1,1 @@
-Add limited v1 support for legacy roles for hosted galaxy instance (tech preview)
+Add limited v1 support for legacy roles specific for use in galaxy.ansible.com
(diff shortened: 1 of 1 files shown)


awcrosby:
@newswangerd @jctanner i think we should have the v1 addition and its intent listed somewhere, perhaps either in changelog and/or docs

jctanner (approved):

CHANGES/1812.feature
```diff
@@ -0,0 +1 @@
+Add limited v1 support for legacy roles for hosted galaxy instance (tech preview)
```
newswangerd, line 1: I don't want customers to see this and think "great, we're getting role support for on prem". I would maybe change this to misc so it doesn't show up in the changelog unless you click on it, or maybe specifically say beta.galaxy.ansible.com in the notes (hosted galaxy instance may not be clear to people who don't know our project well)

newswangerd (changes requested):

CHANGES/1812.feature
awcrosby: Agreed, I want this to be clear this is for community galaxy. I updated the text.

newswangerd (approved):

PR #1502, opened 2022-10-13
Allow env to override test cases in RUN_INTEGRATION.sh

alikins:
#### What is this PR doing:

Allow the env variable TEST_CASES to be passed into env to override the default pattern or file path for finding test cases.

Let's dev or ci specify a more specific location of tests. For example, running a specific test module instead of everything in galaxy_ng/tests/integration.

   TEST_CASES=galaxy_ng/tests/integration/api/test_auth.py RUN_INTEGRATION.sh

No-Issue

Main intended use is to speed up local running of integration tests when dev only needs to run a subset of test cases.

``` shell
   TEST_CASES="galaxy_ng/tests/integration/api/test_pulp_api.py::test_pulp_roles_endpoint galaxy_ng/tests/integration/api/test_auth.py" HUB_LOCAL=1 ./dev/common/RUN_INTEGRATION.sh --log-level=DEBUG --exitfirst -r Pp
```

### DIFF
--- a/dev/common/RUN_INTEGRATION.sh
+++ b/dev/common/RUN_INTEGRATION.sh
@@ -23,6 +23,7 @@ which virtualenv || pip install --user virtualenv
 
 VENVPATH=/tmp/gng_testing
 PIP=${VENVPATH}/bin/pip
+TEST_CASES=${TEST_CASES:-galaxy_ng/tests/integration}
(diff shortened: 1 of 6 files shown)


jctanner:
obsolete with the change to oci-env

PR #1505, opened 2022-10-14
CI Cleanup.

jctanner:
#### What is this PR doing:
* Eliminate unnecessary spinup of the UI
* Move common RUN_INTEGRATION to standalone folder
* Add gettext in dev containers for translation builds
* Disable pulp telemetry in CI 

#### Reviewers must know:


alikins:
Could use a little more detail about in the commit messages.  It's not clear what is being cleaned up or why at first glance.

.github/workflows/ci_insights.yml
```diff
@@ -58,13 +52,7 @@ jobs:
...
 
       - name: build stack
-        run: make docker/build
-
-      - name: run migrations
-        run: make docker/migrate
-
-      - name: load test data
-        run: make docker/loaddata
+        run: make docker/all
```
alikins, line 55: I kind of like the idea of these steps being separate in the process (and more so in the action runs output). 

Is there an advantage to smashing it into one?  
Is it to avoid duplicating 'make docker/all' logic in multiple places?

dev/Dockerfile.base
```diff
@@ -32,6 +32,7 @@ RUN set -ex; \
 RUN set -ex; \
     touch /var/lib/rpm/* \
     && dnf -y install \
+        gettext \
```
alikins, line 35: What changed that makes gettext optional now?

dev/standalone/RUN_INTEGRATION.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
```
alikins, line 1: Could/should this be a separate pr?

.github/workflows/ci_insights.yml
jctanner: Because new developers already have too many things to learn without also having to remember what order to run the make targets that get the stack ready.

dev/standalone/RUN_INTEGRATION.sh
jctanner: dev/common/RUN_INTEGRATION.sh is going away and this continues the trend of having a script per compose profile.

dev/Dockerfile.base
jctanner: It's used when building django translations.

bmbouter:
Thank you for disabling telemetry! We only want end user systems to have it enabled.

awcrosby (approved):

PR #1515, opened 2022-10-20
Fix broken migration

newswangerd:
Issue: AAH-1994


awcrosby (approved):

ironfroggy (approved):
I was able to both reproduce the problem and verify the fix locally 👍

PR #1517, opened 2022-10-21
Update integration tests to be run against stage

chr-stian:
#### What is this PR doing:
Update integration tests to be run against stage. Here the Jenkins job running them:
https://jenkins-insights-qe-ci.apps.ocp4.prod.psi.redhat.com/job/galaxy_ng_integration_tests_stage/

New marker to skip tests that take too long in stage (slow_in_cloud). When tests call this method `get_all_repository_collection_versions` in stage, as there are thousands of collections the test takes too long.

partner_engineer profile is now used to upload artifacts because we don't have basic_user in [vault](https://vault.devshift.net/ui/vault/secrets/insights/list/secrets/qe/stage/users/)

Issue: [AAH-1953](https://issues.redhat.com/browse/AAH-1953)


jctanner (approved):
TY!

galaxy_ng/tests/integration/api/test_artifact_upload.py
```diff
@@ -57,7 +57,7 @@ def test_api_publish(ansible_config, artifact, upload_artifact, use_distribution
     which should indicate a successful import.
     """
 
-    config = ansible_config("basic_user")
+    config = ansible_config("partner_engineer")
```
awcrosby, line 60: I think we can use `basic_user` if we set it up  in stage env like here: https://github.com/ansible/galaxy_ng/blob/28ad4b525abdeb80a0c88a60b2dea9a205352c02/dev/common/setup_test_data.py#L88-L96

I see `basic_user` is tied to `aa_no_access` ... as long as that user has ansible entitlements, we can add their rh-identity group as one of the namespace owners for `autohubtest2` (which currently has 3 groups as namespace owners).

I don't see an `autohubtest3` namespace in stage, and I only see it used in one standalone-only test, so we can probably leave that as-is.

integration_stage_requirements.txt
```diff
@@ -0,0 +1,10 @@
+epdb
+requests
+#ansible-core
+ansible-base
```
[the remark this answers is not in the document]
jctanner: We need to check if ansible-core is working now. Last I tried there were issues with ignoring self-signed ssl certs or something like that with the way we invoked the library.

galaxy_ng/tests/integration/api/test_artifact_upload.py
jctanner: We should fix the tests so that these namespaces aren't assumed to pre-exist. The original IQE tests created them on the fly as part of client spinup or something like that, but we can now be more direct and simply create a randomized namespace for each test as needed instead of relying on static names.

dev/common/RUN_INTEGRATION_STAGE.sh
```diff
@@ -0,0 +1,24 @@
...
+export IQE_VAULT_ROLE_ID=${IQE_VAULT_ROLE_ID}
+export IQE_VAULT_SECRET_ID=${IQE_VAULT_SECRET_ID}
+export TESTS_AGAINST_STAGE=true
+export HUB_USE_MOVE_ENDPOINT=true
+export HUB_API_ROOT="https://console.stage.redhat.com/api/automation-hub/"
+export HUB_AUTH_URL="https://sso.stage.redhat.com/auth/realms/redhat-external/protocol/openid-connect/token/"
+
+mkdir -p "${HOME}/venvs"
+venv_path="${HOME}/venvs/ahub-tests-venv"
+python3 -m venv "${venv_path}"
```
alikins, line 17: Any particular reason this script puts the venvs in ${HOME} while common/RUN_INTEGRATION.sh and standalone/RUN_INTEGRATION.sh use /tmp?

The other RUN_INTEGRATION.sh's use virtualenv vs venv. Either should be fine, just wondering if there was a particular need that requires them to be different?  (and/or, should other RUN_INTEGRATION.sh scripts be updated?)

alikins (approved):
Still curious about the differences in RUN_INTEGRATION.sh but code is fine as is.

awcrosby (approved):

PR #1521, opened 2022-10-24
Fix AttributeError in index_registry task

alikins:
#### What is this PR doing:

The task was previously attempting to use a ValidationError from a 'serializers' import that didn't define it.

Use the rest_framework.exceptions.ValidationError instead.

Issue: AAH-1966

#### Reviewers must know:

To reproduce:

1. Bring a standalone env  (./compose up)
1. run integration tests ( HUB_LOCAL=1 ./dev/common/RUN_INTEGRATION.sh)
1. watch/grep the logs for above traceback  (there should be some before the fix, but none after)

### DIFF
--- a/CHANGES/1966.bugfix
+++ b/CHANGES/1966.bugfix
@@ -1,0 +1,1 @@
+Fix AttributeError in index_registry task when handing serializer errors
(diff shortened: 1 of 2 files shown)


jctanner:
This looks like it can be unit tested.

newswangerd (approved):

PR #1528, opened 2022-11-02
single req file for stage and non stage integration tests

chr-stian:
No-Issue

Use a single requirements file for both integration tests and stage integration tests


alikins (approved):
Looks reasonable to me. I like removing the extra dep file.

jctanner (approved):

PR #1543, opened 2022-11-17
Import iqe rbac tests

chr-stian:
Issue: AAH-2033

Import [RBAC tests from IQE repo](https://gitlab.cee.redhat.com/insights-qe/iqe-automation-hub-plugin/-/blob/master/iqe_automation_hub/tests/api/test_rbac.py). 

And some required tools too.
They are marked with this decorator `@pytest.mark.min_hub_version("4.6dev")` so that they get skipped if hub version is lower than 4.6


galaxy_ng/tests/integration/utils/iqe_utils.py
```diff
@@ -0,0 +1,193 @@
...
+                token = get_standalone_token(
+                    role, url, ignore_cache=ignore_cache, ssl_verify=ssl_verify
+                )  # ignore_cache=True
+                role.update(token=token)
+                auth = role
+
+            container_engine = config.get("container_engine")
+            container_registry = config.get("container_registry")
+
+            g_client = GalaxyClient(
```
awcrosby, line 135: It looks like this call to galaxykit is causing the pr-check.sh to fail with `KeyError: 'access_token'`

awcrosby (changes requested):
pr-check.sh failure

awcrosby:
The workflow `Standalone / integration` goes from 21min to 33min with this change. Maybe they should be in their own workflow with their own pytest mark (similar to https://github.com/ansible/galaxy_ng/blob/master/.github/workflows/ci_standalone-rbac-roles.yml and the `rbac_roles` mark). @chr-stian what do you think?

chr-stian:
> The workflow `Standalone / integration` goes from 21min to 33min with this change. Maybe they should be in their own workflow with their own pytest mark (similar to https://github.com/ansible/galaxy_ng/blob/master/.github/workflows/ci_standalone-rbac-roles.yml and the `rbac_roles` mark). @chr-stian what do you think?

Done

awcrosby (approved):

jctanner:
replaced by https://github.com/ansible/galaxy_ng/pull/1560

PR #1551, opened 2022-11-21
Add CI tests for CRC sync.

newswangerd:
Issue: AAH-1748


.github/workflows/ci_standalone.yml
```diff
@@ -24,15 +24,28 @@ jobs:
...
         run: pip3 install --upgrade docker-compose
 
+      - name: setup insights test instance
+        run: |
+          git clone https://github.com/newswangerd/oci_env.git --branch feature/db-snapshot $OCI_ENV_PATH
+          pip install -e $OCI_ENV_PATH/client/
+          mkdir $OCI_ENV_PATH/db_back/
+          cp dev/sync_tests/fixture.tar.gz $OCI_ENV_PATH/db_back/fixture.tar.gz
+          oci-env -e dev/sync_tests/sync-test.compose.env compose build
+          oci-env -e dev/sync_tests/sync-test.compose.env compose up
```
jctanner, line 47: Does this detach? I assumed with `-d` it would run forever and block any further tasks in the workflow.

newswangerd: @jctanner I'm still debugging the CI actions, so this file is actively being modified. Sorry, I didn't think it would ping you for reviews until I marked it as "not draft"

galaxy_ng/tests/integration/utils/sync.py
```diff
@@ -0,0 +1,82 @@
...
+    synclist = api_client(f"_ui/v1/my-synclists/{synclist_id}/")
+
+    synclist["collections"] = collections
+    api_client(f"_ui/v1/my-synclists/{synclist_id}/", method="PUT", args=synclist)
+
+    # the ansible-galaxy client doesn't know how to handle an empty response
+    try:
+        api_client(f"_ui/v1/my-synclists/{synclist_id}/curate/", method="POST")
+    except AnsibleError:
+        pass
```
awcrosby, line 18: We have removed the curate task and the need to call that endpoint

The UI can stop calling this endpoint (that may already be done), then we can remove this
https://github.com/ansible/galaxy_ng/blob/da51d7ea1cced3007ea81a91c2862be4817bfa8b/galaxy_ng/app/api/ui/viewsets/my_synclist.py#L33-L40

galaxy_ng/tests/integration/api/test_certified_sync.py
```diff
@@ -0,0 +1,180 @@
...
+    assert synced_collections == collections
+
+    # Test that the _ui/v1/repo/ api returns all the synced collections
+    synced_collections = set([(x["namespace"]["name"], x["name"]) for x in iterate_all(
+        client,
+        "_ui/v1/repo/rh-certified/"
+    )])
+    assert synced_collections == collections
+
+    # Test that the _ui/v1/repo/ api returns all the synced collections
```
awcrosby, line 67:
```suggestion
    # Test that the _ui/v1/repo/ api returns all the synced signed collections
```

CHANGES/1748.misc
```diff
@@ -0,0 +1 @@
+Add CI tests for testing sync from console.redhat.com.
```
awcrosby, line 1:
```suggestion
Add CI tests for testing standalone mode syncing against insights mode
```
(since we are not attempting to hit console.redhat.com hostname)

awcrosby (changes requested):
This is great! ... a couple small items left in comments.

jctanner (commented):
What is in dev/data/insights-fixture.tar.gz ? How do we maintain/update it?

awcrosby:
@jctanner some info in [dev/data/README.md](https://github.com/ansible/galaxy_ng/pull/1551/files#diff-87e2a71dc031422b0ae46ce518438d0b4f99555d176760014aee4c7f4bfb1fbf)

awcrosby (approved):
LGTM. I think after the lint items are fixed the rest of the CI jobs will run.

jctanner:
When I initially created the standalone workflow, it was with the intent that the basic development stack would be tested and broadly understood. I quite often point people at standalone_ci.yml as the proven/tested documentation about how to spin up the dev stack for dev/testing work. It seems as though this PR is attempting to change the stack from what the whole team is currently using to instead use the oci-env setup. Have we discussed that as a team and decided that's the path we're going down? I don't want to be in the same situation that existed when I first joined the team where it was poorly documented and generally inconsistent messages about how to spin up the stack to begin working on a PR or how to run any tests.

Can we pursue one of these options for this PR?

A) agree as a team that we want standalone testing to be a bespoke setup with oci-env (current state)
B) agree as a team that sync testing should be it's own bespoke workflow file and not alter the current standalone tests workflow.
C) agree as a team that we're going whole hog on oci-env and change ALL workflows to use it instead of the makefile/compose setups.

newswangerd:
> Can we pursue one of these options for this PR?
> 
> A) agree as a team that we want standalone testing to be a bespoke setup with oci-env (current state)
> B) agree as a team that sync testing should be it's own bespoke workflow file and not alter the current standalone tests workflow.
> C) agree as a team that we're going whole hog on oci-env and change ALL workflows to use it instead of the makefile/compose setups.

@jctanner these are all great points. I was initially going to wait to have the conversation about switching the whole environment to oci-env before changing over CI, but I couldn't get the old setup to work with the sync tests. There are some issues with container to container communication between the old env and the new one, and I don't really want to invest any time into fixing the old env to work for this use case, so I went ahead with the conversion now.

I'm obviously strongly in favor of C, but I can add this to our agenda for monday so we can discuss it with the whole team.

PR #1561, opened 2022-12-05
Sync tests against stage

chr-stian:
Add sync test against stage. This test is to be run by a Jenkins job. For example this one:
https://main-jenkins-csb-aap.apps.ocp-c1.prod.psi.redhat.com/job/AAPQA/job/Sandbox/job/christian/job/galaxy_ng_sync_test_stage/
or
https://main-jenkins-csb-aap.apps.ocp-c1.prod.psi.redhat.com/job/AAPQA/job/Sandbox/job/christian/job/sync_fedora/

More tests will be added in the future

Issue: [AAH-1270](https://issues.redhat.com/browse/AAH-1270)

### DIFF
--- a/galaxy_ng/tests/integration/api/test_sync_stage.py
+++ b/galaxy_ng/tests/integration/api/test_sync_stage.py
@@ -1,0 +1,78 @@
+import logging
+
+import pytest
+from galaxykit.collections import upload_artifact, delete_collection
+
+from ..conftest import get_galaxy_client, get_ansible_config_sync
+from ..utils import wait_for_task, get_client, set_certification
+from orionutils.generator import build_collection
+from ..utils.iqe_utils import is_sync_testing, get_all_collections, retrieve_collection
+from ..utils.tools import generate_random_artifact_version, uuid4
+
+pytestmark = pytest.mark.qa  # noqa: F821
+
+logger = logging.getLogger(__name__)
+
+
+def start_sync(api_client, repo):
+    logger.debug(f"Syncing {repo} repo")
+    url = f'content/{repo}/v3/sync/'
+    resp = api_client(url, method="POST")
+    resp = wait_for_task(api_client, resp["task"], raise_on_error=True)
+    logger.debug(f"Response from wait_for_task_id {resp}!")
+
+
+@pytest.mark.sync
+@pytest.mark.skipif(not is_sync_testing(), reason="This test can only be run on sync-tests mode")
+def test_sync():
+    config_sync = get_ansible_config_sync()
+    galaxy_client = get_galaxy_client(config_sync)
+    gc_remote = galaxy_client("remote_admin", remote=True)
+    user_stage = gc_remote.username
+    pass_stage = gc_remote.password
+    # upload artifact to stage
+    test_version = generate_random_artifact_version()
+    namespace_name = f"namespace_{uuid4()}"
+    namespace_name = namespace_name.replace("-", "")
+    gc_remote.create_namespace(namespace_name, "system:partner-engineers")
+
+    artifact = build_collection(
+        "skeleton",
+        config={"namespace": namespace_name, "version": test_version},
+    )
+    logger.debug(f"Uploading artifact name {artifact.name} version {artifact.version}")
+    resp = upload_artifact(None, gc_remote, artifact)
+    api_client_remote = get_client(
+        {"url": "https://console.stage.redhat.com/api/automation-hub/",
+         "username": user_stage, "password": pass_stage,
+         "use_move_endpoint": True, "upload_signatures": True},
+        request_token=True, require_auth=True)
+    resp = wait_for_task(api_client_remote, resp, raise_on_error=True)
+    assert resp["state"] == "completed"
+
+    set_certification(api_client_remote, artifact)
+
+    # sync and check
+    api_client = get_client({"url": "http://localhost:5001/api/automation-hub/",
+                             "username": "admin", "password": "admin"},
+                            request_token=True, require_auth=True)
+
+    body = {
+        "url": "https://console.stage.redhat.com/api/"
+               "automation-hub/content/1237261-synclist/",
+        "username": user_stage,
+        "password": pass_stage,
+        "tls_validation": False,
+        "signed_only": False,
+        "proxy_url": "http://squid.corp.redhat.com:3128",
+        "requirements_file":
+            f"---\ncollections:\n- {artifact.namespace}.{artifact.name}"
+    }
+    url = "content/community/v3/sync/config/"
+    api_client(url, method="PUT", args=body)
+
+    start_sync(api_client, 'community')
+    local_collections = get_all_collections(api_client, 'community')
+    assert retrieve_collection(artifact, local_collections)
+
+    delete_collection(gc_remote, artifact.namespace, artifact.name)
--- a/galaxy_ng/tests/integration/conftest.py
+++ b/galaxy_ng/tests/integration/conftest.py
@@ -20,7 +20,7 @@ from .utils import (
     set_certification,
 )
 from .utils import upload_artifact as _upload_artifact
-from .utils.iqe_utils import GalaxyKitClient, get_hub_version, is_stage_environment
+from .utils.iqe_utils import GalaxyKitClient, get_hub_version, is_stage_environment, is_sync_testing
 
 # from orionutils.generator import build_collection
 
@@ -61,6 +61,7 @@ slow_in_cloud: tests that take too long to be run against stage
 max_hub_version: This marker takes an argument that indicates the maximum hub version
 min_hub_version: This marker takes an argument that indicates the minimum hub version
 iqe_rbac_test: imported iqe tests checking role permissions
+sync: sync tests against stage
 """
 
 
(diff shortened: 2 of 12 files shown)


galaxy_ng/tests/integration/api/test_sync_stage.py
```diff
@@ -0,0 +1,78 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def start_sync(api_client, repo):
+    logger.debug(f"Syncing {repo} repo")
+    url = f'content/{repo}/v3/sync/'
+    resp = api_client(url, method="POST")
+    resp = wait_for_task(api_client, resp["task"], raise_on_error=True)
+    logger.debug(f"Response from wait_for_task_id {resp}!")
```
alikins, line 22: +1 for debug logging

galaxy_ng/tests/integration/utils/tasks.py
```diff
@@ -10,22 +10,32 @@
...
+                logger.error(resp["error"])
+                if raise_on_error:
+                    raise TaskFailed(resp["error"])
         except GalaxyError as e:
             if "500" not in str(e):
                 raise
         else:
             ready = resp["state"] not in ("running", "waiting")
-        time.sleep(SLEEP_SECONDS_POLLING)
+        time.sleep(5)
```
alikins, line 38: Maybe add a new constant for this case?

galaxy_ng/tests/integration/utils/tasks.py
```diff
@@ -10,22 +10,32 @@
     TaskWaitingTimeout
 )
 
+logger = logging.getLogger(__name__)
+
+
+def wait_for_task(api_client, task, timeout=300, raise_on_error=False):
+    if isinstance(task, dict):
```
alikins, line 17: Whats the reason for needing to support two different types for task?

galaxy_ng/tests/integration/conftest.py
```diff
@@ -317,7 +318,14 @@ def ansible_config():
 
 
 def get_ansible_config():
-    return AnsibleConfigFixture
+    if is_sync_testing():
```
alikins, line 321: I don't have a better suggestion, but this feels like it could get confusing pretty quickly.

Maybe stage_test_sync specific utils and/or a conftest.py with implementations of get_ansible_config() that just do the one thing.

alikins (commented):
Ok I suppose if it works. Kind of concerned about the growing complexity of the integration tests (in particular some of the methods that otherwise "do one thing" are now doing two or more things)

galaxy_ng/tests/integration/utils/tasks.py
chr-stian: right!

galaxy_ng/tests/integration/utils/tasks.py
chr-stian: there are endpoints that return a task_id (url = f"v3/tasks/{task}/") and others that return a URL that has to be appended to the base url to check the status

galaxy_ng/tests/integration/api/test_sync_stage.py
```diff
@@ -0,0 +1,78 @@
...
+        "skeleton",
+        config={"namespace": namespace_name, "version": test_version},
+    )
+    logger.debug(f"Uploading artifact name {artifact.name} version {artifact.version}")
+    resp = upload_artifact(None, gc_remote, artifact)
+    api_client_remote = get_client(
+        {"url": "https://console.stage.redhat.com/api/automation-hub/",
+         "username": user_stage, "password": pass_stage,
+         "use_move_endpoint": True, "upload_signatures": True},
+        request_token=True, require_auth=True)
```
awcrosby, line 49: what is the reason you need a `gc_remote` client and an `api_client_remote`? ... in other tests we just use one client

https://github.com/ansible/galaxy_ng/blob/2a5963fb658949a5f9ca8d9767a51e5e415b13f2/galaxy_ng/tests/integration/api/test_collection_signing.py#L68-L69

galaxy_ng/tests/integration/api/test_sync_stage.py
```diff
@@ -0,0 +1,78 @@
...
+        "proxy_url": "http://squid.corp.redhat.com:3128",
+        "requirements_file":
+            f"---\ncollections:\n- {artifact.namespace}.{artifact.name}"
+    }
+    url = "content/community/v3/sync/config/"
+    api_client(url, method="PUT", args=body)
+
+    start_sync(api_client, 'community')
+    local_collections = get_all_collections(api_client, 'community')
+    assert retrieve_collection(artifact, local_collections)
```
awcrosby, line 76: It seems this uploads 1 artifact to stage env, does a full sync of stage to a local galaxy_ng `master`, checks the 1 artifact was sync’d. Should we also assert the total collection count from stage matches the local total count?

awcrosby (commented):
Part of AAH-1270 is that we want multiple galaxy_ng versions to sync against stage: `master`, `4.4.z`, `4.5.z`, `4.6.z`. I think it is ok for these to come later, as long as this PR makes it straightforward to extend this test for Jenkins to use multiple galaxy_ng versions... @chr-stian do you think any changes are needed for that?

galaxy_ng/tests/integration/api/test_sync_stage.py
chr-stian: I tried to avoid that because stage is packed with collections and the test would take forever. While Calvin's script to clean stage is not ready, I did it like this. Is it what you mean?

galaxy_ng/tests/integration/api/test_sync_stage.py
chr-stian: because we should transition to using galaxykit (gc_remote) but there're methods or calls to the api that are still not available there so that's why I still need to use api_client_remote

chr-stian: @awcrosby we have these markers `@pytest.mark.min_hub_version("4.6dev")
` and `@pytest.mark.max_hub_version("...")`. So we can create specific tests for those versions and tag them and they will be run or skipped accordingly

galaxy_ng/tests/integration/api/test_sync_stage.py
awcrosby: ah i see now that the `requirements_file` is populated with only the one collection, is that why why you use the `community` repo/remote? (not sure if the `rh-certified` repo/remote allows that in the api, I know we don't surface it in the UI)

at some point i think we want to use the `rh-certified` repo/remote without a `requirements_file` to match how PAH users sync against CRC... maybe we can do that once stage env has Collections / CollectionVersions counts closer to prod env.

galaxy_ng/tests/integration/utils/tasks.py
awcrosby: Could this approach be less confusing than mixing types (while keeping backwards-compatible)?

```python
def wait_for_task(api_client, resp, task_id=None, timeout=300, raise_on_error=False):
    if task_id:
        url = f"v3/tasks/{task_id}/"
    else:
        url = url_safe_join(api_client.config["url"], resp["task"])
```

Then when you have a task_id call it like
```python
wait_for_task(api_client, resp=None, task_id=my_task_id, raise_on_error=True)
```

galaxy_ng/tests/integration/conftest.py
awcrosby: @chr-stian it seems this may not be needed since the stage tests explicitly call `get_ansible_config_sync` and not `get_ansible_config`?

galaxy_ng/tests/integration/api/test_sync_stage.py
awcrosby: That is cool we can limit tests to be run on different hub versions if needed. But my question was...

Say we merge a server-side change, and it gets deployed to c.stage.rh.c. We want to test if galaxy_ng 4.6.2 (and 4.5.z, 4.4.z) can continue to sync successfully against c.stage.rh.c.  

@chr-stian can we have Jenkins install 4.6.2 (and 4.5.z, 4.4.z), and then run this test? (with our without backporting this test into those branches)

galaxy_ng/tests/integration/conftest.py
chr-stian: here it's called, and it's a hook that's always called (sync and non sync tests)
https://github.com/chr-stian/galaxy_ng/blob/sync_tests_stage/galaxy_ng/tests/integration/conftest.py#L576

galaxy_ng/tests/integration/api/test_sync_stage.py
chr-stian: ah yes, we can. I defined the Jenkins job to not accept parameters because I thought that was what we needed but I can modify it to accept two branches: local hub branch and tests branch (if different). It's pretty straight-forward but it doesn't belong to this repo.

awcrosby: ok cool... and if we do that, can the test run from galaxy_ng `master` hitting galaxy_ng `4.6.2`... or does the test need to run from galaxy_ng `stable-4.6` hitting galaxy_ng `4.6.2`?

either way sounds like this can be done after this PR

chr-stian: right now tests are taken from the same branch as galaxy_ng has been started from. But I'll separate them so that we can have freedom to pick branches for hub and tests separately

awcrosby (approved):

PR #1580, opened 2022-12-23
Make sync test compatible with version 4.4

chr-stian:
signed_only param is only accepted in galaxy_ng versions higher than 4.4 
endpoint /content/community/v3/sync/config/

No-Issue


awcrosby (commented):
the pulp tests are failing, which should be unrelated to the integration tests and this PR... we'll likely look into that separately

galaxy_ng/tests/integration/api/test_sync_stage.py
```diff
@@ -69,6 +71,11 @@ def test_sync():
         "requirements_file":
             f"---\ncollections:\n- {artifact.namespace}.{artifact.name}"
     }
+
+    hub_version = get_hub_version(config_sync)
+    if parse_version(hub_version) < parse_version('4.5'):
+        del body["signed_only"]
+
```
alikins, line 78: This is fine as is so no change needed.

I might build 'body' dict without 'signed_only' key and then add it if version is >= 4.5. But that's just because 'del' often smells like a kluge so I try to avoid it. The 'del' usage here is correct and makes sense and is fine.

alikins (approved):
LGTM

galaxy_ng/tests/integration/api/test_sync_stage.py
hendersonreed: Right - the positive logic of "add this key if and only if we are in a version with signing" feels _slightly_ better intuitively to me, but it doesn't warrant changing I think.

ironfroggy: counter-point: the versions that support signing are the latest releases and `devel` so this inverted logic means the "main" code implements the actual version of the code the test is in (in its own repo) and the change only happens when talking to versions older than itself.

ironfroggy (approved):

PR #1586, opened 2023-01-05
Move oci-env+sync testing to a separate file.

jctanner:
#### What is this PR doing:
oci-env is not ready to replace our default dev stack.

* requires a secondary adjacent checkout of oci-env
* won't work with just a pip install
* doesn't conform to all of the workflows in our Makefile

### DIFF
--- a/.github/workflows/ci_standalone-certified-sync.yml
+++ b/.github/workflows/ci_standalone-certified-sync.yml
@@ -1,5 +1,5 @@
 ---
-name: Standalone Sync
+name: Standalone Certified Sync
 on: 
   pull_request:
     branches:
@@ -15,7 +15,7 @@ on:
 
 jobs:
 
-  sync_integration:
+  integration:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


.github/workflows/ci_standalone-certified-sync.yml
```diff
@@ -0,0 +1,67 @@
+---
+name: Standalone Certified Sync
```
awcrosby, line 2:
```suggestion
name: Sync local standalone against local insights
```

dev/standalone-certified-sync/RUN_INTEGRATION.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
+
+# Expected to be called by:
+# - GitHub Actions ci_standalone.yml for DeploymentMode.STANDALONE
+# - Developer Env makefile commands for DeploymentMode.STANDALONE
+# - TODO: Ephemeral Env pr_check.sh (merge smoke_test.sh into this) for DeploymentMode.INSIGHTS
```
awcrosby, line 6: think we can remove this block from this file

awcrosby (approved):
I think this makes sense... this change does not preclude us from switching to oci-env, but unblocks our workflows until we do.

PR #1599, opened 2023-01-23
Update to latest pulp releases

awcrosby:
#### What is this PR doing:

Update pulpcore from 3.21 to 3.22
Update pulp_container from 2.14.1 to 2.14.3

No-Issue


galaxy_ng/app/management/commands/set-repo-keyring.py
```diff
@@ -1,16 +1,20 @@
...
+import django_guid
 from django.conf import settings
 from django.core.management.base import BaseCommand
 from django.db import transaction
 from pulp_ansible.app.models import AnsibleRepository
 from pulpcore.plugin.constants import TASK_FINAL_STATES, TASK_STATES
 from pulpcore.plugin.tasking import dispatch
 
+# Set logging_uid, this does not seem to get generated when task called via management command
+django_guid.set_guid("cmd-set-repo-keyring")
```
rochacbruno, line 16: maybe we can set this on `../__init__.py` so it applies to all the commands

rochacbruno: On a quick research I see that this is needed only because the field in pulpcore is required right?
it would be better if that field had a default value, or if pulp set this guid thing inside the dispatch of a task.

awcrosby: The default value was just removed in pulpcore 3.22 https://github.com/pulp/pulpcore/pull/3258/files#diff-b6921076eabe69f8f887286bbac04f5abb8c91b5e2e747b99ea932c5abf17135L179

In task `dispatch` it is retrieved via `get_guid` but not set... perhaps because the logger is expected to set it https://github.com/pulp/pulpcore/blob/main/pulpcore/tasking/tasks.py#L91

Maybe `dispatch` needs something like

```
if not get_guid():
    set_guid(generate_guid())
```

awcrosby: I opened https://github.com/pulp/pulpcore/issues/3528 to try to do this on pulpcore, but it looks like the intention is to force the caller to set the guid.

newswangerd (approved):

galaxy_ng/app/management/commands/set-repo-keyring.py
```diff
@@ -1,16 +1,20 @@
...
+import django_guid
 from django.conf import settings
 from django.core.management.base import BaseCommand
 from django.db import transaction
 from pulp_ansible.app.models import AnsibleRepository
 from pulpcore.plugin.constants import TASK_FINAL_STATES, TASK_STATES
 from pulpcore.plugin.tasking import dispatch
 
+# Set logging_uid, this does not seem to get generated when task called via management command
+django_guid.set_guid(django_guid.utils.generate_guid)
```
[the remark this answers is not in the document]
awcrosby: thanks for the catch!

PR #1618, opened 2023-02-16
feat(api): API endpoint for AI Index Deny List

rochacbruno:
Add AIIndexDenyList database model to store scope and reference for namespaces to be ignored from AI Index (a.k.a Wisdom).

Issue: AAH-2127

To enable the feature the feature flag `ai_deny_index` must be set to true

`on the env file`
```bash
PULP_GALAXY_FEATURE_FLAGS__ai_deny_index=true
```

## Feature

Adds 3 endpoints to manage the AIIndex toggle
urls

**scope** path parameter options are **namespace,legacy_namespace**

**reference** parameter is either a namespace or a legacy_namespace name

> **NOTE** in future we might add **user** scope with some user identifier as reference

### Adding a namespace to the DenyList

Requirements:
- referenced namespace must exist
- authenticated user must have permission to change the namespace

```bash
curl -X 'POST' \
  'http://0.0.0.0:8002/api/automation-hub/_ui/v1/ai_deny_index/namespace/' \
  -H 'Authorization: Basic YWRtaW46YWRtaW4=' \
  -H 'Content-Type: application/json' \
  -d '{"reference": "test"}'
```
api_home_add

[spec](http://0.0.0.0:8002/api/automation-hub/v3/swagger-ui/#/Api%3A%20_Ui%20V1%20Ai_Index/api_automation_hub__ui_v1_ai_index_ai-index-add)

### Removing from DenyList

Requirements:
- referenced namespace must exist
- authenticated user must have permission to change the namespace

```bash
curl -X 'DELETE' \
  'http://0.0.0.0:8002/api/automation-hub/_ui/v1/ai_deny_index/namespace/test/' \
  -H 'accept: application/json' \
  -H 'Authorization: Basic YWRtaW46YWRtaW4=' \
```
api_home_delete
[spec](http://0.0.0.0:8002/api/automation-hub/v3/swagger-ui/#/Api%3A%20_Ui%20V1%20Ai_Index/api_automation_hub__ui_v1_ai_index_ai-index-delete)

### Viewing the list

Think of it as a [robots.txt](https://moz.com/learn/seo/robotstxt) for the AIIndex(wisdom), it is a simple list of namespaces
that the AI scan will look after before including the namespace collections on its database.

Requirements:
- This endpoint is open to read only without authentication

```bash
curl -X 'GET' \
  'http://0.0.0.0:8002/api/automation-hub/_ui/v1/ai_deny_index/' \
  -H 'accept: application/json'
```
```json
{
  "results": [
    {
      "scope": "namespace",
      "reference": "autohubtest2"
    },
    {
      "scope": "namespace",
      "reference": "autohubtest3"
    },
    {
      "scope": "namespace",
      "reference": "auuo6un85"
    },
    {
      "scope": "namespace",
      "reference": "c28wkzhm"
    },
    {
      "scope": "namespace",
      "reference": "dv3y8j0ps03"
    },
    {
      "scope": "namespace",
      "reference": "oynp9v33gkp1"
    },
    {
      "scope": "namespace",
      "reference": "yxn07bd6"
    }
  ],
  "count": 7
}
```
api_home_list
[spec](http://0.0.0.0:8002/api/automation-hub/v3/swagger-ui/#/Api%3A%20_Ui%20V1%20Ai_Index/api_automation_hub__ui_v1_ai_index_ai-index-list)

This endpoint optionally takes a `?scope=namespace|legacy_namespace` query parameter to perform filtering.

## TODO

Required

- [x] Tests
- [x] Logging / Audit


galaxy_ng/app/api/ui/views/ai_index.py
```diff
@@ -0,0 +1,180 @@
...
+        Query params:
+            scope: "namespace" or "legacy_namespace" to filter
+
+        Sorted by reference field.
+        """
+        qs = AIIndexDenyList.objects.all().order_by("reference")
+        if scope := request.GET.get("scope"):
+            qs = qs.filter(scope=scope)
+        items = qs.values_list("reference", flat=True)
+        return Response({"items": list(items)})
```
rochacbruno, line 141: This is returning

```json
{
    "items": [
        "foo",
        "bar", 
        "zaz"
    ]
}
```

And accepts a `?scope=` filter parameter, I think this is easier wherever it is consumed to parse, I am not sure if we want to serialize everything like:

```json
{
    "items": [
        {"scope": "namespace", "reference": "foo"},
        ...
    ]
}
```

@jctanner @awcrosby

jctanner:
I didn't think we were supposed to add more endpoints to api/_ui? Should this go into v3?

galaxy_ng/app/api/ui/views/ai_index.py
```diff
@@ -0,0 +1,180 @@
...
+                "reference": "some_name"
+            }
+
+        responses:
+            201: Created
+            400: Bad Request (invalid payload data)
+            409: Conflict (data already in the model)
+        """
+        data = {"scope": scope, "reference": request.data.get("reference")}
+        obj = AIIndexDenyList(**data)
```
jctanner, line 86: Should we attempt to cast this across both namespace and legacynamespace if they have matching names or if legacynamespace has a foreignkey relationship with the namespace?

galaxy_ng/app/models/aiindex.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+class AIIndexDenyList(models.Model):
+    """Stores the list of content that are not allowed to be added into AI Index/Wisdom.
+
+    Data from this model is exposed on the /ai_index/ endpoint and any namespace added
+    to the table is opt-out from ai scanning.
+
+    The reference field is a namespace name in case of namespace scope, and a
+    legacy role namespave name in case of legacy_namespace scope.
```
jctanner, line 22: legacy role namespa**c**e

himdel:
Looks like this is not letting us filter by name? Given a thousand namespaces, will this return the whole list, or will it try paginating the result?

elyezer:
This is probably a nitpick but wouldn't `ai_deny` make more sense for the path? `ai_index` gives an impression that whatever is in there is on the index, not removed from it.

galaxy_ng/app/api/ui/views/ai_index.py
rochacbruno: I am not sure, if namespace and legacy_namespace have matching names, is it owned by the same owners?
How is the relationship between legacy and namespace?

Should I also add this here? I can duplicate the entry

rochacbruno:
@elyezer thanks for the idea, we can rename the endpoint to be more explicit like `/ai_deny_index/`

galaxy_ng/app/models/aiindex.py
```diff
@@ -0,0 +1,32 @@
+from django.core.validators import RegexValidator
+from django.db import models
+
+SCOPES = (
+    ("namespace", "Namespace"),
+    ("legacy_namespace", "Legacy Namespace"),
+)
+
+
+namespace_validator = RegexValidator(
```
awcrosby, line 10: should probably add a comment since Namespace and LegacyNamespace use different regex patterns, maybe:
```
# this regex is the most restrictive pattern that applies to both Namespace and LegacyNamespace
```

galaxy_ng/tests/integration/api/test_aiindex.py
```diff
@@ -0,0 +1,111 @@
...
+    8. Assert permission error raises
+    """
+    with UIClient(config=ansible_config("admin")) as client:
+        # 2. Add namespace to AIIndex
+        assert (
+            client.post("_ui/v1/ai_index/namespace/", payload={"reference": namespace}).status_code
+            == 201
+        )
+
+        # 3. Assert namespace is listed  on _ui/v1/ai_index/
```
awcrosby, line 66: maybe do the GETs as a `ansible_config("basic_user")`

galaxy_ng/tests/integration/api/test_aiindex.py
```diff
@@ -0,0 +1,111 @@
...
+        assert (
+            uclient.post("_ui/v1/ai_index/namespace/", payload={"reference": namespace}).status_code
+            == 403
+        )
+
+    with UIClient(config=ansible_config("partner_engineer")) as uclient:
+        # 9. add to the AI Index, a namespace owned by PE
+        assert (
+            uclient.post(
+                "_ui/v1/ai_index/namespace/", payload={"reference": pe_namespace}
```
awcrosby, line 103: Do you think we should have a separate test for `scope:legacy_namespace`? since we may need to test slightly differently given the `partner_engineer` applies only to collections

docs/config/featureflags.md
```diff
@@ -0,0 +1,60 @@
+# Feature Flags
+
+Some features are disabled by default. You can enable them by setting the
+feature flag under `GALAXY_FEATURE_FLAGS` setting.
+
+example for enabling EEs endpoints:
```
elyezer, line 6: s/example/Example?

docs/config/featureflags.md
```diff
@@ -0,0 +1,60 @@
...
+    - boolean
+    - Tells UI to show the execution environments tab and enable EE endpoints
+    - default: False
+- `container_signing`
+    - boolean
+    - Tells UI to show the container signing buttons and badges
+    - default: True when Container Signing Service is configured
+- `ai_deny_index`
+    - boolean
+    - Enables AI Deny Index endpoints and tells UI to display the Wisdom opt-put buttons
```
elyezer, line 59: s/opt-put buttons/opt-out button?

elyezer (commented):
Left some suggestions around the featureflags doc

galaxy_ng/app/access_control/statements/standalone.py
```diff
@@ -256,6 +256,22 @@
     ],
 
     'NamespaceViewSet': [
+        {
+            "action": ["ai-deny-index-list"],
```
awcrosby, line 260: should we add an insights mode integration test to ensure these urls give a 404?

himdel: Is this not supposed to work on insights as well?

If so, is there a feature flag specific to the feature, or do we gate any UI changes based on deployment mode?

rochacbruno: without the `ai_deny_index` feature flag the endpoints will raise 404

galaxy_ng/app/access_control/access_policy.py
```diff
@@ -214,6 +214,26 @@ def has_distribution_repo_perms(self, request, view, action, permission):
         return False
 
 
+class AIDenyIndexAccessPolicy(AccessPolicyBase):
```
awcrosby, line 217: cool, this feels better being in its own access policy

galaxy_ng/app/migrations/0034_aiindexdenylist.py
```diff
@@ -0,0 +1,25 @@
+# Generated by Django 3.2.18 on 2023-02-15 13:52
+
+import django.core.validators
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('galaxy', '0033_update_validated_repo'),
```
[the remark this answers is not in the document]
rochacbruno: I used what django recommended `--merge` I can do this way also.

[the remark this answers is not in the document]
awcrosby: yeah, just to limit it to 1 migration file per pr for readability

awcrosby (approved):

rochacbruno:
@MilanPospisil found a problem and I am thinking in ways to resolve it.

> When I delete the namespace, the namespace still stays in the deny list, thus when recreated again, it carries the previous settings, which may not be good, because those two namespaces may be unrelated. 

So, to solve this problem I am thinking about adding 2 signals to https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/signals/handlers.py that would listen to the post_delete on Namespace and LegacyNamespace to perform the appropriate exclusion from aidenyindex, WDYT?

@awcrosby @jctanner

PR #1620, opened 2023-02-21
Switch to using pulp dev branches

newswangerd:
No-Issue


setup.py
```diff
@@ -112,8 +112,8 @@ def _format_pulp_requirement(plugin, specifier=None, ref=None, gh_namespace="pul
 
 requirements = [
     "galaxy-importer>=0.4.6,<0.5.0",
-    "pulpcore>=3.22.1,<3.23.0",
-    "pulp_ansible>=0.16.0,<0.17.0",
+    "pulpcore@git+https://git@github.com/pulp/pulpcore.git@main#egg=pulpcore",
+    "pulp_ansible@git+https://git@github.com/pulp/pulp_ansible.git@main#egg=pulp_ansible",
```
mdellweg, line 116: Do these contradict the main branches? Depending on scm locations and installing from source is not quite the same in pip.
Can you try without this change?

newswangerd: @mdellweg what else do you recommend doing? I'd rather not use `pulp_ansible>2.X.Xdev` because then it can't find anything to install from pypi and and I don't want to make it so that the UI devs have to check out a bunch of pulp checkouts for source installs

rochacbruno: when `LOCK_REQUIREMENTS` variable is set, the setup.py would transform `pulpcore>=...` to just `pulpcore` without version pin, is that not working?

rochacbruno: This is what I mean:
(this code will take `pulpcore==x.y.z` and make it just `pulpcore` as long it is added to `DEV_SOURCE_PATH list.

https://github.com/ansible/galaxy_ng/blob/4a796b72b657d5172ae81d523d6eb17819ef2e2e/setup.py#L139-L160

rochacbruno: @newswangerd nm, now I see that you are trying to avoid having to locally clone, so I think your code is the only option we have

template_config.yml
```diff
@@ -5,7 +5,7 @@
 
 additional_repos:
 - bindings: true
-  branch: 0.16.0
+  branch: main
```
awcrosby, line 8: While we use pulp* `main`, will we need to occasionally run the `plugin-template` script and requirements `make` cmd to ensure we bring in the latest template changes and requirement updates?

rochacbruno (approved):

template_config.yml
newswangerd: It should always bring in the main branch. I just fixed something in pulp ansible that was breaking ci and I didn't have to update anything on the pr other than rerunning the tests

awcrosby (approved):

PR #1621, opened 2023-02-22
Add legacy role filter for tag(s) & reduce default page size to 10

jctanner:
https://issues.redhat.com/browse/AAH-2154
https://issues.redhat.com/browse/AAH-2085

### DIFF
--- a/galaxy_ng/app/api/v1/filtersets.py
+++ b/galaxy_ng/app/api/v1/filtersets.py
@@ -1,4 +1,7 @@
-from django.db.models import Q
+import json
+
+from django.db.models import F, Q, Value
+from django.db.models.functions import Concat
 from django_filters import filters
 from django_filters.rest_framework import filterset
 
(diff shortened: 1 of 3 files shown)


galaxy_ng/tests/integration/api/test_community.py
```diff
@@ -524,3 +524,62 @@ def get_roles(page_size=1, order_by='created'):
 
     # cleanup
     clean_all_roles(ansible_config)
+
+
+@pytest.mark.community_only
+def test_v1_role_tag_filter(ansible_config):
+    """" Tests if v1 roles are auto-sorted by created """
+
+    config = ansible_config("admin")
```
awcrosby, line 533: if just testing search, can probably test with a non admin profile

galaxy_ng/app/api/v1/filtersets.py
```diff
@@ -57,6 +57,8 @@ class LegacyRoleFilter(filterset.FilterSet):
 
     github_user = filters.CharFilter(method='github_user_filter')
     keywords = filters.CharFilter(method='keywords_filter')
+    tags = filters.CharFilter(method='tags_filter')
+    tag = filters.CharFilter(method='tags_filter')
```
awcrosby, line 61: i see in the integration test that `tag` is a query param... is `tags` used by the UI also?

jctanner: The UI uses "tags", but it doesn't make contextual sense if you are filtering on a single tag. We could extend "tags" to mean a comma separated list, but there's no convention or ask for that yet.

galaxy_ng/tests/integration/api/test_community.py
jctanner: Sync requires an admin to run.

awcrosby (approved):

PR #1624, opened 2023-02-23
Replace galaxy_ng Namespace with pulp_ansible Namespace (viewsets, serializers, not models yet)

alikins:
#### What is this PR doing:
The models, viewsets, and serializers for Namespace objects is moving into pulp_ansible.
This pr starts removing Namespace from galaxy_ng.

Issue: AAH-313

#### Reviewers must know:
This is a partial removal of Namespace related things. 

Note that this doesn't remove any Namespace models.

Since we are going to need to do a data migration of galaxy_ng Namespaces to
pulp_ansible AnsibleNamespaces via a mgmt command. And that mgmt command
needs to run before the galaxy_ng Namespaces models and objects are removed.


galaxy_ng/app/access_control/access_policy.py
```diff
@@ -218,10 +218,6 @@ class AppRootAccessPolicy(AccessPolicyBase):
     NAME = "AppRootViewSet"
 
 
-class NamespaceAccessPolicy(AccessPolicyBase):
-    NAME = "NamespaceViewSet"
```
newswangerd, line 222: We need to apply our existing access policies to the new pulp namespace

newswangerd (commented):
This is hard to really review since nothing is working, but it's broadly the right idea.

We need to remove the existing namespace APIs and replace them with the one's available in pulp ansible. This will unfortunately come with some breaking changes, but I don't know that there's much we can do about that. Keep in mind that the new APIs will have to:

- provide the exact same rbac experience (requiring namespace perms to upload, etc)
- provide enough functionality for the UI to do it's thing

galaxy_ng/app/access_control/access_policy.py
alikins: If the pulp_ansible namespace apis provide an equivalent but different and incompatible API, almost anything using them currently will need to be updated.

Do we want to provide any sort of compatibility layer ?

I've been working on the expectation we are attempting to avoid implementing any new API in galaxy_ng. (ie, no subclassing of pulp_ansible viewsets etc). Is that still a valid assumption?

PR #1631, opened 2023-03-02
Add labels to pulp repos

newswangerd:
Issue: AAH-805


galaxy_ng/app/migrations/0036_repository_labels.py
```diff
@@ -0,0 +1,50 @@
...
+def add_repository_labels(apps, schema_editor):
+    """Add labels to repositories."""
+
+    # Add hide_from_search label to staging, rejected, inbound and synclist repos
+    # Add pipeline=staging,rejected,approved to staging, rejected and published repos
+    AnsibleRepository = apps.get_model('ansible', 'AnsibleRepository')
+    AnsibleDistribution = apps.get_model('ansible', 'AnsibleDistribution')
+
+    for repo in AnsibleRepository.objects.filter(
+        Q(name__startswith="inbound-") | Q(name__startswith="inbound-")):
```
rochacbruno, line 17: is this needed to query for the same field and value twice?

galaxy_ng/app/migrations/0036_repository_labels.py
```diff
@@ -0,0 +1,50 @@
...
+
+        repo.pulp_labels = {"hide_from_search": ""}
+        repo.save()
```
mdellweg, line 20: Want to move the loop inside Postgres?
```suggestion
    AnsibleRepository.objects.filter(
        Q(name__startswith="inbound-") | Q(name__startswith="inbound-")
    ).update(pulp_labels={"hide_from_search": ""})
```

If you needed to be cautious with existing labels, we'd need to add some more hstore magic. ;)

newswangerd: Good suggestion.

We don't use any existing labels, so resetting them all is fine

rochacbruno (approved):

PR #1635, opened 2023-03-08
Repo management approval workflow

newswangerd:
#### What is this PR doing:
- allow users to create custom approved and staging repositories
- update auto approval to move approved content into all approved repos
- refactor and fix rbac tests

NOTE: need galaxykit release with this pr for iqe tests to pass: https://github.com/ansible/galaxykit/pull/85

Issue: AAH-2170


galaxy_ng/app/api/v3/viewsets/collection.py
```diff
@@ -257,55 +256,15 @@ def copy_content(self, request, *args, **kwargs):
...
-
-        # collection deprecation
-        if 'ansible.collection_deprecation' in content_types:
-            deprecations_pks = AnsibleCollectionDeprecated.objects.filter(
-                pk__in=source_pks,
-                namespace=collection_version.namespace,
-                name=collection_version.name,
-            ).values_list("pk", flat=True)
-            if deprecations_pks:
-                content.append(*deprecations_pks)
```
rochacbruno, line 292: append() takes one argument, how did I miss that in the previous review 🤦🏻

galaxy_ng/app/tasks/promotion.py
```diff
@@ -1,92 +1,161 @@
...
+    content = [cv_pk]
+
+    # collection signatures
+    if 'ansible.collection_signature' in content_types:
+        signatures_pks = CollectionVersionSignature.objects.filter(
+            signed_collection=cv_pk,
+            pk__in=source_pks
+        ).values_list("pk", flat=True)
+        if signatures_pks:
+            content.append(*signatures_pks)
```
rochacbruno, line 116: This will not work I guess

```python
In [4]: content = []

In [5]: t = [1,2,3]

In [6]: content.append(*t)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Input In [6], in <cell line: 1>()
----> 1 content.append(*t)

TypeError: append() takes exactly one argument (3 given)

```

Must use extend method

```suggestion
            content.extend(signatures_pks)
```

rochacbruno (changes requested):

newswangerd:
Gonna refactor this after https://github.com/pulp/pulp_ansible/pull/1393 is merged

galaxy_ng/app/api/v3/viewsets/collection.py
jerabekjiri: ooops, sorry :facepalm: :face_with_spiral_eyes:

galaxy_ng/app/access_control/access_policy.py
```diff
@@ -213,6 +243,56 @@ def has_distribution_repo_perms(self, request, view, action, permission):
 
         return False
 
+    def signatures_not_required_for_repo(self, request, view, action):
+        """
+        Validate that collections are being added with signatures to approved repos
+        when signatures are required.
+        """
+        print("______--------____------__------__-----__--------")
```
rochacbruno, line 251: Is this print needed?

galaxy_ng/app/access_control/access_policy.py
```diff
@@ -213,6 +243,56 @@ def has_distribution_repo_perms(self, request, view, action, permission):
...
+        repo = view.get_object()
+        repo_version = repo.latest_version()
+
+        src_pipeline = repo.pulp_labels.get("pipeline", None)
+
+        if settings.GALAXY_REQUIRE_SIGNATURE_FOR_APPROVAL:
+            if src_pipeline == "approved":
+                return True
+        else:
+            return True
```
rochacbruno, line 261: I don't understand this block, both branches returns True, are those redundant?
logically the same

```suggestion
        if settings.GALAXY_REQUIRE_SIGNATURE_FOR_APPROVAL:
            return True
```

also, why is it returning True when that setting is set? no need for the extra checks below?

newswangerd: I was thinking to skip the check if you're copying from an approved repo, because those should theoretically have signatures already, but I guess it makes more sense to just check the destinations.

galaxy_ng/app/access_control/access_policy.py
```diff
@@ -213,6 +243,50 @@ def has_distribution_repo_perms(self, request, view, action, permission):
...
+        for cv in data["collection_versions"]:
+            sig_exists = repo_version.get_content(
+                ansible_models.CollectionVersionSignature.objects
+            ).filter(signed_collection=cv).exists()
+
+            if not sig_exists:
+                raise ValidationError(detail={"collection_versions": _(
+                    "Signatures are required in order to add collections into any 'approved'"
+                    "repository when GALAXY_REQUIRE_SIGNATURE_FOR_APPROVAL is enabled."
+                )})
```
rochacbruno, line 286: on 254 it checks for `settings.GALAXY_REQUIRE_SIGNATURE_FOR_APPROVAL` and early returns `True` 
but in line 283 it raises error when that setting is enables, is that intended?

newswangerd: that's wrong... why didn't my tests catch this...

galaxy_ng/app/tasks/promotion.py
```diff
@@ -1,92 +1,88 @@
...
+    add_and_remove(
+        src_repo_pk,
+        add_content_units=[cv_pk],
+        remove_content_units=[],
+    )
+
+    try:
+        signing_service = AUTO_SIGN and SigningService.objects.get(name=SIGNING_SERVICE_NAME)
+    except SigningService.DoesNotExist:
+        raise RuntimeError(_('Signing %s service not found') % SIGNING_SERVICE_NAME)
```
rochacbruno, line 30: I think it is better to translate only user facing messages (UI,API) and keep logs and errors untranslated so it is better to find when troubleshooting.

```suggestion
        raise RuntimeError('Signing %s service not found' % SIGNING_SERVICE_NAME)
```

rochacbruno (approved):

PR #1639, opened 2023-03-13
Wisdom-aa-collector

dmzoneill:
#### What is this PR doing:
Weekly export of collection information to s3

Issue: AAH-1624
https://issues.redhat.com/browse/AA-1624


galaxy_ng/app/management/commands/analytics/galaxy_collector.py
```diff
@@ -0,0 +1,208 @@
...
+
+
+
+
+
+
+
+
+
+
```
rochacbruno, line 176: too many blank lines

rochacbruno:
The text `Issue: AAH-1624` must be part of the commit message
You also need to add a `CHANGES/1624.feature` with one liner for the changelog

setup.py
```diff
@@ -122,6 +122,9 @@ def _format_pulp_requirement(plugin, specifier=None, ref=None, gh_namespace="pul
     "social-auth-app-django>=3.1.0,<4.0.0",
     "dynaconf>=3.1.12",
     "django-auth-ldap==4.0.0",
+    "distro",
+    "boto3",
+    "insights_analytics_collector@git+https://git@github.com/RedHatInsights/insights-analytics-collector.git@main#egg=insights_analytics_collector"
```
Ladas, line 127: will you need new release of the insights_analytics_collector? Talk to Martin

Ladas:
This looks good, we can comment out the "SELECT * FROM<table_name>", lets add them back as we'll work on Cloud Hub and Private Hub collection and we'll have these as populated.

setup.py
```diff
@@ -122,6 +122,9 @@ def _format_pulp_requirement(plugin, specifier=None, ref=None, gh_namespace="pul
     "social-auth-app-django>=3.1.0,<4.0.0",
     "dynaconf>=3.1.12",
     "django-auth-ldap==4.0.0",
+    "distro",
+    "boto3",
```
awcrosby, line 126: @dmzoneill turns out we already have these 2 dependencies in our galaxy-ng env used for beta-galaxy... these are pulled in via pulp and the way we use storages. You can remove them from here.

```
(venv) sh-4.4$ pip freeze | fgrep distro
distro==1.8.0
(venv) sh-4.4$ pip freeze | fgrep boto3
boto3==1.26.104
```

setup.py
awcrosby: yes, any dependency here will need to come from pypi rather than github prior to merge

@dmzoneill we'll also want a dependency range specified here, something like `>=1.2.3,<2.0.0`

CHANGES/1639-misc
```diff
@@ -0,0 +1 @@
+Add exporter command for project wisdom
```
awcrosby, line 1: @dmzoneill this PR will need a link to a Jira ticket in the AAH project. You can either move `AA-1624` to AAH or I can help create an AAH ticket linked to the AA one, just let me know.

Then this file can be renamed `####.misc` to match the AAH ticket number.

You'll also need to rebase this PR so it picks up our recent changes, which may include CI checks.

awcrosby (changes requested):
I added a couple comments that can help get some of the CI items passing and match how we use dependencies.

galaxy_ng/app/management/commands/analytics/galaxy_collector.py
```diff
@@ -0,0 +1,220 @@
...
+        "redis_connection": "todo",
+        "storage": "todo",
+    }
+
+
+@register("ansible_collection_table", "1.0", format="csv", description="Data on ansible_collection")
+def ansible_collection_table(since, full_path, until, **kwargs):
+    source_query = """
+        COPY (
+            SELECT "ansible_collection"."pulp_id",
```
Ladas, line 61: @dmzoneill  talking with @matburt and @awcrosby , can we try to combine the models with sql here?

Something like "COPY ({model_query.sql}) TO STDOUT WITH CSV HEADER". Not sure if django models support copy natively.

The queries here are fairly simple, so using models makes sense.

awcrosby: @Ladas it sounds like the ORM queries are not going to work well. And upon further thought, they might not give us many benefits in the event of a schema change. It does seem like most of these queries hit mainly one table with few joins, so maybe schema changes will not often impact these queries. So IMO the manual queries should be sufficient.

Ladas: @awcrosby yeah, in Controller we have raw queries and we had to do maybe 1 or 2 changes last couple of years, due to schema drift.

So yeah it doesn't give us many benefits. Especially if we are not able to do all queries and we'd have a mix of ORM/SQL here then. It would be good to use ORM, if we'd be exporting data that have a lot of business logic in models, but that doesn't seem to be the case. (and the business logic in models would slow this down a lot, so we'd rather replicate in DB I guess...)

setup.py
dmzoneill: added dependency range

setup.py
```diff
@@ -122,6 +122,8 @@ def _format_pulp_requirement(plugin, specifier=None, ref=None, gh_namespace="pul
     "social-auth-app-django>=3.1.0,<4.0.0",
     "dynaconf>=3.1.12",
     "django-auth-ldap==4.0.0",
+    "distro",
+    "insights_analytics_collector@git+https://git@github.com/RedHatInsights/insights-analytics-collector.git@main#egg=insights_analytics_collector"
```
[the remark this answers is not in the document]
dmzoneill: updated setup.py to reference pypi

galaxy_ng/tests/unit/app/management/commands/test_analytics_export_s3.py
```diff
@@ -0,0 +1,55 @@
...
+        assert os.getenv("aws_access_key_id") == "blah"
+        assert os.getenv("aws_secret_access_key") == "blah"
+        assert os.getenv("aws_region") == "blah"
+        assert os.getenv("aws_bucket") == "blah"
+
+        call_command('analytics-export-s3', stdout=out)
+
+        # self.assertTrue(mock.called)
+        mock_file.assert_called_with(ANY)
+        self.assertIn("Completed successfully")
```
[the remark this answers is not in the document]
dmzoneill: mmm, i'm only doing the unit test here.  Would that not be part of an integration test?
@Ladas

galaxy_ng/tests/unit/app/management/commands/test_analytics_export_s3.py
```diff
@@ -0,0 +1,55 @@
...
+
+    @patch("builtins.open", new_callable=mock_open, read_data="data")
+    @patch('boto3.client')
+    @patch.dict(os.environ, s3_details, clear=True) 
+    def test_write_file_to_s3_failure(self, boto3, mock_file):
```
awcrosby, line 43: it looks like the contents of `test_write_file_to_s3_failure` is the same as `test_write_file_to_s3_success`

dmzoneill: removed

requirements/requirements.common.txt
```diff
@@ -293,9 +310,14 @@ pyyaml==6.0
     #   yamllint
 redis==4.5.1
     # via pulpcore
+<<<<<<< HEAD
 requests==2.28.2
+=======
+requests==2.30.0
+>>>>>>> 1411f3a (Wisdom-aa-collector)
```
awcrosby, line 317: this is breaking most ci jobs

CHANGES/1639.misc
```diff
@@ -0,0 +1 @@
+Add exporter command for project wisdom
```
awcrosby, line 1: since this PR is linked to issue AAH-1624, this filename will need to be renamed to `1624.misc` to pass ci

awcrosby (changes requested):

CHANGES/1639.misc
awcrosby: sorry, i read `Issue: AAH-1624` in the PR description and i thought that was the issue key, but i see from the commit message `Issue: AAH-2343` that is the correct issue key. so you'll want 1 file in `CHANGES` called `2343.misc`

awcrosby:
3 of our CI jobs are not running, im not sure why. Maybe it was because I had to approve the workflows to run. I just changed the setting so I should not need to approve the next run, hopefully those 3 CI job run on the next push.

jctanner (approved):

PR #1640, opened 2023-03-15
Fix repo modify permissions.

newswangerd:
No-Issue


himdel (approved):
LGTM, should fix editing repositories :+1:

PR #1647, opened 2023-03-22
Refactor accesss policy loading

newswangerd:
- Insights access policy now inherits from standalone to avoid duplication
- All pulp access policies have been grouped together under statements/pulp.py
- Container access policies are no longer loaded from the database
- Setting queryset scoping and creation hooks is now enabled on pulp access policies


galaxy_ng/app/access_control/statements/insights.py
```diff
@@ -471,13 +138,11 @@
             "action": "create",
             "principal": "authenticated",
             "effect": "allow",
-            "condition": "has_model_perms:galaxy.change_group"
         },
         {
             "action": "destroy",
             "principal": "authenticated",
             "effect": "allow",
-            "condition": "has_model_perms:galaxy.change_group"
```
awcrosby, line 480: do we need to remove these conditions?

newswangerd: I didn't mean to do that

awcrosby:
Overall looks good after tests pass. I want to do a quick check of standalone vs the items deleted from insights, i will do that this afternoon and then i will approve.

galaxy_ng/app/access_control/statements/insights.py
```diff
@@ -385,70 +70,53 @@
...
-    "DistributionViewSet": _deny_all,
     "MyDistributionViewSet": [
         {
             "action": ["list", "retrieve"],
             "principal": "authenticated",
             "effect": "allow",
-            "condition": "has_rh_entitlements",
         },
     ],
     "ContainerRepositoryViewSet": _deny_all,
```
awcrosby, line 122: We have a deny on ContainerRepositoryViewSet... the current insights.py does not have deny on ContainerRegistryRemoteViewSet, ContainerRemoteViewSet, and ContainerReadmeViewset. But since we are inheriting standalone which has some things allowed for authenticated users, we probably should put deny on them.

awcrosby: I think we can also deny `LegacyAccessPolicy` and `pulp_ansible/v3/legacy-redirected-viewset` since they also allow access for all authenticated users. But otherwise the diff looks good.

jctanner (approved):

PR #1649, opened 2023-03-23
Custom querysets for private repos

bmclaughlin:
#### What is this PR doing:

Creates custom querysets when user LIST/RETRIEVE AnsibleRepository/AnsibleDistribution api endpoints.
Users should be able to see all repositories or distributions unless the (associated) repository is marked private.

Issue: AAH-2000

#### Reviewers must know:

### DIFF
--- a/galaxy_ng/app/access_control/access_policy.py
+++ b/galaxy_ng/app/access_control/access_policy.py
@@ -152,25 +152,6 @@ class AccessPolicyBase(AccessPolicyFromDB):
             }
         )
 
-    def _get_public_repo_pks(self):
-        return ansible_models.AnsibleRepository.objects.filter(private=False)
-
-    def _get_private_repo_pks(self, user, perm):
-        return get_objects_for_user(
-            user,
-            perm,
-            ansible_models.AnsibleRepository.objects.filter(private=True),
-        )
-
-    def get_ansible_repository_qs(self, view, qs, repo_perm):
-        qs = ansible_models.AnsibleRepository.objects.all()
-        if view.request.user.has_perm(repo_perm):
-            return qs
-        else:
-            public_repository_qs = self._get_public_repo_pks()
-            private_repository_qs = self._get_private_repo_pks(view.request.user, repo_perm)
-            return public_repository_qs | private_repository_qs
-
     def scope_by_view_repository_permissions(self, view, qs, field_name="", is_generic=True):
         """
         Returns objects with a repository foreign key that are connected to a public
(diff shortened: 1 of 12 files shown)


galaxy_ng/app/access_control/access_policy.py
```diff
@@ -146,8 +150,66 @@ def get_access_policy(cls, view):
...
+        )
+
+    def get_ansible_repository_qs(self, view, qs, repo_perm):
+        qs = ansible_models.AnsibleRepository.objects.all()
+        if view.request.user.has_perm(repo_perm):
+            return qs
+        else:
+            public_repository_pks = self._get_public_repo_pks()
+            private_repository_pks = self._get_private_repo_pks(view.request.user, repo_perm)
+            return qs.filter(Q(pk__in=public_repository_pks) | Q(pk__in=private_repository_pks))
```
newswangerd, line 170: I think you can merge the querysets rather than doing an exensive `pk__in`: https://github.com/pulp/pulp_container/blob/a003472be52e456114387d9ce6efe39605c1370c/pulp_container/app/viewsets.py/#L1275-L1276

galaxy_ng/app/access_control/access_policy.py
```diff
@@ -146,8 +150,66 @@ def get_access_policy(cls, view):
...
+        return ansible_models.AnsibleRepository.objects.filter(private=False)
+
+    def _get_private_repo_pks(self, user, perm):
+        return get_objects_for_user(
+            user,
+            perm,
+            ansible_models.AnsibleRepository.objects.filter(private=True),
+        )
+
+    def get_ansible_repository_qs(self, view, qs, repo_perm):
```
newswangerd, line 163: I think we should look into using annotations for the queryset. You could annotate each queryset with an attribute called "has_view_permission" that would return true when the user has the view permission on that object. Then you can just return `Ansible.repository.filter(Q(private=False) ! Q(has_view_permission=True))`, but I think that might take some work to get right

galaxy_ng/app/access_control/access_policy.py
newswangerd: I think this should work for your other querysets too

bmclaughlin: I'm getting an error when doing this in `get_ansible_distribution_qs`.  2 out of 3 ain't bad. 

```AssertionError: Cannot combine queries on two different base models.```

newswangerd (approved):

PR #1659, opened 2023-03-29
Update PO files

newswangerd:
No-Issue

PR #1683, opened 2023-04-19
Update translations automatically  

jerabekjiri:
Issue: AAH-2265

Automate running `django-admin makemessages --all`, ` django-admin compilemessages` and committing translation files weekly.
         

Solution taken from https://github.com/ansible/ansible-hub-ui/blob/master/.github/workflows/i18n.yml

### DIFF
--- a/.github/workflows/i18n.yml
+++ b/.github/workflows/i18n.yml
@@ -1,0 +1,62 @@
+---
+name: Update translations
+
+on:
+  # allow running manually
+  workflow_dispatch:
+  # run weekly
+  schedule:
+  - cron: '30 5 * * 0'
+
+jobs:
+  i18n:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        branch:
+        - 'master'
+        - 'stable-4.4'
+        - 'stable-4.5'
+        - 'stable-4.6'
+        - 'stable-4.7'
+
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ matrix.branch }}
+
+      - name: Set up Python
+        uses: actions/setup-python@v4
+
+      - name: Update apt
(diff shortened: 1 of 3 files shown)


himdel:
Looks good! :)

2 potential issues:

* django makemessages will try to match new strings and add fuzzy translations .. which will cause linter issues when the untranslated text does interpolation but the "matching" translation doesn't, or vice versa
  * pretty much what happened in https://github.com/ansible/galaxy_ng/pull/1659
  * I guess the easy version is just to run the perl script from there before compiling the updated translations (after makemessages but before compilemessages)
  * alternatively, `msgattrib --clear-fuzzy --empty -o /path/to/output.po /path/to/input.po` (from the `gettext` apt package) should *probably* do the same thing (https://code.djangoproject.com/ticket/10852, https://stackoverflow.com/questions/7372414/removing-all-fuzzy-entries-of-a-po-file)

* git push *might* trigger some branch protection rules here and fail, I know we had to change ui repo settings for this to work .. but merging is probably the only way to know for sure :)

jerabekjiri:
@himdel I thought of fuzzy matching, weird that it doesn't generate/add any fuzzy strings when running  `django-admin makemessages` locally. I'll test it more

himdel:
Try adding new strings, it only does that if the string is not already in the pot file. (And it has to be similar enough to an existing string.)

jerabekjiri:
> Try adding new strings, it only does that if the string is not already in the pot file. (And it has to be similar enough to an existing string.) .. or you can try reverting https://github.com/ansible/galaxy_ng/pull/1659 locally and trying again

reverting worked, thanks :) 

Added `msgattrib --clear-fuzzy` script for every language folder

himdel (approved):

jerabekjiri:
Merging this, so it can be tested.

PR #1695, opened 2023-05-01
Update community userguide with new information.

jctanner:


awcrosby (approved):
text updates LGTM... could we have collection section come first at the top then roles?

jctanner:
We could, but i think that is a different PR.

PR #1701, opened 2023-05-03
[WIP] Add a test for upload concurrency.

jctanner:
WIP


kdelee:
@kaushiki29 might want to take a look at this

galaxy_ng/app/api/v3/viewsets/collection.py
```diff
@@ -62,9 +62,19 @@ def _dispatch_upload_collection_task(self, args=None, kwargs=None, repository=No
         task_group = TaskGroup.objects.create(description=f"Import collection to {repository.name}")
 
         if settings.GALAXY_REQUIRE_CONTENT_APPROVAL:
-            return dispatch(import_to_staging, kwargs=kwargs, task_group=task_group)
+            return dispatch(
+                import_to_staging,
+                kwargs=kwargs,
+                task_group=task_group,
+                exclusive_resources=[repository]
```
newswangerd, line 69: Unless I'm misunderstanding something, it looks like this is going to undo the whole https://issues.redhat.com/browse/AAH-2104 fix. The point of AAH-2104 was to split the import task into two jobs:

1. Run galaxy importer (time consuming, doesn't require a repo lock)
2. Move the content into the desired repo (fast, requires a repo lock)

If the galaxy importer step does actually require a repository lock for some reason, then that needs to be fixed, rather than just locking the repository on the import stage. Otherwise, galaxy will only be able to run the import task on a single collection at a time, even if all the other task workers are free. With the importer step taking several seconds to a minute for more complex collections, that will cause pretty serious bottlenecks on galaxy.ansible.com, especially since locking the repository will also prevent any other operations on it (signing, approval, etc).

newswangerd (approved):

PR #1705, opened 2023-05-05
Add compatibility with new pulp namespaces APIs

newswangerd:
- [ ] Migrate permissions to pulp namespaces
- [ ] Disable galaxy namespace API
- [ ] Remove namespace viewsets/serializers

Issue: AAH-2293


ShaiahWren (commented):
Could you add additional filter `or__name__icontains` on the global ns api and `or__repository_name__icontains` on the namespace metadata search viewset api? This will prevent the 404 api error when querying multiple filter tags.

PR #1706, opened 2023-05-05
Attempt to fix docker/urllib problem by using compose v2

rochacbruno:
No-Issue

We where seeing CI errors due to https://github.com/docker/docker-py/issues/3113

Problem:

When we do "pip install docker-compose" we are installing unsopported version of docker-compose, which comes with docker-py, which depends on request, which depends on a broken version of urllib3.

Solution:

Start using docker compose v2

New problem, container names are different in compose v2

This PR fixes both problems


slonopotamus:
That issue is already fixed in docker-py 6.1.0.

rochacbruno:
thanks @slonopotamus but in any case we have to upgrade the docker compose used in CI

jctanner (approved):

PR #1717, opened 2023-05-10
Fix publishing collection to specified repository when content approval is set to false

jerabekjiri:
Issue: AAH-2328


jerabekjiri:
@jctanner Good idea, we have integration tests with auto_approval, but they get skipped in CI. So if the community decorator runs with the approval off, I can add it there.

newswangerd (approved):

PR #1721, opened 2023-05-16
Lets use galaxy ng tests everywhere

chr-stian:
Integration tests have been adapted so that Jenkins AAP integration pipelines can run them
Issue: AAH-2341


awcrosby:
@chr-stian this PR introduces 3 new `ci.int.devshift.net` pr_check.sh failures (6 total failures, 3 should get fixed with #1720).

PR #1733, opened 2023-05-23
increase wait timeouts

chr-stian:
No-Issue


awcrosby (approved):
LGTM, this will hopefully help address very slow tasks due to AAH-2337

PR #1752, opened 2023-05-31
Pushing updated strings for localization

rooftopcellist:
#### What is this PR doing:

* Updates translation strings for es and nl now that they are done.  

> Note: on the previous PR, I probably should have just omitted the changes for es, nl.  Turns out those were just the Machine translations and the humans weren't done translating.  In the future, the machine translations will only be used if there is a 100% match.


himdel (approved):

rooftopcellist:
Feel free to merge if you are OK with this change (I don't have permission).  Thanks!

PR #1756, opened 2023-06-02
Fix stage tests

chr-stian:


chr-stian:
this PR is to fix stage tests but CI failed so I needed to update these tests:
galaxy_ng/tests/integration/api/test_cross_repository_search.py
please, let me know if this is wrong. I don't understand why test_cross_repository_search failed [here](https://github.com/ansible/galaxy_ng/actions/runs/5147059443/jobs/9270077645) but not [here](https://github.com/ansible/galaxy_ng/actions/runs/5148104636/jobs/9269364716)
I locally reproduced the failure and saw that the test was expecting an artifact to be in staging, but it was in published. But the same passed in a different PR that didn't touch anything related to this.

awcrosby (approved):

PR #1766, opened 2023-06-08
integration test import role when owner != github user

EamonTracey:
No-Issue


EamonTracey:
This test ensures that an owner of a namespace can import into that namespace even if the owner's name differs from the namespace/github_user name.

PR #1769, opened 2023-06-12
Bool or str

chr-stian:
the feature-flags endpoint return true or 'True' for execution_environments depending on the AAP operator on openshift

```
AAP 2.3, galaxy_ng 4.6.5
GET /api/galaxy/_ui/v1/feature-flags/
{
    "legacy_roles": false,
    "execution_environments": "True",
    ...
}
AAP 2.4 galaxy_ng 4.7.1
GET /api/galaxy/_ui/v1/feature-flags/

{
    "legacy_roles": false,
    "execution_environments": true,
    ...
}
```

with this change, we cover both cases and make the tests pass

### DIFF
--- a/galaxy_ng/tests/integration/api/test_ui_paths.py
+++ b/galaxy_ng/tests/integration/api/test_ui_paths.py
@@ -332,7 +332,7 @@ def test_api_ui_v1_feature_flags(ansible_config):
         validate_json(instance=ds, schema=schema_featureflags)
 
         # assert ds['ai_deny_index'] is False
-        assert ds['execution_environments'] in [True or 'True']
+        assert ds['execution_environments'] in [True, 'True']
         assert ds['legacy_roles'] is False
 
 
--- a/galaxy_ng/tests/integration/api/test_ui_paths.py
+++ b/galaxy_ng/tests/integration/api/test_ui_paths.py
@@ -332,7 +332,7 @@ def test_api_ui_v1_feature_flags(ansible_config):
         validate_json(instance=ds, schema=schema_featureflags)
 
         # assert ds['ai_deny_index'] is False
-        assert ds['execution_environments'] is True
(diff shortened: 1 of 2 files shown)


chr-stian:
I don't know why this happens and why this changed (it depends on the AAP version). I have asked multiple times and I got no answer. I didn't want to insist too much because it looks very low priority but makes our tests fail, that's why I put up this PR. I'll update it with your proposal (even though I'm afraid it involves too much logic: it depends on the AAP version and only if hub has been installed from the operator)

rochacbruno:
This is caused because installer is setting the boolean value as a string, it is probably outputing `'True'` to the /etc/pulp/settings.py

Dynaconf parses all the values using TOML parser, this parser is used to allow type inference for environment variables and other flat files.


```py
# Set to 'True' and get a string
In [10]: settings.set("FOO", "True", tomlfy=True)
In [11]: settings.FOO
Out[11]: 'True'
In [12]: type(settings.FOO)
Out[12]: str

# Set to 'true' (lowercase) and get a boolean
In [13]: settings.set("FOO", "true", tomlfy=True)
In [14]: settings.FOO
Out[14]: True
In [15]: type(settings.FOO)
Out[15]: bool

# or force a boolean
In [16]: settings.set("FOO", "@bool 0", tomlfy=True)
In [19]: settings.set("FOO", "@bool off", tomlfy=True)

In [20]: settings.FOO
Out[20]: False
In [21]: type(settings.FOO)
Out[21]: bool
```

rochacbruno:
> Is this the desired behavior or is a stringified bool actually a bug? If it's something we can't fix, then i think we should have a duplicate test for the older version that allows the string so we can eventually deprecate it..

The stringified boolean is a bug, I would not make this workaround on hub, better to try to solve it on the installer (or wherever the setting is being set)

chr-stian:
Closing it. We are not doing this.

PR #1770, opened 2023-06-13
Switch testing over to oci-env

newswangerd:
#### What is this PR doing:
- simplifying the process of running integration tests locally
  - `make oci-env/integration` will run the integration tests against your currently running dev environment (as specified by your `oci-env/compose.env` file). This command will automatically run the default set of tests against your current environment and should automatically run the tests that are expected to pass for your environment.
  - `make gh-action/<ACTION>` this will spin up a new environment and run the tests for that environment in the same way that github actions runs the test. This is useful for debugging actions that are failing on CI, as it allows for a reproducible environment to be easily instantiated locally
- enable all of the integration test to run against keycloak and ldap profiles
- overhaul the test data setup to work with keycloak and ldap
- consolidate the integration test marks under categories for specific deployment parameters
- switch github actions over to run using oci-env

Issue: AAH-2393


newswangerd:
Replacing with #1773

PR #1772, opened 2023-06-14
copy_expert is deprecated in django 4.x + psycopg

jctanner:
https://issues.redhat.com/browse/AAH-2408

We need a way to validate this code without having s3 in the mix, so there's a new command to just copy the generated tarball to a new folder in /tmp. We can later look at adding minio to the compose stack and have the s3 based command copy there, but we'll need to revise the Package code to allow for a different s3 host url ... but that's beyond the scope of this particular PR.

### DIFF
--- a/galaxy_ng/app/management/commands/analytics/galaxy_collector.py
+++ b/galaxy_ng/app/management/commands/analytics/galaxy_collector.py
@@ -1,6 +1,9 @@
 import os
 import platform
 import distro
+
+from pathlib import PosixPath
+from typing import List
 from django.conf import settings
 
 from insights_analytics_collector import CsvFileSplitter, register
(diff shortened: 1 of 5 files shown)


galaxy_ng/app/management/commands/analytics/galaxy_collector.py
```diff
@@ -189,42 +183,80 @@ def container_tag_table(since, full_path, until, **kwargs):
...
+    Returns:
+        A list of strings that are filepaths where the data
+        was written to. It is usually just one filepath.
+    """
     file_path = _get_file_path(full_path, file_name)
     tfile = _get_csv_splitter(file_path, max_data_size)
 
+    # use a normal .execute to get the data as rows ...
     with Collector.db_connection().cursor() as cursor:
-        cursor.copy_expert(query, tfile)
```
Ladas, line 226: see this article on how to write copy_expert in psycopg3 https://www.psycopg.org/articles/2020/11/15/psycopg3-copy/

Then the rest of this PR shouldn't be needed. (we can't actually do it that way to avoid OOM, also COPY command is much more effective than basic select)

Ladas:
@jctanner just related to your comment, we'll be adding Minio for CI testing for Analytics. So we can test the S3 integration. So we can do it also here after. cc @dmzoneill

PR #1774, opened 2023-06-14
Integration test updates for CRC

awcrosby:
* Clean up repos and distros for repo tests running in crc
* Explicitly skip signing service tests in crc

No-Issue

### DIFF
--- a/galaxy_ng/tests/integration/api/test_collection_signing.py
+++ b/galaxy_ng/tests/integration/api/test_collection_signing.py
@@ -426,7 +426,7 @@ def test_copy_collection_without_signatures(api_client, config, settings, flags,
 
 @pytest.mark.collection_signing
 @pytest.mark.collection_move
-@pytest.mark.standalone_only  # This test can't run on crc yet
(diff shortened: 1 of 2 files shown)


jctanner:
How exactly are these tests going to run in standalone mode when they only operate when there's no signing service? Either they need to be decorated as insights mode only or disabled explicitly for CRC or at least give more context about why these are being changed.

chr-stian (approved):

awcrosby:
@jctanner for `test_collection_signing.py` tests, my attempt was to disable explicitly for CRC.

They already had a `pytest.skip` when `not settings.get("GALAXY_REQUIRE_SIGNATURE_FOR_APPROVAL")` and I added a `@pytest.mark.standalone_only` decorator to make that explicit.

jctanner (approved):
Long term, I'd like to copy the auto-cleanup pattern that pulp has in functional tests https://github.com/pulp/pulpcore/blob/main/pulpcore/tests/functional/__init__.py#L952-L975

PR #1776, opened 2023-06-15
Community profile for oci environment

jerabekjiri:
Issue: AAH-2382

Fixed by overwriting `PULP_GALAXY_CONTAINER_SIGNING_SERVICE='@none None'`
~~@newswangerd If I don't remove signing services and ENABLE_SIGNING in the base profile, then I always get a warning related to the signing variable being set, maybe we should update the base profile and the profile should decide if signing is enabled or not~~ 

(for the community profile we hide the execution environment, so setting the singing service for containers is wrong)


jerabekjiri:
@jctanner I haven't tried, I was unable to run integration tests with the oci_env and I don't know any other way to test it.

profiles/community/init.sh
```diff
@@ -0,0 +1,14 @@
+# #!/bin/bash
+
+set -o nounset
+set -o errexit
+set -o pipefail
+
+source /src/galaxy_ng/profiles/base/init.sh
```
newswangerd, line 7: Don't run the init script from the base profile. You can make galaxy_ng/base a required profile for this profile by doing this: https://github.com/ansible/galaxy_ng/blob/master/profiles/insights/profile_requirements.txt. That will cause oci-env to exit if the base profile isn't included in your compose profile.

profiles/community/init.sh
```diff
@@ -0,0 +1,14 @@
...
+
+cp -r /src/galaxy_ng/profiles/community/galaxy-importer /etc/galaxy-importer
+
+setup_collection_signing_service
+
+# collect rest_framewrok static files
+django-admin collectstatic --no-input --clear
```
newswangerd, line 14: This needs to be added to oci-env itself. It got removed at some point and I don't know why

profiles/community/pulp_config.env
```diff
@@ -0,0 +1,49 @@
+PULP_CONTENT_PATH_PREFIX=/api/v3/artifacts/collections/
+
+PULP_GALAXY_API_PATH_PREFIX=/api/
+PULP_GALAXY_AUTHENTICATION_CLASSES=['rest_framework.authentication.SessionAuthentication', 'rest_framework.authentication.TokenAuthentication', 'rest_framework.authentication.BasicAuthentication', 'django.contrib.auth.backends.ModelBackend']
+PULP_GALAXY_DEPLOYMENT_MODE=standalone
+PULP_GALAXY_REQUIRE_CONTENT_APPROVAL=false
+PULP_GALAXY_AUTO_SIGN_COLLECTIONS=false
+PULP_RH_ENTITLEMENT_REQUIRED=insights
+
+PULP_ANSIBLE_API_HOSTNAME=http://localhost:5001
```
newswangerd, line 10: Don't duplicate settings in galaxy_ng/base. These should be inherited from the base profile.

newswangerd (changes requested):

dev/common/setup_test_data.py
```diff
@@ -15,6 +17,8 @@
 
 TEST_NAMESPACES = {}
 
+auth_backend = os.environ.get('HUB_TEST_AUTHENTICATION_BACKEND')
```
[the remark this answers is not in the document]
jerabekjiri: thanks, should this change be made here as well? https://github.com/ansible/galaxy_ng/blob/f4c01ffab63d456d18461942d90bb4351dc46bf0/galaxy_ng/tests/integration/conftest.py#L109

profiles/community/init.sh
jerabekjiri: It's still being used here, but I didn't want to include the UI part, since the UI configuration doesn't work with community (`API_PATH_PREFIX` is different)
https://github.com/ansible/galaxy_ng/blob/5c301f69c1075700758f5e8db783c66f20d085bb/profiles/base/init.sh#L104-L108

jerabekjiri:
@newswangerd I noticed in https://github.com/ansible/galaxy_ng/blob/620cc506a8d78e99855e6583f8f384300026df6c/profiles/base/pulp_config.env there's a duplicate of `PULP_GALAXY_REQUIRE_CONTENT_APPROVAL`, first one is set to `False` and the second one overrides to `true`, what's the default for a base profile: `PULP_GALAXY_REQUIRE_CONTENT_APPROVAL=true`?

dev/common/setup_test_data.py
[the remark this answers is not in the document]
jerabekjiri: I'm unable to get it working with `settings.get('HUB_TEST_AUTHENTICATION_BACKEND')`

profiles/community/init.sh
jerabekjiri: I'm doing this so I can use the `setup_collection_signing_service` command, without this import, the command doesn't exist.

dev/common/setup_test_data.py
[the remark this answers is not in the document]
newswangerd: Ah, apologies. I was getting this variable confused with the GALAXY_AUTHENTICATION configuration. Loading this from the env var is fine.

profiles/community/init.sh
newswangerd: if the signing service is required, couldn't you just set the ENABLE_SIGNING env var in the pulp_config.env file?

newswangerd (approved):

profiles/community/github_mock/flaskapp.py
```diff
@@ -0,0 +1,232 @@
+#!/usr/bin/env python
```
jctanner, line 1: Why is this copied instead of symlinked?

jctanner (changes requested):

jctanner (approved):

PR #1779, opened 2023-06-20
Role download count fail gracefully if read-only database

EamonTracey:
Issue: [AAH-2415](https://issues.redhat.com/browse/AAH-2415)

### DIFF
--- a/CHANGES/2415.misc
+++ b/CHANGES/2415.misc
@@ -1,0 +1,1 @@
+Gracefully fail the legacy role download count increment if the database is in read-only mode.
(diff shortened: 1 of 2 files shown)


EamonTracey:
@jctanner I don't have permissions to request a review, so I'm pinging you. Also, the TDD validation fails because tests weren't modified, but I don't see how to write a test for when the database is read-only.

jctanner:
I don't think we have a good way to test read-only in CI, so we'll have to make a TDD exception on this PR.

jctanner (approved):

PR #1788, opened 2023-07-05
adapt test_ansible_lint_exception to different versions

chr-stian:
No-Issue

### DIFF
--- a/galaxy_ng/tests/integration/api/test_artifact_upload.py
+++ b/galaxy_ng/tests/integration/api/test_artifact_upload.py
@@ -6,9 +6,10 @@ from unittest.mock import patch
 
 import pytest
 from orionutils.generator import build_collection, randstr
+from pkg_resources import parse_version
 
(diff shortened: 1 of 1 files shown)


galaxy_ng/tests/integration/api/test_artifact_upload.py
```diff
@@ -6,9 +6,10 @@
 
 import pytest
 from orionutils.generator import build_collection, randstr
+from pkg_resources import parse_version
 
 from galaxy_ng.tests.integration.constants import USERNAME_PUBLISHER
-from ..conftest import is_hub_4_5
+from ..conftest import is_hub_4_5, get_hub_version
```
jctanner, line 12: I know this line was already here, but I think we probably shouldn't import from conftest. Stuff in there should be fixtures that are "magically" imported via pytest. https://stackoverflow.com/questions/68738078/is-it-discouraged-to-import-conftest-py-within-test-modules

awcrosby (approved):

PR #1794, opened 2023-07-12
Vendor django-automated-logging

bmclaughlin:
#### What is this PR doing:
The `django-automated-logging` module we use for api access logging hasn't been touched in ~2 years.  It is not Django 4.x compatible.  A decision was made to vendor the code to maintain current functionality and not block our Django 4 upgrade.

I'm in the process of determining the best way to verify the logging in tests, currently planned for a follow-up PR. 

Issue: AAH-2388

#### Reviewers must know:

When reviewing/testing in oci-env, `PULP_GALAXY_ENABLE_API_ACCESS_LOG=true` must be set in the environment and the resulting log can be inspected with the following commands:
```
oci-env shell
tail -f /var/log/galaxy_api_access.log
```
Navigating around the app should log CRUD actions as well as username, ip address, and other details.  Full details are available in the original jira epic https://issues.redhat.com/browse/AAH-643.


rochacbruno:
Instead of dealing with setup.py and packaging I would simply copy 	`automated_logging` to 	 `galaxy_ng/` and then on INSTALLED_APPS add `galaxy_ng.automated_logging` - if there are internal absolute imports looking for `from automated_logging` etc.. then we can force it to sys.modules on galaxy_ng `__init__.py`  

```
import sys
import automated_logging
sys.modules.setdefault('automated_logging', automated_logging)
```

cutwater:
Minor: It would be nice if the vendored package is placed in a subpackage (e.g. `galaxy_ng._vendor.automated_logging`)

bmclaughlin:
@cutwater Moved to a package as suggested.

cutwater (approved):

jctanner (approved):

jctanner:
Tests will be added in a future PR.

PR #1795, opened 2023-07-12
add new CI workflow to test against the automation hub collection

EamonTracey:
Issue: AAH-2419

### DIFF
--- a/.github/workflows/ci_automation_hub_collection.yml
+++ b/.github/workflows/ci_automation_hub_collection.yml
@@ -1,0 +1,62 @@
+---
+name: Automation Hub Collection CI
+on: 
+  pull_request:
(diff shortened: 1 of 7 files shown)


jctanner:
The ".ci" directory is mostly controlled via https://github.com/pulp/plugin_template. I think we should avoid putting anything else in there. The "dev" directory is probably a better place.

EamonTracey:
Includes uses of the following modules: ah_approval, ah_collection, ah_ee_registry, ah_ee_registry_index, ah_ee_repository, ah_group, ah_namespace, ah_repository, ah_repository_sync, ah_role, ah_token, ah_user

rochacbruno (approved):

PR #1807, opened 2023-07-20
Fix namespace logos on s3

newswangerd:
Add minio CI tests

Issue: AAH-2575


jctanner (approved):

jctanner:
The new proxy scripts needs to be copied to profiles/insights/proxy and then tests should start passing.

PR #1810, opened 2023-07-21
switch to py3.11 in all oci builds

jctanner:
Some helpers have been added for oci. The action lib is too much abstraction and python subprocess calling. We need something more direct for troubleshooting and debugging locally.


jctanner:
I have reproduced the test_namespace_sync failure locally. Attempting to debug.

newswangerd (approved):
I'm not sure I like the oci-env scripts, but I can't find any substantial faults with it and it's not breaking the gh-action targets, so I guess this is okay.

PR #1811, opened 2023-07-24
Beta galaxy

chr-stian:
test cases to be run against
thttps://beta-galaxy-stage.ansible.com/

this change needs this galaxykit PR 
https://github.com/ansible/galaxykit/pull/96

Jira ticket
https://issues.redhat.com/browse/AAH-2178


galaxy_ng/tests/integration/utils/iqe_utils.py
```diff
@@ -218,6 +238,12 @@ def is_ephemeral_env():
     )
 
 
+def is_beta_galaxy():
```
jctanner, line 241: can we rename this to something like is_beta_galaxy_stage ?

jctanner (approved):

integration_requirements.txt
```diff
@@ -8,4 +8,4 @@ openapi-spec-validator
 jsonschema
 hvac
 importlib_resources
-galaxykit @ git+https://github.com/ansible/galaxykit
+galaxykit @ git+https://github.com/ansible/galaxykit@beta_galaxy
```
jctanner, line 11: this branch is merged, so we can move back to the branchless url

dev/common/RUN_INTEGRATION_BETA_GALAXY.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
```
jctanner, line 1: I'd rather this go into a path like dev/beta/...

galaxy_ng/tests/integration/constants.py
```diff
@@ -251,3 +251,36 @@
         "token": "abcdefghijklmnopqrstuvwxyz1234567895",
     },
 }
+
+BETA_GALAXY_PROFILES = {
```
jctanner, line 255: Another place where the context is stage only, so i'd prefer a var like BETA_GALAXY_STAGE_PROFILES

PR #1817, opened 2023-07-26
Remove lru caching on get_access_policy for memory performance concerns.

jctanner:
https://github.com/pulp/pulpcore/pull/4090

We subclass and override the get_access_policy function, so pulp's upstream change probably won't directly help us.


jctanner:
Not requiring tests for this one.

rochacbruno (approved):

PR #1820, opened 2023-07-27
Bump pulpcore to 3.28.7 for httpdownloader+async fix.

jctanner:

PR #1822, opened 2023-08-01
Undo skip marker now that httpdownloader+async is fixed in pulpcore.

jctanner:
https://github.com/ansible/galaxy_ng/pull/1820#issuecomment-1659046788


EamonTracey (approved):
Cool 👍

PR #1830, opened 2023-08-02
fix jinja templating call from ah collection workflow

EamonTracey:
~~The automation hub collection now has a true `galaxy.yml` file in the repository, so the workflow no longer must create the `galaxy.yml`~~


EamonTracey:
Looks like the collection was also renamed...

EamonTracey:
This fixes the workflow given the changes in https://github.com/ansible/automation_hub_collection/commit/0cd14ac0fde5bbd9a4853565b53e99d19d76ef0d . Continue to use the `galaxy.yml.j2` template since the name of the collection is not yet set in stone.

jctanner (approved):

rochacbruno (approved):

PR #1837, opened 2023-08-09
Add Support for Dynamic Settings

rochacbruno:
Issue: AAH-2009

Feature is deactivated by default, 
to activate Requires Dynaconf 3.2.2 

## What this PR is doing?

1. This PR adds a new table to the models `config.Setting`, this table will store `key:value` pairs to be dynamically added to Dynaconf data and be available through `django.conf.settings`
    **IMPORTANT** key/value are stores in the database as `str`, upon access value can be read raw or parsed by dynaconf/toml, model also stores a timestamp

2. Along with the model, the PR also adds a schema on `dynamic_settings.py:DYNACMIC_SETTINGS_SCHEMA` which defines which keys can be stored on the database, as we cannot allow all settings to be dynamic.

3. A management command `django-admin galaxy-settings` is also added with commands to `set,get,delete,list,inspect` the settings on the database.

4. Every time a Setting is mutated on database it will also update a cache on Redis, if Redis is not available then caching will be skipped (and logged)

5. Finally, the PR adds a Dynaconf method hook to the `dynaconf:Settings:get` method, making access to any variables listed on the schema to trigger an immediate lookup on `cache -> db` **IMPORTANT** this last feature requires Dynaconf 3.2.2 (to be released) 

**NOTE** The PR can be merged before Dynaconf 3.2.2 is released, there will be no breaking, only the feature n 5 will not be enabled.

6. This PR adds a new variable to settings.py `GALAXY_DYNAMIC_SETTINGS=False` this is the main toggle for the feature n 5, this is by default set to False, which means, `django.conf.settings` will lookup for keys in the cache/db only if this is set to True.

## How to add settings to the database?

```console
django-admin galaxy-settings set --key=GALAXY_FEATURE_FLAGS__power_mode --value=true
django-admin galaxy-settings get --key GALAXY_FEATURE_FLAGS 
```

More details on `django-admin galaxy-settings --help`

Or via script/shell

```py
from galaxy_ng.app.models.config import Setting
Setting.set_value_in_db("GALAXY_FEATURE_FLAGS", "@merge power_mode=true")
#or equivalent 
Setting.set_value_in_db("GALAXY_FEATURE_FLAGS__power_mode", "true")
```

**Currently there is no API (WIP)** so command and script is the only way to manage the settings.

Due to the validation, it is not recommended to write directly to the database.

## How to read settings from database with current version of dynaconf?

```python
from galaxy_ng.app.models.config import Setting
result = Setting.get("GALAXY_FEATURE_FLAGS")
# Will lookup DB and then merge with current django/dynaconf settings
# before returning the result
```

## How to read settings from db/cache after the upgrade to Dynaconf 3.2.2

1st enable `GALAXY_DYNAMIC_SETTINGS=True` via settings.py or PULP_ prefixed envvar.

then access settings normally using django.

```python
result = settings.GALAXY_FEATURE_FLAGS  
# upon access to key listed on DYNAMIC_SETTINGS_SCHEMA
# dynaconf will magically lookup on cache/db to merge data
# before returning the result.
```

## What about the API/UI?

- **API** will  be added in another PR
- **UI** will be created based on DYNAMIC_SETTINGS_SCHEMA and will be added after the API


jerabekjiri (approved):

slemrmartin (approved):
:+1: Awesome work

PR #1851, opened 2023-08-28
[oci-env] Add pulp_certs volume

jerabekjiri:
Issue: AAH-2648


himdel:
Works for me :+1: 
I can access remote execution environments and remote registries after a rebuild&restart without 500s.
(Assuming triggering this doesn't depend on something extra.)

PR #1854, opened 2023-08-29
test for test_x_repo_search_acl_social_user instability in CI

jerabekjiri:
No-Issue [AAH-2638](https://issues.redhat.com/browse/AAH-2638)

https://gist.github.com/jerabekjiri/f05ed52c1f3c1fdb68522f063f98fa1f

The `search_url` response includes the incremented collection version (`autohubtest2.immijnge-1.0.1`) in both the `staging` and `published` repositories at the same time. 
I'm not sure why this is happening, since in https://github.com/ansible/galaxy_ng/blob/01e89102e973c57581d3b6bafca6eb6be1b74080/galaxy_ng/tests/integration/conftest.py#L249C32-L249C32, after the collection is built, it waits for the collection to be in the published repository. Nevertheless, this solution fixes the issue by searching only in the published repository.

### DIFF
--- a/CHANGES/2638.misc
+++ b/CHANGES/2638.misc
@@ -1,0 +1,1 @@
+Improve ``test_x_repo_search_acl_social_user`` and ``test_x_repo_search_acl_anonymous_user`` integration test stability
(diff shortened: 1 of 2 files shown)


jctanner (approved):
Nice find, ty!

PR #1875, opened 2023-09-10
Insights proxy broken pipe retry

jctanner:
The insights profile tests fail randomly because of broken pipe errors when connecting to galaxy_ng.

### DIFF
--- a/profiles/insights/proxy/main.go
+++ b/profiles/insights/proxy/main.go
@@ -13,6 +13,8 @@ import (
 	"regexp"
 	"strings"
 	"time"
+    "net"
+    "syscall"
 )
 
 var refreshTokens = map[string]string{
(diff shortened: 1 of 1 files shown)


profiles/insights/proxy/main.go
```diff
@@ -1,193 +1,195 @@
...
-	"io/ioutil"
-	"log"
-	"math/rand"
-	"net/http"
-	"net/url"
-	"os"
-	"regexp"
-	"strings"
-	"time"
+    "encoding/base64"
```
rochacbruno, line 4: have you used ` go fmt` or it is your text editor fixing indentation?

rochacbruno: AFAIK in Golang the standard is to use tabs instead of spaces, and that is what `go fmt` sets by default.

rochacbruno (approved):

PR #1898, opened 2023-09-22
fix and improve sync tests

chr-stian:
More galaxykit usage
No-Issue


jerabekjiri (approved):
LGTM :+1:, tests are passing now

PR #1901, opened 2023-09-27
optimize user v3 namespace owner lookups.

jctanner:
the old code is probably causing timeouts during social logins because of the number of namespaces it has to iterate through.


gerrod3 (approved):
We should definitely look into adding a new utility function that accepts role strings so we don't have to do the conversion role->permissions, but for now this looks good.

drodowic (approved):

PR #1908, opened 2023-10-02
`/api/v1/users?username=test` throws server error (500)

jerabekjiri:
Issue: AAH-2731


jctanner (approved):

jctanner:
tests are going to fail until we fix all the v1 related code to use old-galaxy instead of galaxy.ansible.com

jctanner:
@jerabekjiri can you rebase please?

PR #1922, opened 2023-10-06
Add `download_count` ordering to legacy roles endpoint

jerabekjiri:
Issue: AAH-2760


jctanner (approved):
Looks good but can you add a test please?

jerabekjiri:
@jctanner
I added a test, and also added sorting by the `modified` field as @himdel suggested.

PR #1923, opened 2023-10-06
github_user should override namespace.name for legacy roles.

jctanner:
there are 5408 roles on old galaxy where the github_user doesn't match the role's namespace name

So when you install laniakea.galaxy, the role found via `?owner_username=laniakea&name=galaxy` is http://old-galaxy.ansible.com/api/v1/roles/51041/ ... which has a github_user key with value Laniakea-elixir-it

The way that worked was the role model linked to a repository model and the repository model linked to a provider namespace model ... the provider namespace's "name" property is what the role serializer emits for github_user

I'm not sure what this means for future imports of the roles considering the authors need to provide the proper org|user name to the CLI ... so I'll have to do some testing on that.

After this fix is deployed, we'll have to inject the missing field into all of the relevant roles' full_metadata column.

### DIFF
--- a/galaxy_ng/app/api/v1/tasks.py
+++ b/galaxy_ng/app/api/v1/tasks.py
@@ -149,6 +149,7 @@ def legacy_role_import(
             'clone_url': clone_url,
             'tags': galaxy_info["galaxy_tags"],
             'commit': github_commit,
+            'github_user': github_user,
             'github_repo': github_repo,
             'github_reference': github_reference,
             'issue_tracker_url': galaxy_info["issue_tracker_url"] or clone_url + "/issues",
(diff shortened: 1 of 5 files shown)


rochacbruno (approved):

galaxy_ng/app/api/v1/tasks.py
```diff
@@ -261,6 +262,7 @@ def legacy_sync_from_upstream(
         rkey = (ruser, rname)
         remote_id = rdata['id']
         role_versions = rversions[:]
+        github_user = rdata['github_user']
         github_repo = rdata['github_repo']
```
bittner, line 266: `rdata['github_user']` is already read above into variable `ruser`. Then again, `ruser` is used instead of this variable to construct `clone_url`. Maybe this could be consolidated?

ifalatik:
Is this also why my role https://galaxy.ansible.com/ui/standalone/roles/ifalatik/docker_project_deployment/ (GitHub -> https://github.com/ifalatik/docker-project-deployment) can't be updated from the galaxy UI? I can't see the versions / documentation either.

PR #1931, opened 2023-10-11
Add tags/roles and tags/collections endpoints

jerabekjiri:
Issue: [AAH-2761](https://issues.redhat.com/browse/AAH-2761)

- add sorting by `name` and `count` for both endpoints
- filter by `name, name__icontains, name__contains, name__startswith` for both endpoints

### `_ui/v1/tags/collections`:
   - add new endpoint, so it doesn't break existing one (`_ui/v1/tags/`)
   - add `count` attribute to `TagSerializer` (https://github.com/pulp/pulp_ansible/issues/1612)

### `_ui/v1/tags/roles`:
- **To populate tags endpoint: `django-admin populate-role-tags`**
```
GET /api/_ui/v1/tags/roles/
{
    "meta": {
        "count": 10
    },
    "links": {
        "first": "/api/_ui/v1/tags/roles/?limit=10&offset=0",
        "previous": null,
        "next": null,
        "last": "/api/_ui/v1/tags/roles/?limit=10&offset=0"
    },
    "data": [
        {
            "name": "firefox",
            "count": 5
        },
        {
            "name": "development",
            "count": 3
        },
        {
            "name": "workstation",
            "count": 1
        },
        {
            "name": "bfg",
            "count": 1
        }
    ]
}
```


galaxy_ng/app/api/ui/viewsets/tags.py
```diff
@@ -21,3 +30,106 @@ def list(self, request, *args, **kwargs):
...
+        if sort is not None and sort in ["name", "-name", "count", "-count"]:
+            reverse = True if "-" in sort else False
+            sort_field = sort.replace("-", "")
+            queryset = sorted(queryset, key=lambda x: x[sort_field], reverse=reverse)
+        elif sort is not None:
+            raise serializers.ValidationError(_("Invalid Sort: '{}'").format(sort))
+
+        return queryset
+
+    def list(self, request, *args, **kwargs):
```
jctanner, line 123: Is this performant with 40K+ roles?

jerabekjiri: so I did some performance testing on my localhost with locust, synced 5k roles:
Screenshot from 2023-10-24 13-58-09


with 100 users concurently it's different story though Screenshot from 2023-10-24 13-47-53

I'm wondering if caching would be possible and helpful, or to do this we would have to create a role tags model (and probably a model for `full_metadata` as well)?

galaxy_ng/app/api/ui/viewsets/tags.py
```diff
@@ -21,3 +30,106 @@ def list(self, request, *args, **kwargs):
...
         return self.get_paginated_response(serializer.data)
+
+
+class TagFilterOrdering(filters.OrderingFilter):
+    def filter(self, qs, value):
+        if value is not None and any(v in ["count", "-count"] for v in value):
+            order = "-" if "-count" in value else ""
+
+            return qs.filter(
+                ansible_collectionversion__ansible_crossrepositorycollectionversionindex__is_highest=True  # noqa: E501
```
rochacbruno, line 41: does it really need to use the `crossrepo` index? for community we have only one repo and it would be faster to just take the is_highest from the single repo.

jerabekjiri: That's true, but then this wouldn't work for PAH and CRC

jctanner (approved):
Let's try it out and see how it goes.

galaxy_ng/app/migrations/0042_legacyroletag_legacyrole_tags.py
```diff
@@ -0,0 +1,31 @@
+# Generated by Django 4.2.6 on 2023-10-25 21:26
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("galaxy", "0041_alter_containerregistryremote_remote_ptr"),
```
himdel, line 8: `0042_namespace_created_namespace_updated` ?

(there's a new migration if you rebase .. and this bit seems to produce an infinite loop when starting)

himdel:
The collections endpoint seems to work, but the items are missing `count`.
The roles endpoint returns 0 items even after I imported some roles with tags.

Do I need to trigger some manage command to regenerate or something?

jerabekjiri:
@himdel  
> The collections endpoint seems to work, but the items are missing count.

`count` will be included in next pulp_ansible release https://github.com/pulp/pulp_ansible/issues/1612

> The roles endpoint returns 0 items even after I imported some roles with tags.
> 
> Do I need to trigger some manage command to regenerate, have a flag enabled, or something?

Good point, I forgot to mention that you need to run `django-admin populate-role-tags` to populate tags. Sorry about that

galaxy_ng/app/migrations/0042_legacyroletag_legacyrole_tags.py
jerabekjiri: Good catch :+1:, rebased.

himdel (approved):
Thanks! :) Works for me, I can see role tag counts now :+1:

PR #1934, opened 2023-10-13
Enable social auth users to see other users.

jctanner:
Users can't see the list of users when trying to add owners to their namespaces.


jctanner:
`I'd probably create a custom access policy ([here's one for deleting collections](https://github.com/ansible/galaxy_ng/pull/1900/files#diff-2a9ccd7546893821c4c3488d210b249f56e466222f93e8c1d1d7dc8a32e1caaeR262)) to maintain the current access and allow for the new case.`

jctanner:
Failing test_copy_cv_endpoint works locally with this patch, so ignoring the failure in CI.

PR #1937, opened 2023-10-16
fix social auth namespace creation (again)

jctanner:
The loop that checked for existing users by email was always skipping the checks because it was evaluating the "email" property from github which is usually null.

https://issues.redhat.com/browse/AAH-2787


dev/oci_env_integration/oci_env_configs/community.compose.env
```diff
@@ -12,8 +12,8 @@ DJANGO_SUPERUSER_USERNAME=admin
 DJANGO_SUPERUSER_PASSWORD=admin
 
 ## Enable github social auth by setting these two values
-SOCIAL_AUTH_GITHUB_KEY='abcd1234'
-SOCIAL_AUTH_GITHUB_SECRET='abcd1234'
+#SOCIAL_AUTH_GITHUB_KEY='abcd1234'
+#SOCIAL_AUTH_GITHUB_SECRET='abcd1234'
```
bmclaughlin, line 16: Should these lines be removed to reduce confusion?

bmclaughlin (approved):

dev/oci_env_integration/oci_env_configs/community.compose.env
jctanner: tbh, i'm not sure why we have these env files here and also profile specific files in the profiles dir.

PR #1939, opened 2023-10-17
Metrics Collection: Added GALAXY_API_PATH_PREFIX to API call

slemrmartin:
#### What is this PR doing:

Fixes current API call, which uses only base url and `/pulp/api/v3/status` and doesn't work on galaxy-dev. 

Issue: [AA-1892](https://issues.redhat.com/browse/AA-1892)

#### Reviewers must know:

### DIFF
--- a/galaxy_ng/tests/unit/app/management/commands/analytics/automation_analytics/test_data.py
+++ b/galaxy_ng/tests/unit/app/management/commands/analytics/automation_analytics/test_data.py
@@ -1,9 +1,6 @@
 import galaxy_ng.app.metrics_collection.common_data
 from django.test import TestCase, override_settings
 from unittest.mock import MagicMock, patch
-from django.conf import settings
-from urllib.parse import urljoin
-import os
 
 
 class TestAutomationAnalyticsData(TestCase):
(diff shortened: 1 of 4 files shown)


rochacbruno (approved):

galaxy_ng/tests/unit/app/management/commands/analytics/automation_analytics/test_data.py
```diff
@@ -0,0 +1,26 @@
...
+        mock_response = MagicMock(name="mock_response")
+        mock_response.status_code = 200
+        mock_request.return_value = mock_response
+
+        json_response = MagicMock(name="json")
+        mocked_api_status = MagicMock(name="api_status")
+        json_response.return_value = mocked_api_status
+        mock_response.json = json_response
+
+        response = galaxy_ng.app.metrics_collection.common_data.api_status()
```
rochacbruno, line 20: This function `api_status` is never called, so the mock on request above has no effect, you can verify this problem by placing a `1 / 0` on the first line of the api_status function and seeing that it never raises DivisionError.

And that happens because of this https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/tests/unit/app/management/commands/analytics/automation_analytics/test_collector.py#L61 

The other test is replacing/shadowing the method with a magic mock globaly, that affects the python global namespace, must be doing that via mock instead of assignment.

rochacbruno: If I comment out the line on the other test https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/tests/unit/app/management/commands/analytics/automation_analytics/test_collector.py#L61

I am able to run the test successfuly

```
❯ make docker/test/unit TEST=.app.management.commands.analytics
api django-admin test galaxy_ng.tests.unit.app.management.commands.analytics
Running on existing container 5ff2a5a973be
INFO: Using compose profile standalone
INFO: pulp_ansible:galaxy_ng packages installed from source
INFO: Image suffix is unset
INFO: Volume suffix is unset
Found 10 test(s).
Creating test database for alias 'default'...
System check identified no issues (0 silenced).
....pulp [None]: insights-analytics-collector.collector:ERROR: 'config' collector data is missing
.....@@@@@@@@@@@@@@@@@@@@
https://example.com
/api-test/xxx
.
----------------------------------------------------------------------
Ran 10 tests in 0.297s

OK
Destroying test database for alias 'default'...
```

slemrmartin: wow, very good catch, thanks!

slemrmartin:
@rochacbruno the current failing tests is not related to this PR, right?

PR #1940, opened 2023-10-17
`username_autocomplete` filter in legacy roles endpoint 

jerabekjiri:
Issue: [AAH-2782](https://issues.redhat.com/browse/AAH-2782)

### DIFF
--- a/CHANGES/2782.feature
+++ b/CHANGES/2782.feature
@@ -1,0 +1,1 @@
+Added ``username_autocomplete`` filter to ``LegacyRole``.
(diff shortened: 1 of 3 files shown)


jctanner (changes requested):
I think the filter was meant to only reduce the namespaces per https://github.com/ansible/galaxy/blob/b63a816a1fc307987c8f3a314b1b6407a8b1ece1/galaxy/api/views/search.py#L151

jctanner (approved):

PR #1943, opened 2023-10-20
Add edit support for some legacyrole fields.

jctanner:
A lot of the old legacy roles had their github_user, github_repo, repository.name, namespace.name, name changed over the years. We need a way to alter those fields from the API as people change the source data.

This PR does some hacky things because the data is still all in the json fullmetadata column. In a future PR I want to migrate all the important keys out to 1st class model fields and apply proper constraints.

### DIFF
--- a/galaxy_ng/app/api/v1/serializers.py
+++ b/galaxy_ng/app/api/v1/serializers.py
@@ -313,6 +313,15 @@ class LegacyRoleSerializer(serializers.ModelSerializer):
                 'pulp_href': pulp_href
             }
 
+        # FIXME - repository is a bit hacky atm
+        repository = {}
+        if obj.full_metadata.get('repository'):
+            repository = obj.full_metadata.get('repository')
+        if not repository.get('name'):
+            repository['name'] = obj.full_metadata.get('github_repo')
+        if not repository.get('original_name'):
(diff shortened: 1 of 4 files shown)


galaxy_ng/app/api/v1/serializers.py
```diff
@@ -336,6 +342,40 @@ def get_download_count(self, obj):
         return 0
 
 
+class LegacyRoleRepositoryUpdateSerializer(serializers.Serializer):
+    name = serializers.CharField(required=False, allow_blank=False, max_length=50)
+    original_name = serializers.CharField(required=False, allow_blank=False, max_length=50)
```
jerabekjiri, line 347: Should I be able to edit `original_name`?

jctanner: That's a good question. I think longterm, no you shouldn't be able to. For the short term though, we need to be able to set that based on the old api data.

jerabekjiri:
Updating some values like `username` or `id` in browsable API throws
Screenshot from 2023-10-31 15-22-26

jerabekjiri (approved):
LGTM

but question: we don't allow updating `tags` or `description`, AFAIK that happens when a user creates the role and to update, he has to create a new version with updated desc and tags?

PR #1944, opened 2023-10-22
Turn on dnf fastest mirrors to prevent getting stuck on slow default mirrors

jctanner:


profiles/base/Dockerfile
```diff
@@ -1,4 +1,5 @@
 FROM localhost/oci_env/pulp:base
+RUN echo "fastestmirror=1" >> /etc/dnf/dnf.conf
```
cutwater, line 2: To prevent additional unnecessary layer creation it would be better to combine this operation with the one that uses `dnf`.

cutwater (changes requested):

profiles/base/Dockerfile
jctanner: We don't have any direct dnf calls in this Dockerfile.

cutwater (approved):

PR #1946, opened 2023-10-23
Revamp legacyrole versions.

jctanner:
https://forum.ansible.com/t/role-import-from-github-results-in-master-release-not-the-tag-release-breaking-installs/1856

After digging around and mapping out the code from github.com/ansible/galaxy ...

- At import time, all tags for the repo are enumerated and added as versions (if semantic version compliant).
- tags should be valid semantic versions or a semantic version starting with 'v'
- the "name" field for a role version is the original tag name.
- the "version" field for a role version is the tag name stripped of any 'v' prefix.
- if the import command includes `--branch=<older-tag>` the old tag will set the role's commit, but ALL tags get added as versions anyway.
- if a tag is removed from the repo, the related role version is also removed.
- in the list view and detail view, versions are limit to the 10 most recent in descending order and only a few fields
- in the versions list page, more fields are given and the order is ascending
- role repos without tags do not end up with a list of versions
- the role's commit, commit date, etc are updated from HEAD of the checkout -after- the branch or tag is switched
- in old-galaxy if github_branch is given at import time with a tag name, the role will end up with HEAD as the github_branch ... which we don't want to do new galaxy. We should either use the tag value or the default branch.
- the cli's `--role-name` option passes the value as `alternate_role_name` in the POST request, but this will not change the actual name in the current old-galaxy code. It instead transforms the repository name's hyphens to underscores and nothing more. In the API `alternate_role_name` is stored in the `repository_alt_name` column.
- the galaxy_info.role_name field in meta/main.yml can change the role's name in the database.

### DIFF
--- a/galaxy_ng/app/api/v1/tasks.py
+++ b/galaxy_ng/app/api/v1/tasks.py
@@ -169,6 +169,65 @@ def do_git_checkout(clone_url, checkout_path, github_reference):
     return gitrepo, github_reference, last_commit
 
 
+def compute_all_versions(this_role, gitrepo):
+    """
+    Build a reconciled list of old versions and new versions.
+    """
+
+    # Combine old versions with new ...
+    old_metadata = copy.deepcopy(this_role.full_metadata)
+    versions = old_metadata.get('versions', [])
+
+    # Clean out non-semver versions due to bad import code
+    for version in versions:
+        try:
+            parse_version_tag(version['version'])
+        except ValueError:
+            versions.remove(version)
+
+    # ALL semver tags should become versions
+    current_tags = []
+    for cversion in versions:
+        # we want tag but the sync'ed roles don't have it
+        # because the serializer returns "name" instead.
+        tag = cversion.get('tag')
+        if not tag:
+            tag = cversion.get('name')
+        current_tags.append(tag)
+    for tag in gitrepo.tags:
+
+        # must be a semver compliant value ...
+        try:
+            version = parse_version_tag(tag.name)
+        except ValueError:
+            continue
+
+        if str(version) in current_tags:
+            continue
+
+        ts = datetime.datetime.now().isoformat()
+        vdata = {
+            'id': str(uuid.uuid4()),
+            'tag': tag.name,
+            'version': str(version),
+            'commit_date': tag.commit.committed_datetime.isoformat(),
+            'commit_sha': tag.commit.hexsha,
+            'created': ts,
+            'modified': ts,
+        }
+        logger.info(f'adding new version from tag: {vdata}')
+        versions.append(vdata)
+
+    # remove old tag versions if they no longer exist in the repo
+    git_tags = [x.name for x in gitrepo.tags]
+    for version in versions[:]:
+        if version.get('version') not in git_tags:
+            logger.info(f"removing {version['version']} because it no longer has a tag")
+            versions.remove(version)
+
+    return versions
+
+
 def legacy_role_import(
     request_username=None,
     github_user=None,
@@ -322,52 +381,8 @@ def legacy_role_import(
                 name=role_name
             )
 
-        # Combine old versions with new ...
-        old_metadata = copy.deepcopy(this_role.full_metadata)
-        versions = old_metadata.get('versions', [])
-
-        # Clean out non-semver versions due to bad import code
-        for version in versions:
-            try:
-                parse_version_tag(version['version'])
-            except ValueError:
-                versions.remove(version)
-
-        # ALL semver tags should become versions
-        current_tags = []
-        for cversion in versions:
-            # we want tag but the sync'ed roles don't have it
-            # because the serializer returns "name" instead.
-            tag = cversion.get('tag')
-            if not tag:
-                tag = cversion.get('name')
-            current_tags.append(tag)
-        for tag in gitrepo.tags:
-
-            # must be a semver compliant value ...
-            try:
-                version = parse_version_tag(tag.name)
-            except ValueError:
-                continue
-
-            if str(version) in current_tags:
-                continue
-
-            ts = datetime.datetime.now().isoformat()
-            vdata = {
-                'id': str(uuid.uuid4()),
-                'tag': tag.name,
-                'version': str(version),
-                'commit_date': tag.commit.committed_datetime.isoformat(),
-                'commit_sha': tag.commit.hexsha,
-                'created': ts,
-                'modified': ts,
-            }
-            logger.info(f'adding new version from tag: {vdata}')
-            versions.append(vdata)
-
         # set the enumerated versions ...
-        new_full_metadata['versions'] = versions
+        new_full_metadata['versions'] = compute_all_versions(this_role, gitrepo)
 
         # Save the new metadata
         this_role.full_metadata = new_full_metadata
--- a/galaxy_ng/app/api/v1/tasks.py
+++ b/galaxy_ng/app/api/v1/tasks.py
@@ -182,6 +182,9 @@ def sort_versions(versions):
 
 
 def normalize_versions(versions):
+
+    original = [copy.deepcopy(x) for x in versions]
+
     # convert old integer based IDs to uuid
     for vix, version in enumerate(versions):
         if isinstance(version.get('id', ''), int):
@@ -201,6 +204,9 @@ def normalize_versions(versions):
     # where previous galaxy_ng import code mistakenly thought
     # the branch+commit should be a version instead of only tags.
     for version in versions[:]:
+        if not version.get('tag'):
+            versions.remove(version)
+            continue
         lver = LooseVersion(version['tag'].lower())
         if not all(isinstance(x, int) for x in lver.version):
             versions.remove(version)
--- a/galaxy_ng/app/api/v1/tasks.py
+++ b/galaxy_ng/app/api/v1/tasks.py
@@ -10,6 +10,8 @@ import semantic_version
 
 from django.db import transaction
 
+from ansible.module_utils.compat.version import LooseVersion
+
 from galaxy_importer.config import Config
 from galaxy_importer.legacy_role import import_legacy_role
 
@@ -171,6 +173,41 @@ def do_git_checkout(clone_url, checkout_path, github_reference):
     return gitrepo, github_reference, last_commit
 
 
+def sort_versions(versions):
+    """
+    Use ansible-core's LooseVersion util to sort the versions.
+    """
+    sorted_versions = sorted(versions, key=lambda x: LooseVersion(x['tag'].lower()))
+    return sorted_versions
+
+
+def normalize_versions(versions):
+    # convert old integer based IDs to uuid
+    for vix, version in enumerate(versions):
+        if isinstance(version.get('id', ''), int):
+            versions[vix]['upstream_id'] = version['id']
+            versions[vix]['id'] = str(uuid.uuid4())
+
+    # Normalize keys
+    for vix, version in enumerate(versions):
+        if not version.get('tag'):
+            if version.get('name'):
+                versions[vix]['tag'] = version['name']
+            else:
+                versions[vix]['tag'] = version['version']
+
+    # if looseversion can't make a numeric version from this tag
+    # it's not going to work later. This also should cover the case
+    # where previous galaxy_ng import code mistakenly thought
+    # the branch+commit should be a version instead of only tags.
+    for version in versions[:]:
+        lver = LooseVersion(version['tag'].lower())
+        if not all(isinstance(x, int) for x in lver.version):
+            versions.remove(version)
+
+    return versions
+
+
 def compute_all_versions(this_role, gitrepo):
     """
     Build a reconciled list of old versions and new versions.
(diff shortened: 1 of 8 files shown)


jctanner:
@geerlingguy any input you have on this would be great

geerlingguy:
For my own roles, I always use a semantic version (e.g. `1.2.3`), but I believe a few people in the past would use more 'arbitrary' versioning (not semantic, and not even `v1.2.3`), so for those edge cases, this might not cover it? (Not entirely sure... I don't know anyone that does that, but I know in old-old galaxy it was allowed for practically anything to be a version string. Quite confusing!

A couple caveats (edge cases to consider):

  - Please make sure versions like 1.10.0 would be listed after 1.9.0, and not end up sorting after 1.1.0 but before 1.2.0. I know that bug caught the old galaxy up one time.
  - Regarding the role naming, I have a few old roles that were grandfathered in, like https://galaxy.ansible.com/ui/standalone/roles/geerlingguy/repo-epel/ (`geerlingguy.repo-epel`) that have a `-` in the name, instead of `_`. Would those still work after a new import? See [`galaxy_info.role_name` in the role's meta.yml](https://github.com/geerlingguy/ansible-role-repo-epel/blob/master/meta/main.yml#L5).

galaxy_ng/app/api/v1/tasks.py
```diff
@@ -76,12 +90,154 @@ def find_real_role(github_user, github_repo):
...
+        - the git.Repo object
+        - the enumerated github_reference (aka branch or tag)
+        - the last commit object from the github_reference or $HEAD
+    :rtype: tuple(git.Repo, string, git.Commit)
+    """
+    logger.info(f'CLONING {clone_url}')
+
+    # pygit didn't have an obvious way to prevent interactive clones ...
+    pid = subprocess.run(
+        f'GIT_TERMINAL_PROMPT=0 git clone --recurse-submodules {clone_url} {checkout_path}',
```
gotmax23, line 115: It looks like this code was copied from elsewhere in the file, but this type of string interpolation with `shell=True` is really not a good idea...

galaxy_ng/app/api/v1/tasks.py
```diff
@@ -24,6 +29,15 @@
 logger = logging.getLogger(__name__)
 
 
+def parse_version_tag(value):
+    value = str(value)
+    if not value:
+        raise ValueError('Empty version value')
+    if value[0].lower() == 'v':
+        value = value[1:]
+    return semantic_version.Version(value)
```
gotmax23, line 38:
```suggestion
    return semantic_version.Version(value.lstrip('v'))
```

sivel (approved):

galaxy_ng/app/api/v1/tasks.py
jctanner: @gotmax23 i think this suggestion neglects the case where "v" is an uppercase "V".

jctanner:
@geerlingguy re versions that aren't semantic ...

I pulled the data from old-galaxy's database and ran some checks. There are 5208 unique version strings in the system, 205 of which aren't semantic version compliant (after stripping the leading `v` and `V`). There are 1243 total instances of those version numbers. The count of each those !semantic versions ranges from 1 to 9 with 1 being the statistical mode. 

Here's the whole list of !semantic versions ...

```
(Epdb) sorted(imap.items(), key=lambda x: x[1])
[('2.3', '1'), ('1.4.3.1', '1'), ('2.7', '1'), ('v1.0.7.2', '1'), ('17.2.3.post1', '1'), ('2.3.1.1', '1'), ('v2.4', '1'), ('0.10', '1'), ('presudo', '1'), ('openshift-ansible-3.0.2-1', '1'), ('v.0.0.2', '1'), ('V.1.1.2', '1'), ('v2.7.7.1', '1'), ('v20', '1'), ('v.1.3.0', '1'), ('beta', '1'), ('1.2.0_OWN_EXPORTER', '1'), ('ansible_galaxy_fixes', '1'), ('release-4.5.0-rc3', '1'), ('v1.0-alpha', '1'), ('17.03.01', '1'), ('v1.06', '1'), ('17.2.12b2', '1'), ('docker', '1'), ('0.6.1beta', '1'), ('v1.12', '1'), ('v1.50', '1'), ('a.2.4.0', '1'), ('1.0.0.beta.8', '1'), ('test_tag', '1'), ('ansible-asb-modules-0.0.2-1', '1'), ('0.12', '1'), ('pgsql-9.5', '1'), ('1.3.1.2', '1'), ('v11.3', '1'), ('v.1.0.13', '1'), ('1.4.0.1', '1'), ('v2.10', '1'), ('atomics11.1', '1'), ('v4.9', '1'), ('1.11', '1'), ('1.0-beta', '1'), ('1.0b1', '1'), ('v2.0.0rc1', '1'), ('0.15.2.1', '1'), ('v1.08', '1'), ('2', '1'), ('iambryancs.repo-webtatic_v1.0.0', '1'), ('v0.1.0pre', '1'), ('v6.2', '1'), ('v.0.3.1', '1'), ('v2', '1'), ('pre-role', '1'), ('3.2', '1'), ('1.1-cybera', '1'), ('2.2', '1'), ('0.1.0.a1', '1'), ('v0.15', '1'), ('v.1.6.9', '1'), ('v2016.07.0', '1'), ('v0.12', '1'), ('V1.2', '1'), ('c7-kickoff', '1'), ('version', '1'), ('0.9-beta', '1'), ('17.04.17', '1'), ('v2.18', '1'), ('vTEST1', '1'), ('1.8.5.0', '1'), ('2017.1', '1'), ('1.9-tr2', '1'), ('mitaka_1', '1'), ('17.2.4-9024-20171127.023607', '1'), ('v.1.0.1', '1'), ('6.3.0.1', '1'), ('v1.09', '1'), ('v5.0', '1'), ('test', '1'), ('v2.3.2.1', '1'), ('2.4', '1'), ('v0.102', '1'), ('1.12', '1'), ('v0.2-beta.1', '1'), ('2.5', '1'), ('2018.01.01', '1'), ('v0.1.10.pre', '1'), ('0.9.10.1', '1'), ('7.3', '1'), ('v0.11', '1'), ('v2.1.1.1', '1'), ('ansible-kubernetes-modules-0.3.1-2', '1'), ('v4.0', '1'), ('develop', '1'), ('18.01.14-13_31', '1'), ('2.0.1.1', '1'), ('v3.5', '1'), ('v1.9', '1'), ('2017-04-21', '1'), ('v1.5-1', '1'), ('v1.0.0.1', '1'), ('v1.0.2d', '1'), ('v0.17', '1'), ('3.3.2.2', '1'), ('breakpad', '1'), ('v2.0beta', '1'), ('2017.04.02.22.04', '1'), ('v2.12', '1'), ('v0.9alpha', '1'), ('v.0.1.3', '1'), ('5.0', '1'), ('dev', '1'), ('r1.0.0_v1.0.2', '1'), ('v2.5', '1'), ('v.0.1.0', '1'), ('v0.1.0.1', '1'), ('1.8.171.1', '1'), ('upload-exxo-bz2', '1'), ('3.4.5.1', '1'), ('skeleton', '1'), ('v01.00.001', '1'), ('v2.6.1_1.0', '1'), ('v.1.0.0', '1'), ('0.19', '1'), ('v.2.3.1', '1'), ('14.0.0.0b1', '1'), ('deep-dev', '1'), ('0.0.4.1', '1'), ('v3.8.0.1', '1'), ('v8u171', '1'), ('v.1.6.2', '1'), ('1.1.1.1', '1'), ('0.1a', '1'), ('v1.2.0.0', '1'), ('5.5', '1'), ('vEXPR', '1'), ('1.4', '10'), ('v0.5', '10'), ('v0.1', '103'), ('v0.4', '11'), ('v6.3', '12'), ('0.9', '12'), ('0.4', '12'), ('latest', '12'), ('2.0', '12'), ('0.3', '12'), ('1.0', '122'), ('v0.7', '15'), ('0.5', '16'), ('v0.2', '17'), ('v0.3', '17'), ('v3.2', '2'), ('v.0.0.1', '2'), ('v0.4beta', '2'), ('0.15', '2'), ('1.7.2.1', '2'), ('v.1.1.4', '2'), ('1.0.0.beta.1', '2'), ('v.1.2.4', '2'), ('v1.11', '2'), ('v1.6', '2'), ('v1.0.0_r044', '2'), ('1.8', '2'), ('1.2', '20'), ('v1.0', '218'), ('1.3', '22'), ('v1.2', '27'), ('v2.0', '28'), ('0.2', '28'), ('3.1', '3'), ('0.11', '3'), ('v2.3', '3'), ('v1', '3'), ('1.7', '3'), ('0.0.1.1', '3'), ('v0.1.0.pre', '3'), ('0', '3'), ('0.8', '3'), ('1604-compability', '3'), ('v2.2', '3'), ('v3.1', '4'), ('v1.7', '4'), ('v0.9.1.1', '4'), ('v0.10', '4'), ('0.0', '4'), ('1.6', '4'), ('0.7', '4'), ('1.9', '4'), ('indigo_2', '4'), ('v1.1', '47'), ('1.1', '49'), ('1.5', '5'), ('v0.8', '5'), ('1', '5'), ('v0.9', '6'), ('devel-20170405', '6'), ('2.1', '6'), ('v3.0', '7'), ('v1.4', '8'), ('v1.5', '8'), ('v2.1', '8'), ('0.1', '83'), ('2015.1', '9'), ('0.6', '9'), ('v0.6', '9'), ('v1.3', '9')]
```

So yes this situation is a very small edge case. The old code skips adding them https://github.com/ansible/galaxy/blob/devel/galaxy/worker/tasks/repository.py#L337-L344 so they must have gotten in earlier in old-old galaxy's history like you mentioned. 

I think the best we can do for this is to ignore those strings for existing versions but continue NOT adding them. I'll revise the patch, but I still need to try to include removal of the incorrect versions like you got for "master"/"main" on your last import(s).

galaxy_ng/app/api/v1/tasks.py
gotmax23: Ah, I missed the `.lower()`

```suggestion
    return semantic_version.Version("vV")
```

would do that, but it would also remove multiple leading `v`s, which might not be what you want.

/resolve

geerlingguy:
@jctanner sounds like a plan. I wonder if you also cross-referenced those unique instances with download count, if any of those roles have even been downloaded recently (or much at all). I would say that's a reasonable way to deal with the oddballs.

jctanner:
@geerlingguy re your comment about repo-epel ...

I disagree with the decision made previously to replace hyphens with underscores. I attribute (without having researched yet) the decision to trying to make everything more like collections, which roles definitely are not. I intentionally did not re-implement the character swapping in this PR. However, I did test your repo-epel role locally just to make sure and it seemed to do what I expect. As long as you keep the same value in https://github.com/geerlingguy/ansible-role-repo-epel/blob/master/meta/main.yml#L5 and don't pass something else for `--role-name` in a future import, the name shall remain the same.

richm:
Thanks - how long until this shows up in galaxy?

jctanner:
The code is live on galaxy-dev.ansible.com right now, so I'll be running some tests/QE there before we decide to push it to galaxy.ansible.com. I'd give it by the end of day today or tomorrow depending on how it goes.

I'll also have to craft some cleanup for the bad role versions we made prior to this patch. Theoretically this new code will auto-cleanup once someone re-imports their role(s), but I'm going to have to take care of what I can for those that don't want to re-import.

richm:
But https://galaxy-dev.ansible.com/ui/standalone/roles/linux-system-roles/timesync versions are sorted correctly in descending numeric order

richm:
> https://galaxy-dev.ansible.com/ui/standalone/roles/willshersystems/sshd/ - the versions are not sorted

In github, the tags have the `v` prefix e.g. `v0.22.0` - but in linux system roles we use a strictly numeric version/tag

jctanner:
I've opened https://github.com/ansible/galaxy_ng/pull/1953 to re-add sorting at view/serialization time and also to allow versions with -any- numerical component.

Re https://galaxy-dev.ansible.com/ui/standalone/roles/willshersystems/sshd/ showing or not showing the "v" prefix on the versions. There's some concepts under the covers that I noted in this PR's description.

* the "name" field for a role version is the original tag name.
* the "version" field for a role version is the tag name stripped of any 'v' prefix.

So with that said, here's the API data for that role ...

```[jtanner@p1 ~]$ curl -s https://galaxy-dev.ansible.com/api/v1/roles/30752/versions/ | jq -c '.results[] | {name: .name, vesion: .version}' | head
{"name":"v0.6.1","vesion":"0.6.1"}
{"name":"v0.12.0","vesion":"0.12.0"}
{"name":"v0.2.0","vesion":"0.2.0"}
{"name":"v0.5.1","vesion":"0.5.1"}
{"name":"v0.7.6","vesion":"0.7.6"}
{"name":"v0.7.3","vesion":"0.7.3"}
{"name":"v0.4.3","vesion":"0.4.3"}
{"name":"v0.17.0","vesion":"0.17.0"}
{"name":"v0.16.1","vesion":"0.16.1"}
{"name":"v0.11.1","vesion":"0.11.1"}
```

What you are seeing in the UI is a mix of the "name" on on the summary page(s) and the "version" on the versons pages. We should definitely make that consistent in the UI code.

PR #1969, opened 2023-11-07
Add the github id to the v1/user serializer for auditing purposes.

jctanner:
To match data from old and new galaxy, we need the github id for each user. It's squirreled away in the social auth table and can be fetched by filtering on the user.

### DIFF
--- a/galaxy_ng/app/api/v1/serializers.py
+++ b/galaxy_ng/app/api/v1/serializers.py
@@ -2,8 +2,6 @@ import datetime
 
 from rest_framework import serializers
 
-from social_django.models import UserSocialAuth
-
 from pulpcore.plugin.util import get_url
 
 from galaxy_ng.app.models.auth import User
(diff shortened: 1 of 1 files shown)


galaxy_ng/app/api/v1/serializers.py
```diff
@@ -178,6 +181,12 @@ def get_avatar_url(self, obj):
         url = f'https://github.com/{username}.png'
         return url
 
+    def get_github_id(self, obj):
```
rochacbruno, line 184: You going to catch exception here because USerSocialAuth migration runs only for community profile https://github.com/ansible/galaxy_ng/blob/master/galaxy_ng/app/dynaconf_hooks.py#L197 

This query will fail for other profiles because the table doesnt exist.

Or you do EAFP (try/except) or LBYL (if "social_auth" in settings.INSTALLED_APPS)

jctanner: I didn't code for that because the entire v1 api isn't added to the router unless social auth is enabled. I'll add it anyway.

rochacbruno: So maybe the problem is at import time, it still errors not finding the table.

jctanner: Possibly, I'll look into it.

rochacbruno (approved):

PR #1974, opened 2023-11-13
[WIP] content surveys

jctanner:


galaxy_ng/app/api/v3/serializers/survey.py
```diff
@@ -0,0 +1,54 @@
+from rest_framework import serializers
+
+from galaxy_ng.app.models import (
+    CollectionSurveyRollup,
+    LegacyRoleSurveyRollup,
+)
+
+
+class CollectionSurveyRollupSerializer(serializers.ModelSerializer):
+
```
rochacbruno, line 10: If we make a Foreign Key with Collection, we gain the ability to have cascade deletion of survey data when collection is deleted.

If not, then we might need to handle the cleanup separately (the same problem we need to address with aiindex)

PR #1979, opened 2023-11-15
fix(community): Search index for roles

rochacbruno:
#### What is this PR doing:

Add Search index for roles in a separate table instead of building it at runtime


galaxy_ng/app/migrations/0047_update_role_search_vector_trigger.py
```diff
@@ -0,0 +1,64 @@
...
+        DO UPDATE SET
+            search_vector = _search_vector, modified = current_timestamp;
+    RETURN NEW;
+END;
+$$
+LANGUAGE plpgsql;
+
+DROP TRIGGER IF EXISTS update_ts_vector ON galaxy_legacyrole;
+
+CREATE TRIGGER update_ts_vector
```
fabriziomello, line 37: Since PG14 we support `OR REPLACE` option on `CREATE TRIGGER` statement: https://www.postgresql.org/docs/current/sql-createtrigger.html

galaxy_ng/app/migrations/0047_update_role_search_vector_trigger.py
```diff
@@ -0,0 +1,64 @@
...
+    AFTER INSERT OR UPDATE
+    ON galaxy_legacyrole
+    FOR EACH ROW
+EXECUTE PROCEDURE update_role_ts_vector();
```
fabriziomello, line 41:
```suggestion
EXECUTE FUNCTION update_role_ts_vector();
```

galaxy_ng/app/migrations/0047_update_role_search_vector_trigger.py
```diff
@@ -0,0 +1,64 @@
...
+END;
+$$
+LANGUAGE plpgsql;
+
+DROP TRIGGER IF EXISTS update_ts_vector ON galaxy_legacyrole;
+
+CREATE TRIGGER update_ts_vector
+    AFTER INSERT OR UPDATE
+    ON galaxy_legacyrole
+    FOR EACH ROW
```
fabriziomello, line 40: I don't know about your data model but you can define some rules for when firing triggers using the `WHEN (condition)` statements. So you can evaluate if is worth using it to don't unecessary fire the trigger.

rochacbruno: In this case we want to update the search_vector whenever there is any change in the table.

galaxy_ng/app/migrations/0047_update_role_search_vector_trigger.py
rochacbruno: Is there any advantage on using FUNCTION instead of PROCEDURE? or is it just to not be confusing with stored procedures?

galaxy_ng/app/migrations/0047_update_role_search_vector_trigger.py
rochacbruno: I have to confirm, but I think we have some PG12 running somewhere.

galaxy_ng/app/migrations/0047_update_role_search_vector_trigger.py
fabriziomello: In your case both PROCEDURE or FUNCTION will work as expected. The difference is in Postgres PROCEDUREs you can do transaction control inside (aka COMMIT/ROLLBACK). But in this case I guess you want to be full atomic so just to don't confusing people because you created a function and then in the trigger definition you're referencing a procedure.

jctanner (approved):

PR #2009, opened 2023-12-12
Add a provider filter for legacy namespaces.

jctanner:
https://issues.redhat.com/browse/AAH-2870


himdel (approved):
Works for me :+1: 

`GET /api/v1/namespaces/?provider=gh_0x0i`

filters by provider namespace name :)

PR #2032, opened 2024-01-15
Update remote rh-certified default url

drodowic:
#### What is this PR doing:

* Updates the remote rh-certified default url

AAP-15623

#### Reviewers must know:

### DIFF
--- a/galaxy_ng/app/migrations/0048_update_collection_remote_rhcertified_url.py
+++ b/galaxy_ng/app/migrations/0048_update_collection_remote_rhcertified_url.py
@@ -1,0 +1,29 @@
+from django.db import migrations
+
+def update_collection_remote_rhcertified_url(apps, schema_editor):
+    """
+    Updates the existing collection remote `rh-certified` url field  
+    to add `content/published/`.
+    """
+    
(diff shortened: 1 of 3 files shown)


jctanner (approved):

jctanner:
Are we backporting this or should those labels be removed?

jctanner (changes requested):
If a system has already run this migration, it won't run it again if the file changed.  I think it'd be better to make a new migration file which is sure to be run.

drodowic:
> If a system has already run this migration, it won't run it again if the file changed. I think it'd be better to make a new migration file which is sure to be run.

The intent in modifying an existing migration is so that only new installations will get the updated URL so that we don't impact customers who might have already modified it.

jctanner:
> > If a system has already run this migration, it won't run it again if the file changed. I think it'd be better to make a new migration file which is sure to be run.
> 
> The intent in modifying an existing migration is so that only new installations will get the updated URL so that we don't impact customers who might have already modified it.

A new migration file can do a conditional update by checking the current value.

jctanner:
With the new migration file, we can remove the change in the old file right? Also, should the backport labels be removed?

jctanner (approved):

PR #2039, opened 2024-01-19
Fix create-user management command

cutwater:
#### What is this PR doing:

* Do not display message "User already exists" when used is actually created for the first time.
* Remove redundant call of `user.save()`. The record is already saved to the database when the `create()` method is called.
* Remove unused variables.

No-Issue


jctanner (approved):
Seems fine, but couldn't most of this be replaced with User.objects.get_or_create() ?

cutwater:
@jctanner You're absolutely right and this makes code shorter and cleaner. Thanks, fixed.

rochacbruno (approved):

rochacbruno (approved):

PR #2041, opened 2024-01-22
Bump pulp* dependencies

rochacbruno:
1. Bumped pulpcore, pulp-ansible, pulp_container and dynaconf on setup.py
2. executed `make requirements/pip-upgrade-all` to generate all the requirements/ lock files.
3. executed on `plugin_template` folder `./plugin-template --github galaxy_ng` to upgrade all the CI to use latest changes.

CI is failing probably because:

- Workflow workdir changed, so post-job or pre-job templates must be fixed to include the fix for script paths
- there are still some pip install conflict or missing dependency, need to identify which is the conflict to adjust the pinnings or identify which is the missing lib to add
- maybe, template-config.yaml needs some adjust to be compliant with the new changes on CI


rochacbruno:
Ci changes made on #2087 

Pulp bumps coming on a separate PR

PR #2071, opened 2024-02-15
Add organization and team models

cutwater:
* Add new `Oranization` and `Team` models, that implement respective abstract models from DAB.
* Add related database migrations and data migration, that creates a new default model.
* Implement custom manager for the `Organization` model that retrieves the default organization.

No-Issue


galaxy_ng/app/models/organization.py
```diff
@@ -0,0 +1,34 @@
...
+    )
+
+    objects = OrganizationManager()
+
+
+class Team(AbstractTeam):
+    """A team model."""
+
+    users = models.ManyToManyField(
+        settings.AUTH_USER_MODEL,
```
newswangerd, line 31: I'd recommend using the `get_user_model()` function that django provides for this.

newswangerd (approved):

galaxy_ng/app/models/organization.py
cutwater: I don't think it's the case here and actually Django documentation explicitly states to use `settings.AUTH_USER_MODEL` in foreign keys.

https://docs.djangoproject.com/en/5.0/topics/auth/customizing/

> When you define a foreign key or many-to-many relations to the user model, you should specify the custom model using the [AUTH_USER_MODEL](https://docs.djangoproject.com/en/5.0/ref/settings/#std-setting-AUTH_USER_MODEL) setting.

However there is vague statement, that `get_user_model` may be used instead, though it is not obvious which benefits it gives in this specific case.

> Generally speaking, it’s easiest to refer to the user model with the [AUTH_USER_MODEL](https://docs.djangoproject.com/en/5.0/ref/settings/#std-setting-AUTH_USER_MODEL) setting in code that’s executed at import time, however, it’s also possible to call `get_user_model()` while Django is importing models, so you could use `models.ForeignKey(get_user_model(), ...)`.

The `get_user_model()` function call is useful where you need to use a model instance, not a string. For instance to make a query:

```
User = get_user_model()
User.objects.get(...)
```

PR #2081, opened 2024-02-21
Chmod +x entrypoint and readyz

rcarrillocruz:
We leverage both entrypoint and readyz scripts in downstream images, and we have to chmod +x at Docker build time.


rochacbruno (approved):

rochacbruno:
commit is not signed

PR #2084, opened 2024-02-28
gateway

chr-stian:
tests have been updated so that they all use galaxykit and can authenticate with the gateway


jctanner:
This is fine, but the integration tests need to pass for all oci deployment types.

jctanner (approved):

PR #2085, opened 2024-02-28
Do not include group members when fetching the list of users for a namespace

jctanner:
https://issues.redhat.com/browse/AAH-3121


drodowic (approved):

bmclaughlin (approved):

jctanner:
The failing test_namespace_edit_logo test works for me locally. I suspect it's failing in CI because of timing and performance problems with whatever async tasks it kicks off.

PR #2094, opened 2024-03-07
Add organization and team models

cutwater:
* Add new `Oranization` and `Team` models, that implement respective abstract models from DAB.
* Add related database migrations and data migration, that creates a new default model.
* Implement custom manager for the `Organization` model that retrieves the default organization.
* Team model has a one-to-one relationship with the standard Group model.
* When new Team record is created, a related Group model is automatically created as well.

No-Issue


newswangerd (approved):

newswangerd (approved):

AlanCoding:
Organization & team admins & members are implemented as a role in both DAB RBAC.

Yes, they are implemented as relationships in Gateway. It will make our lives much easier if they never need to co-exist. I added a sub-feature of DAB RBAC to cope with that co-existence, but I had plans to remove it - https://github.com/ansible/django-ansible-base/issues/159

Looking at this, it seems that https://github.com/ansible/django-ansible-base/issues/159 may need to be closed, as with the way this is going, org admins may co-exist as both a relationship and a role at some point.

PR #2104, opened 2024-03-25
Add make target and compose file for DAB.

jctanner:
This makes it so we can start the DAB profile just like we can all the others.


rochacbruno (approved):

dev/oci_env_integration/oci_env_configs/dab.compose.env
```diff
@@ -0,0 +1,15 @@
+COMPOSE_PROFILE=galaxy_ng/base:galaxy_ng/dab
+COMPOSE_PROJECT_NAME=ci-dab
+
+DEV_SOURCE_PATH=galaxy_ng
```
rochacbruno, line 4: @jctanner shouldn't it also add django_ansible_base as a DEV_SOURCE_PATH? is this profile intended for DAB development?

rochacbruno: nm, just figured out that this is for CI

PR #2107, opened 2024-03-27
Switch collection testing over to oci-env.

jctanner:
collectstatic is broken in the old dev stack (thereby breaking the collection test), so this is an immediate need.


rochacbruno (approved):

jctanner:
other integration tests are failing, but it's not due to this PR.

PR #2109, opened 2024-03-28
Fix openapi client generation in the playbooks.

jctanner:
Functional tests were failing because the VERSION string for the generated clients was an empty string which in turn causes pip to fail the install.


jctanner:
sigh, it's still not working here?

PR #2111, opened 2024-03-30
Use an artifact server for hosting the logo images.

jctanner:


jctanner:
given https://github.com/ansible/galaxy_ng/pull/2112, we probably don't need this.

PR #2116, opened 2024-04-05
Pin the base image to resolve oci-env incompatibility.

jctanner:
https://github.com/pulp/oci_env/pull/155#issuecomment-2039739002


jctanner:
I guess we don't need this anymore.

PR #2122, opened 2024-04-10
Fix test_artifact_upload.py::test_ansible_lint_exception_AAH_2606

bmclaughlin:
ansible-lint no longer returns a warning on one of the warnings being tested against.

No-Issue


jctanner (changes requested):
We need to remove the skip decorator too, right?

jctanner (approved):

PR #2120, opened 2024-04-09
Use tmp dir in build_deploy.sh

bsquizz:
Use temp dir instead of $WORKSPACE to store docker config in build_deploy

No-Issue

PR #2123, opened 2024-04-12
Use tmp dir in build_deploy.sh

drodowic:
#### What is this PR doing:
Use temp dir instead of $WORKSPACE to store docker config in build_deploy

No-Issue


bsquizz:
Already opened in #2120

PR #2134, opened 2024-04-30
dynamic API_HOSTNAME & CONTENT_ORIGIN via dynaconf+django-crum

jctanner:
The end result of this is that the download_url for a collection version and the redirected urls from those url will resemble whatever protocol+host your client connected to the server with. This should enable downloadable collections from the hub AND through a gateway at the same time.


dev/oci_env_integration/oci_env_configs/dab.compose.env
```diff
@@ -1,7 +1,7 @@
 COMPOSE_PROFILE=galaxy_ng/base:galaxy_ng/dab
 COMPOSE_PROJECT_NAME=ci-dab
 
-DEV_SOURCE_PATH=galaxy_ng
+DEV_SOURCE_PATH=galaxy_ng:pulp_ansible
```
rochacbruno, line 4: does DAB compose need pulp_ansible as dev source?

galaxy_ng/app/dynaconf_hooks.py
```diff
@@ -644,6 +645,7 @@ def configure_dynamic_settings(settings: Dynaconf) -> Dict[str, Any]:
     change the value before it is returned allowing reading overrides from
     database and cache.
     """
+    print(f'configure_dynamic_settings ... {settings.get("GALAXY_DYNAMIC_SETTINGS")}')
```
rochacbruno, line 648: we can remove the prints?

dev/oci_env_integration/oci_env_configs/dab.compose.env
jctanner: Nope, we just need a new release (hopefully today) of pulp_ansible before I can take this out.

galaxy_ng/app/dynaconf_hooks.py
jctanner: Yeah, once we're sure this is the path forward architecturally

galaxy_ng/app/dynaconf_hooks.py
```diff
@@ -709,8 +712,42 @@ def read_settings_from_cache_or_db(
...
+        ALLOWED_KEYS = ['CONTENT_ORIGIN', 'ANSIBLE_API_HOSTNAME']
+
+        # if 'CACHE' in key.upper():
+        #    print(f'PID:{os.getpid()} READY:{apps.ready} KEY:{key} VAL:{value.value}')
+
+        if not apps.ready or key.upper() not in DYNAMIC_SETTINGS_SCHEMA:
+            # If app is starting up or key is not on allowed list bypass and just return the value
+            if key.upper() in ALLOWED_KEYS:
+                print(f"SKIPPING {key}")
+            return value.value
```
rochacbruno, line 732: If the only two keys you are interestes are those 2 this part can be simplified to

```suggestion
        if not apps.ready or key.upper() not in ALLOWED_KEYS:
            return value.value
```

galaxy_ng/app/dynaconf_hooks.py
```diff
@@ -709,8 +712,42 @@ def read_settings_from_cache_or_db(
...
+
+        if key.upper() in ALLOWED_KEYS:
```
rochacbruno, line 734: this condition was already checked above, when execution reaches this line it is assumed the key.upper() is in ALLOWED_KEYS

galaxy_ng/app/dynaconf_hooks.py
```diff
@@ -709,8 +712,42 @@ def read_settings_from_cache_or_db(
...
+            req = get_current_request()
+            if req is not None:
+                headers = dict(req.headers)
+                print(f'REQ: {req}')
+                print(f'REQ.headers: ...')
```
rochacbruno, line 739: remove prints?

rochacbruno: or replace with logging

galaxy_ng/app/dynamic_settings.py
```diff
@@ -62,4 +62,6 @@
     "AUTOMATION_ANALYTICS_LAST_ENTRIES": {},
     "FOO": {},
     "BAR": {},
+    'ANSIBLE_API_HOSTNAME': {},
+    'CONTENT_ORIGIN': {}
```
rochacbruno, line 66: This will allow those 2 keys to be changed via database, is that desired?

galaxy_ng/app/dynaconf_hooks.py
```diff
@@ -709,8 +712,37 @@ def read_settings_from_cache_or_db(
...
+                host = headers.get('Host', 'localhost:5001')
+                baseurl = proto + '://' + host
+                return baseurl
+
+        return temp_settings.get(key, value.value)
+
     return {
         "_registered_hooks": {
-            Action.AFTER_GET: [Hook(read_settings_from_cache_or_db)]
+            Action.AFTER_GET: [Hook(read_settings_from_cache_or_db), Hook(alter_hostname_settings)]
```
rochacbruno, line 746: We can probably change these lines, and have the hooks to be registered conditionally, instead of using the `GALAXY_DYNAMIC_SETTINGS` check above.

Something like:

On settings:

```py
DYNACONF_AFTER_GET_HOOKS=["read_settings_from_cache_or_db", "alter_hostname_settings"]
```

Here:

```python
# replace the check for dynamic_settings above
enabled_hooks = settings.get("DYNACONF_AFTER_GET_HOOKS")
if not enabled_hooks:
    return

...

return {
   "_registered_hooks": {
       Action.AFTER_GET: [Hook(locals()[function]) for function in enabled_hooks]
   }
}
```

the use of `locals` require that hook function is always defined inside the dynaconf_hooks file.

galaxy_ng/app/dynaconf_hooks.py
```diff
@@ -709,8 +712,37 @@ def read_settings_from_cache_or_db(
...
+        if key.upper() in ALLOWED_KEYS:
+            req = get_current_request()
+            if req is not None:
+                headers = dict(req.headers)
+                proto = headers.get('X-Forwarded-Proto', 'http')
+                host = headers.get('Host', 'localhost:5001')
+                baseurl = proto + '://' + host
+                return baseurl
+
+        return temp_settings.get(key, value.value)
```
rochacbruno, line 742: The `if` condition is redundant here, already checked above,
and as there is no merging of data structure involved (dicts,lists) we can
just return value, no need to build a temp_settings.

```suggestion
        
        req = get_current_request()
        if req is not None:
            headers = dict(req.headers)
            proto = headers.get('X-Forwarded-Proto', 'http')
            host = headers.get('Host', 'localhost:5001')
            baseurl = proto + '://' + host
            return baseurl
        else:
            return value.value
```

rochacbruno: we could just put a final `return value.value` however using the else is more readable IMO

PR #2143, opened 2024-05-14
remove manifest update scripts

drodowic:
#### What is this PR doing:

Clean up old manifest update scripts now that we have frontend containers for insights.

No-Issue


himdel (approved):

himdel:
:+1: 
This also makes `.github/workflows/scripts/deploy_manifest.gpg` obsolete, right?

drodowic:
> This also makes `.github/workflows/scripts/deploy_manifest.gpg` obsolete, right?

Good catch. Yes, updated the PR to remove this file as well.

PR #2150, opened 2024-05-29
Enable gateway tests w/ dab_jwt

jctanner:


profiles/dab_jwt/run_integration.sh
```diff
@@ -21,7 +21,9 @@ export HUB_ADMIN_PASS=admin
 export HUB_USE_MOVE_ENDPOINT=1
 export HUB_LOCAL=1
 export ENABLE_DAB_TESTS=1
-export HUB_TEST_MARKS="deployment_standalone"
+export HUB_TEST_MARKS="deployment_standalone and not skip_in_gw"
```
chr-stian, line 24: "and not package and not iqe_ldap" I have also this in my tests, not sure if they should be excluded here thought

chr-stian (approved):

profiles/dab_jwt/run_integration.sh
jctanner: hrm, lets check this in another PR if necessary. ATM all tests are passing, so the ldap stuff must be skipped.

bmclaughlin (approved):

PR #2157, opened 2024-06-06
Add a system_auditor role who has read on everything.

jctanner:
https://issues.redhat.com/browse/AAP-24730

```
jtanner@p1:~/workspace/github/jctanner.redhat/galaxy_ng.system_auditor$ curl -L -u bob:redhat1234 http://localhost:55001/api/galaxy/_ui/v1/users/ | jq .                                                                                                                                                                      
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   117  100   117    0     0    338      0 --:--:-- --:--:-- --:--:--   339
{
  "errors": [
    {
      "status": "403",
      "code": "permission_denied",
      "title": "You do not have permission to perform this action."
    }
  ]
}
jtanner@p1:~/workspace/github/jctanner.redhat/galaxy_ng.system_auditor$ curl -L -u sa:redhat1234 http://localhost:55001/api/galaxy/_ui/v1/users/ | jq .                                                                                                                                                                      
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   780  100   780    0     0   2090      0 --:--:-- --:--:-- --:--:--  2091
{
  "meta": {
    "count": 3
  },
  "links": {
    "first": "/api/galaxy/_ui/v1/users/?limit=10&offset=0",
    "previous": null,
    "next": null,
    "last": "/api/galaxy/_ui/v1/users/?limit=10&offset=0"
  },
  "data": [
    {
      "id": 1,
      "username": "admin",
      "first_name": "",
      "last_name": "",
      "email": "admin@example.com",
      "groups": [],
      "date_joined": "2024-06-06T19:46:09.181235Z",
      "is_superuser": true,
      "auth_provider": [
        "django"
      ]
    },
    {
      "id": 2,
      "username": "sa",
      "first_name": "system",
      "last_name": "auditor",
      "email": "sa@localhost",
      "groups": [
        {
          "id": 1,
          "name": "sa-group"
        }
      ],
      "date_joined": "2024-06-06T19:59:29.183763Z",
      "is_superuser": false,
      "auth_provider": [
        "django"
      ]
    },
    {
      "id": 3,
      "username": "bob",
      "first_name": "bob",
      "last_name": "barker",
      "email": "bob@localhost",
      "groups": [],
      "date_joined": "2024-06-06T20:12:02.936366Z",
      "is_superuser": false,
[...]

PR #2162, opened 2024-06-10
system auditor role

jctanner:
https://issues.redhat.com/browse/AAP-24730

replaces https://github.com/ansible/galaxy_ng/pull/2157

```
[root@35f0b6ad6e2c src]# pulpcore-manager shell < script.py 

Collection Namespaces
        ADD :: 467 :: Can view namespace :: view_namespace

Collections
        ADD :: 17 :: Can view collection :: view_collection
        ADD :: 29 :: Can view ansible repository :: view_ansiblerepository

Users
        no changes needed

Groups
        no changes needed

Collection Remotes
        no changes needed

Ansible Repository
        no changes needed

Execution Environments
        ADD :: 161 :: View any push repository in a namespace :: namespace_view_containerpushrepository
        ADD :: 135 :: Can view container repository :: view_containerrepository
        ADD :: 154 :: Can view container namespace :: view_containernamespace
        ADD :: 508 :: Can view container namespace :: view_containernamespace
        ADD :: 157 :: View any distribution in a namespace :: namespace_view_containerdistribution

Container Registry Remotes
        ADD :: 500 :: Can view container registry remote :: view_containerregistryremote

Task Management
        no changes needed
```

image

### DIFF
--- a/galaxy_ng/app/constants.py
+++ b/galaxy_ng/app/constants.py
@@ -66,10 +66,10 @@ PERMISSIONS = {
         "global_description": _("Delete any existing collection."),
         "ui_category": _("Collections"),
     },
-    "ansible.view_ansible_repo_content": {
+    "ansible.view_ansiblerepository": {
         "name": _("View Ansible repo content"),
-        "object_description": _("View content of this Ansible repository."),
-        "global_description": _("View collections in any existing namespace."),
+        "object_description": _("View this Ansible repository."),
+        "global_description": _("View this Ansible repository."),
         "ui_category": _("Collections"),
     },
     "ansible.modify_ansible_repo_content": {
@@ -247,6 +247,12 @@ PERMISSIONS = {
         "object_description": _("Manage container namespace roles."),
         "global_description": _("Manage container namespace roles existing in the system."),
         "ui_category": _("Execution Environments"),
+    },
+    "galaxy.view_containerregistryremote": {
+        "name": _("View remote registry"),
+        "object_description": None,
+        "global_description": _("View remote registries in the system."),
+        "ui_category": _("Container Registry Remotes"),
     },
     "galaxy.add_containerregistryremote": {
         "name": _("Add remote registry"),
--- a/galaxy_ng/app/constants.py
+++ b/galaxy_ng/app/constants.py
@@ -54,10 +54,22 @@ PERMISSIONS = {
         "global_description": _("Upload collections to any existing namespace."),
         "ui_category": _("Collection Namespaces"),
     },
+    "ansible.view_collection": {
+        "name": _("View collection"),
+        "object_description": _("View this collection."),
+        "global_description": _("View any existing collection."),
+        "ui_category": _("Collections"),
+    },
     "ansible.delete_collection": {
         "name": _("Delete collection"),
         "object_description": _("Delete this collection."),
         "global_description": _("Delete any existing collection."),
+        "ui_category": _("Collections"),
+    },
+    "ansible.view_ansible_repo_content": {
+        "name": _("View Ansible repo content"),
+        "object_description": _("View content of this Ansible repository."),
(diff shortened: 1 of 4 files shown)


galaxy_ng/app/constants.py
```diff
@@ -30,6 +30,12 @@ class DeploymentMode(enum.Enum):
         # Category to group the permission in the UI.
         "ui_category": _("Collection Namespaces"),
     },
+    "galaxy.view_namespace": {
```
AlanCoding, line 33: If you add a new permission here, would you need to add it to existing roles like `galaxy.collection_namespace_owner` which currently offers `galaxy.change_namespace`? Would some other API access logic need to be rewritten to utilize the new permission?

galaxy_ng/app/constants.py
```diff
@@ -182,18 +200,36 @@ class DeploymentMode(enum.Enum):
         ),
         "ui_category": _("Ansible Repository"),
     },
+    "container.view_containernamespace": {
```
newswangerd, line 203: A lot of the permissions you've added here don't actually do anything. Namespaces, for example, are viewable to everyone who is authenticated. We don't have the notion of a private namespace. I'd like to keep this file limited to the set of permissions that are actually enforced.

I think these will start to make sense once we fully integrate organizations, but for now this will just make the UI confusing.

galaxy_ng/app/constants.py
```diff
@@ -48,12 +54,24 @@ class DeploymentMode(enum.Enum):
...
+        "global_description": _("View any existing collection."),
+        "ui_category": _("Collections"),
+    },
     "ansible.delete_collection": {
         "name": _("Delete collection"),
         "object_description": _("Delete this collection."),
         "global_description": _("Delete any existing collection."),
         "ui_category": _("Collections"),
     },
+    # "ansible.view_ansiblerepository": {
```
newswangerd, line 69: Why did you comment this one out? We do support private repositories, so view permissions here should matter.

jctanner: Because this data structure is a dictionary and the key can only have one instance, which makes it impossible to add the same key to multiple categories.

galaxy_ng/app/constants.py
jctanner: I need a discrete list of what's needed in that case. I had to write a script to find the missing "read" perms for each category because there's no pointers otherwise.

newswangerd: This file is the discrete list of what's needed. That's why I sent it to you. These are the permissions we care about.

galaxy_ng/app/constants.py
newswangerd: Why does this need to be added to multiple categories?

galaxy_ng/app/constants.py
newswangerd: @AlanCoding @jctanner This file should contain the set of permissions that we actually care about for enforcement. The problem is that pulp/django ship with a huge pile of random permissions, most of which we don't care about. `galaxy.view_namespaces` is a good example. Namespaces are always readable by everyone who is authenticated. We don't want to expose that permission to the UI or to users because adding it to a role does nothing.

@AlanCoding this is why I was so against using django permissions for DAB RBAC. It's practically impossible to get a full list of permissions that are actually enforced and actually do something in the system.

jctanner:
replaced by https://github.com/ansible/galaxy_ng/pull/2168

galaxy_ng/app/constants.py
AlanCoding: This led to me file https://github.com/ansible/django-ansible-base/issues/457. But it should be super minor, I want to declare a change is probably necessary in DAB RBAC for the documented reasons, but I'm completely confident that won't be challenging to do. We've just haven't done it yet.

PR #2165, opened 2024-06-11
skip if jira is open and hub from aap installation

chr-stian:
No-Issue


chr-stian:
@jctanner they are running on both profiles. They are only skipped when DEV_ENV_STANDALONE=0 (and I think it only happens when hub comes from an AAP installation)

jctanner (approved):

PR #2174, opened 2024-06-17
Fix: Set default value for DAB reserved names

rochacbruno:
AAP-25505
No-Issue


rochacbruno:
Do we want tos et it to `[]` or to the value coming from DAB on #2176 ? @newswangerd 

If #2176 is merged, then we can close this one.

rochacbruno:
replaced by #2176

PR #2175, opened 2024-06-17
Schedule Sync on Container Entrypoint + Better logs

rochacbruno:
Added a call to 
```bash
django-admin task-scheduler --id dab_sync --interval 15 --path "galaxy_ng.app.tasks.resource_sync.run"
```
on container entrypoint.sh, so the schedule is added on dev environment and also on all the container builds without the need for operators to call it.

No-Issue

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -104,6 +104,8 @@ run_service() {
         setup_repo_keyring &
     fi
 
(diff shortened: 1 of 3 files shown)


rooftopcellist (approved):
fwiw, I like this approach and it will make things much easier in the operator and other install methods.

PR #2183, opened 2024-06-20
Override the token server setting with crum also.

jctanner:
The TOKEN_SERVER setting is also something that needs to not send the clients away from the original host they are connecting to.

https://issues.redhat.com/browse/AAP-24997


rochacbruno (approved):

jctanner:
failing test_can_directly_to_repo_if_pipeline_staging on the insights profile is passing locally, so chalking it up to github infra glitch.

PR #2194, opened 2024-07-15
Whitespace edits to copy docker run cmd to terminal

AlanCoding:
#### What is this PR doing:
I followed the docs link in the README.md https://ansible.readthedocs.io/projects/galaxy-ng/en/latest/ and then went to User Guides -> Installation

Pasted this in terminal:

```
docker run \                                                        
    --name=galaxy_ng \
    -v $(pwd)/galaxy-importer.cfg:/etc/galaxy-importer/galaxy-importer.cfg
    --env-file=pulp_settings.env \
    -p 8080:80 \                                                    
    quay.io/pulp/galaxy:4.9.0 
```

This didn't run because of whitespace/wrapping problems, and I believe this fixes it.

No-Issue

#### Reviewers must know:
I changed the tag as well, because `PULP_GALAXY_AUTHENTICATION_CLASSES` in that same file is outdated and causes a traceback because it references a class that doesn't exist in 4.9.0. The latest is on 4.9.2 which still doesn't work, but `:latest` seems more honest because it will at least _eventually_ work.

### DIFF
--- a/docs/usage_guide/installation.md
+++ b/docs/usage_guide/installation.md
@@ -103,12 +103,12 @@ Understanding every setting in the file is beyond the scope of this document, bu
 
 Start the container with these docker args ...
 ```
-docker run \                                                        
+docker run \
     --name=galaxy_ng \
-    -v $(pwd)/galaxy-importer.cfg:/etc/galaxy-importer/galaxy-importer.cfg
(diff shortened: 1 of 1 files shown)


rochacbruno:
I am closing this PR as it is almost one year old and we are moving away from OCI-env anyway.

PR #2202, opened 2024-07-25
Install ansible_base.rbac in the dab profile(s).

jctanner:


galaxy_ng/app/dynaconf_hooks.py
```diff
@@ -738,3 +739,17 @@ def configure_dab_required_settings(settings: Dynaconf) -> Dict[str, Any]:
...
+    data = {}
+    if not hasattr(settings, 'ANSIBLE_BASE_MANAGED_ROLE_REGISTRY'):
+        data['ANSIBLE_BASE_MANAGED_ROLE_REGISTRY'] = {}
+    if not hasattr(settings, 'ANSIBLE_BASE_ALLOW_SINGLETON_ROLES_API'):
+        data['ANSIBLE_BASE_ALLOW_SINGLETON_ROLES_API'] = True
+    if not hasattr(settings, 'ANSIBLE_BASE_ALLOW_SINGLETON_USER_ROLES'):
+        data['ANSIBLE_BASE_ALLOW_SINGLETON_USER_ROLES'] = False
+    if not hasattr(settings, 'ANSIBLE_BASE_ALLOW_SINGLETON_TEAM_ROLES'):
+        data['ANSIBLE_BASE_ALLOW_SINGLETON_TEAM_ROLES'] = False
+    data["INSTALLED_APPS"] = settings.INSTALLED_APPS[:] + ["ansible_base.rbac"]
```
rochacbruno, line 754: Using a unique merge will prevent duplication if this app is for some reason already in the list of installed apps.
```suggestion
    data["INSTALLED_APPS"] = ["ansible_base.rbac", "dynaconf_merge_unique"]
```

galaxy_ng/app/dynaconf_hooks.py
```diff
@@ -738,3 +739,17 @@ def configure_dab_required_settings(settings: Dynaconf) -> Dict[str, Any]:
         if key.isupper() and settings.get(key, notset) is notset:
             data[key] = getattr(dynamic_settings, key)
     return data
+
+
+def configure_dab_rbac_settings(settings: Dynaconf) -> Dict[str, Any]:
+    data = {}
+    if not hasattr(settings, 'ANSIBLE_BASE_MANAGED_ROLE_REGISTRY'):
```
rochacbruno, line 746: I usually avoid the [usage of](https://hynek.me/articles/hasattr/) `hasattr`
```suggestion
    if settings.get('ANSIBLE_BASE_MANAGED_ROLE_REGISTRY', None) is None:
```

galaxy_ng/app/urls.py
```diff
@@ -18,6 +18,10 @@
     urlpatterns as resource_api_urls,
 )
 
+from ansible_base.rbac.urls import (
+    api_version_urls as dab_rbac_urls,
```
AlanCoding, line 22: It's really all DAB URLs. This might add more endpoints than you're not expecting, but that is intended on the DAB side.

galaxy_ng/app/dynaconf_hooks.py
AlanCoding: That's in general a good point, but with a quick look, RBAC has many settings that default to False, but none I see default to None.

AlanCoding:
https://github.com/jctanner/galaxy_ng/pull/1

rochacbruno:
the function on https://github.com/ansible/galaxy_ng/blob/2aa16e87c3122ea344cdee3c7da90bc4baf4fa6e/galaxy_ng/app/dynaconf_hooks.py#L732-L740 

does not evaluate the Python code on DAB, that only gets its member variables,

To evaluate the settings from DAB we would need to add the DAB settings file as a dynconf source on Pulp, or workaround by using `eval` or `exec` which has some problems

galaxy_ng/app/migrations/0055_galaxy_role_defs_to_dab_defs.py
```diff
@@ -0,0 +1,25 @@
+from django.db import migrations
+
+
+def copy_permissions_to_role_definitions(apps, schema_editor):
+    Permission = apps.get_model('auth', 'Permission')
+    RoleDefinition = apps.get_model('ansible_base.rbac', 'RoleDefinition')
```
AlanCoding, line 6: 'ansible_base.rbac' --> 'dab_rbac'

AlanCoding:
I put up https://github.com/jctanner/galaxy_ng/pull/2, and I hope that should be non-controversial. That would make a few things done in the management command here defunct, but should put us in a lot better state for migrate-ability.

galaxy_ng/app/migrations/0055_galaxy_role_defs_to_dab_defs.py
```diff
@@ -0,0 +1,25 @@
...
+            name=perm.name,
+            permissions=[perm.codename],
+            content_type=perm.content_type
+        )
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('galaxy', '0054_dab_resource_views'),
```
newswangerd, line 20: I wonder if the migrations would work if you put in the last migration from the DAB rbac app as a dependency here.

galaxy_ng/app/api/ui_v2/views.py
```diff
@@ -0,0 +1,41 @@
...
+from .filters import TeamResourcesViewFilter
+from .models import UserResourcesView
+from .models import OrganizationResourcesView
+from .models import TeamResourcesView
+from .serializers import UserSerializer
+from .serializers import OrganizationSerializer
+from .serializers import TeamSerializer
+
+
+class UserViewSet(viewsets.ModelViewSet):
```
newswangerd, line 20: All of this stuff here is unnecessary (except for the filter). We can add the resource obj on the serializer like so:

```diff
diff --git a/galaxy_ng/app/models/auth.py b/galaxy_ng/app/models/auth.py
index 0933c2b5..144c9285 100644
--- a/galaxy_ng/app/models/auth.py
+++ b/galaxy_ng/app/models/auth.py
@@ -3,6 +3,7 @@ import logging
 from django.contrib.auth import models as auth_models

 from pulpcore.plugin.models import Group as PulpGroup
+from ansible_base.resource_registry.fields import AnsibleResourceField

 log = logging.getLogger(__name__)

@@ -21,7 +22,7 @@ RH_PARTNER_ENGINEER_GROUP = f"{SYSTEM_SCOPE}:partner-engineers"
 class User(auth_models.AbstractUser):
     """Custom user model."""

-    pass
+    resource = AnsibleResourceField(primary_key_field="id")


 class GroupManager(auth_models.GroupManager):
```

```diff
diff --git a/galaxy_ng/app/api/ui/serializers/user.py b/galaxy_ng/app/api/ui/serializers/user.py
index 9d54926f..a17cb1a3 100644
--- a/galaxy_ng/app/api/ui/serializers/user.py
+++ b/galaxy_ng/app/api/ui/serializers/user.py
@@ -23,6 +23,7 @@ class GroupSerializer(serializers.ModelSerializer):

 class UserSerializer(serializers.ModelSerializer):
     auth_provider = serializers.SerializerMethodField()
+    resource = serializers.SerializerMethodField()

     class Meta:
         model = auth_models.User
@@ -37,12 +38,16 @@ class UserSerializer(serializers.ModelSerializer):
             'date_joined',
             'is_superuser',
             'auth_provider',
+            'resource',
         )
         extra_kwargs = {
             'date_joined': {'read_only': True},
             'password': {'write_only': True, 'allow_blank': True, 'required': False}
         }

+    def get_resource(self, user):
+        return user.resource.summary_fields()
+
     def get_auth_provider(self, user):
         if hasattr(user, 'social_auth') \
                 and 'social_django' in settings.INSTALLED_APPS \
```

The same thing should work for the Teams and Orgs APIs (though I don't know that we ever formally added APIs for those things, so they may have to be net new).

AlanCoding:
Can you just update the targets to use `docker compose` instead of `docker-compose` for the galaxy_collection tests? most things have been moving to that anyway.

jctanner:
> Can you just update the targets to use `docker compose` instead of `docker-compose` for the galaxy_collection tests? most things have been moving to that anyway.

No, that's a can of worms I tried opening and didn't have enough time to deal with https://github.com/pulp/oci_env/pull/162

AlanCoding:
Really good progress, since I see "standalone" is passing now. I merged https://github.com/ansible/django-ansible-base/pull/551 and that is expected to make the "community" check pass.

AlanCoding:
I would like to merge https://github.com/jctanner/galaxy_ng/pull/5 to add that test case to the mix.

rochacbruno (approved):

galaxy_ng/app/api/ui_v2/views.py
```diff
@@ -0,0 +1,173 @@
...
+
+from galaxy_ng.app.models.auth import User
+from galaxy_ng.app.models.auth import Group
+from galaxy_ng.app.models.organization import Organization
+from galaxy_ng.app.models.organization import Team
+
+
+def version_view(request):
+    data = {
+        "version": "2024-08-05T18:00:00"
```
rochacbruno, line 28: Is this going to be dynamically fetched?

profiles/dab_jwt/pulp_config.env
```diff
@@ -21,3 +21,6 @@ PULP_DYNACONF_AFTER_GET_HOOKS=["read_settings_from_cache_or_db", "alter_hostname
 
 # disable user/group modifications
 PULP_ALLOW_LOCAL_RESOURCE_MANAGEMENT=false
+
+# triggers role migrations ...
+DAB_RBAC_ENABLED="1"
```
rochacbruno, line 26: Does this need to be included in the installers' environment?

galaxy_ng/tests/integration/dab/test_dab_rbac_contract.py
```diff
@@ -0,0 +1,584 @@
...
+    assert_user_in_group(user["id"], group["id"], expected=True)
+
+    gc.patch(f"_ui/v2/users/{user['id']}/", body={"groups": old_groups})
+    assert_user_in_group(user["id"], group["id"], expected=False)
+
+
+@pytest.mark.skip(reason=(
+    "Cannot be fixed until "
+    + "https://github.com/ansible/django-ansible-base/pull/562"
+    + " stops it giving a 400"
```
AlanCoding, line 557: I'm still looking for reviews on https://github.com/ansible/django-ansible-base/pull/562 from @fosterseth @relrod and anyone else who will be involved in the next stage of that work.

Getting a 400 in this case is a fairly major breakage, and it would be a lot better to get that resolved before merging.

galaxy_ng/app/api/ui_v2/views.py
jctanner: Nah, i'm going to remove it. I added it as a debug thing last week when the UI was having issues getting the right DAB version.

profiles/dab_jwt/pulp_config.env
jctanner: No, we need to talk to QE about what marks they are or are intending to use for downstream pipelines.

jctanner: No, this is no longer necessary because the management command it triggered to run no longer exists.

PR #2207, opened 2024-07-29
misc(settings):Load DAB settings in a controversial way

rochacbruno:
No-Issue

This experiment is to check if galaxy would be able to fully "import" dab settings and it looks like it can´t


rochacbruno:
It works (loads all dab settings the way it was designed to be loaded)

It fails tests (Dab settings, if included, will override pulp_settings)

Summary: Settings must be manipulated the "Dynaconf way"

PR #2218, opened 2024-08-12
Wrote first api performance tests

stevensonmichel:
#### What is this PR doing:

Issue: AAH-####

#### Reviewers must know:


galaxy_ng/tests/performance/test_performance.py
```diff
@@ -0,0 +1,50 @@
...
+        difference = (elapsed_time - baseline) / baseline
+        if difference > threshold:
+            results.append(f"{url} exceeds threshold with {difference}")
+        else:
+            results.append(f"{url} does not exceed threshold and its difference is {difference}")
+            
+    print("httttttttttttthe result is")       
+    print(results)
+        
+    assert True
```
rochacbruno, line 49: You can use a local fixture for the client, this way other tests could use the same.

Also it looks like a good idea to use parameterized test so each URL is tested individually and we can see each one as a result in the tests

```py

@pytest.fixture(scope="function")
def api_client(ansible_config):
    config = ansible_config("admin")
    api_client = get_client(
        config=config,
        request_token=False,
        require_auth=False
    )
    return api_client


@pytest.mark.deployment_community
@pytest.mark.parametrize("url,baseline", URLS.items())
def test_api_performance(api_client, url, baseline):
    
    threshold = 0.25
    results = []

    start_time = time.time()
    resp = api_client('http://localhost:5001' + url, method='GET')
    elapsed_time = time.time() - start_time
    
    difference = (elapsed_time - baseline) / baseline
    assert difference > threshold
```

galaxy_ng/tests/performance/constants.py
```diff
@@ -0,0 +1,16 @@
...
+    '/api/_ui/v1/imports/collections/' : 0.5,
+    '/api/_ui/v1/landing-page/': 0.5,
+    '/api/_ui/v1/me/' : 0.5,
+    '/api/_ui/v1/remotes/' : 0.5,
+    '/api/_ui/v1/search/' : 0.5,
+    '/api/_ui/v1/settings/' : 0.5,
+    '/api/_ui/v1/synclists/': 0.5,
+    '/api/_ui/v1/users/': 0.5,
+    '/api/_ui/v1/tags/roles/': 0.5,
+}
```
rochacbruno, line 16: Maybe this object could get more nesting data, so we can play with different methods.

```py
URLS = {
    '/api/_ui/v1/settings/': {
         'baseline': 0.5,
         'method': 'GET'
    },
    '/api/other_url': {
        'baseline': 0.7,
        'method': 'POST',
        'payload': {...}
    }
}
```

stevensonmichel: Thank you! I will integrate your advice.

PR #2236, opened 2024-08-20
Use pytest-subtest on the rbac integration tests for more granular output

jctanner:
These tests take a long time and iterate through a lot of "parameters" without reporting anything.


bmclaughlin (approved):
This is great for the extra feedback during CI runs instead of only in the logs, :+1:

PR #2237, opened 2024-08-20
Try to parallelize rbac roles tests

jerabekjiri:
No-Issue


galaxy_ng/tests/integration/api/test_rbac_roles.py
```diff
@@ -650,7 +650,7 @@ def _get_reusable_extras(gc):
     return REUSABLE_EXTRA
 
 
-@pytest.mark.rbac_roles
+@pytest.mark.rbac_parallel_group_1
```
jctanner, line 653: I think ideally we keep the rbac_roles mark too so that we could run them all at once locally

jctanner (approved):

PR #2244, opened 2024-08-23
[WIP] jwtproxy role assignment CRUD

jctanner:
JWTProxy changes:

- support for create&delete on orgs / teams / users
- serviceindex client that will trigger POST/DELETE to the downstream service when objects are created or deleted
- basic integration test to validate the behavior of the proxy and the service client code


galaxy_ng/tests/integration/aap/test_aap_rbac.py
```diff
@@ -0,0 +1,117 @@
...
+
+    # access galaxy as the user to process claims ...
+    uc = BasicAuthClient(gc.galaxy_root, random_username, 'redhat1234')
+    new_data = uc.get(f'/api/galaxy/_ui/v2/users/?username={random_username}')
+    assert new_data['count'] == 1
+    new_user = new_data['results'][0]
+    assert new_user['username'] == random_username
+    assert not new_user['organizations']
+    new_teams = [x['name'] for x in new_user['teams']]
+    assert new_teams == [team_name]
```
AlanCoding, line 93: So basically, the user was placed in the team by using the gateway, and here, is reflected in /api/galaxy/. That's the key assertion.

AlanCoding:
The test looks good, have some comments to clean up.

I can't really review the go code, but I do see many things which are clearly required by the test.

PR #2252, opened 2024-09-04
Enable DAB service redirect auth

AlanCoding:
#### What is this PR doing:
Internal reference AAP-30011

These will be used later, but we believe it is formulated in a way that won't interfere with other deployments.

The social auth class is described in:

https://github.com/ansible/django-ansible-base/blob/devel/docs/apps/service_backed_sso.md

No-Issue

#### Reviewers must know:
Specifically, the reverse-sync setting change is made okay by this change in DAB

https://github.com/ansible/django-ansible-base/commit/2af5f34dd4887eeba65f48e01e488745e90f8f04#diff-8623b469b6e5d0b1173a741c5702169d5a06e643a8d416c860127373d662217d


AlanCoding:
DAB is on a devel pin:

https://github.com/ansible/galaxy_ng/blob/08dc2d69219afe761780cbe3f4d68a6c9aed92ad/setup.py#L113-L117

AlanCoding:
the dab_jwt tests failing look like:

```

2024-09-04T18:02:44.8187024Z galaxy_ng/tests/integration/package/test_package_install.py::test_package_install[env_vars0] [31mFAILED[0m[31m [ 99%][0m
2024-09-04T18:03:07.1038587Z galaxy_ng/tests/integration/package/test_package_install.py::test_package_install[env_vars1] [31mFAILED[0m[31m [100%][0m
```

that might maybe be okay??

h-kataria:
@elyezer please review

elyezer (commented):
Looks good assuming that this will only take effect if the resource server is defined and enabled.

rochacbruno:
Following @elyezer comment, 
> Looks good assuming that this will only take effect if the resource server is defined and enabled. 

I agree, the current code is adding for any install using keycloak or social_auth (the latest affects community galaxy)

I think it is better to change it to be conditional on RESOURCE_SERVER

AlanCoding:
> I think it is better to change it to be conditional on RESOURCE_SERVER

I agree, yes, this is the structure we want for the overall topic of settings enablement of these. But I believe we are already doing this on the level of logic within the method here:

https://github.com/ansible/django-ansible-base/blob/f9b284a74ea42d4e2b2833ebcf45a6aa431f9f68/ansible_base/resource_registry/utils/service_backed_sso_pipeline.py#L9

```python
    if (not getattr(settings, 'ENABLE_SERVICE_BACKED_SSO', False)) or (not resource_server_defined()):
        return None
```

The method `resource_server_defined` is intended to capture any nuance of the `RESOURCE_SERVER` setting.

You could also exclude the entry based (or not) in the settings definition logic (this code), but that doesn't seem necessary to me since this is also specific to certain scenarios.

jctanner (approved):

PR #2255, opened 2024-09-05
ALLOW_LOCAL_RESOURCE_MANAGEMENT truthiness

jctanner:
https://issues.redhat.com/browse/AAP-30274
https://github.com/ansible/eda-server/pull/1043


galaxy_ng/app/api/ui/v2/permissions.py
```diff
@@ -16,3 +20,77 @@ def has_permission(self, request, view):
...
+
+    To try to break down the logic when ALLOW_LOCAL_RESOURCE_MANAGEMENT=False:
+        - users CAN NOT be created/edited/deleted directly in galaxy, except ...
+        - if the request is a PATCH or a PUT that is only modifying the value
+          of is_superuser, then it can be allowed
+        - if the caller is a superuser, they can set the is_superuser flag for
+          a user to True or False
+        - if the caller is the user being modified, they can only set the flag
+          to False
+        - No user can be demoted from superuser when they are the last superuser
```
AlanCoding, line 41: Note that I don't think we need this condition, because it could be that no superusers are found in the local system, but JWT could make a user a superuser same-request later on.

galaxy_ng/app/api/ui/v2/permissions.py
```diff
@@ -16,3 +20,77 @@ def has_permission(self, request, view):
             return True
         # Otherwise, deny access
         return False
+
+
+class ComplexUserPermissions(AnsibleBaseUserPermissions):
```
AlanCoding, line 25: My current thought - it might be safer to parent from the existing `AccessPolicyBase`, as opposed to `AnsibleBaseUserPermissions`. The `has_object_permission` doesn't defer to the super class at all right now. That concerns me somewhat. I believe you have decent rules for editing the `is_superuser` field, but even with that established, it looks like edits to unrelated field (like, say, email, or first_name) are restricted to superuser. I don't know what the prior policies were, but I'm worried about throwing those out and using the DAB policies right now, when all the other access checks are done using the pulp access control models. Also, as written, I don't think `has_object_permission` will have any consistency with `has_permission`

PR #2261, opened 2024-09-11
Add dynaconf hook for aap user migration handling.

jctanner:
https://issues.redhat.com/browse/AAP-30374

We don't have a reasonable way to test this outside of a full AAP platform.


rochacbruno (approved):

rochacbruno:
On the radar for dynaconf 4 https://github.com/dynaconf/dynaconf/issues/1155#issuecomment-2346602303

galaxy_ng/app/dynaconf_hooks.py
```diff
@@ -557,6 +557,14 @@ def configure_authentication_backends(settings: Dynaconf) -> Dict[str, Any]:
     if choosen_preset in presets:
         data["AUTHENTICATION_BACKENDS"] = presets[choosen_preset]
```
rochacbruno, line 559:
```suggestion
    if choosen_preset in presets:
        data["AUTHENTICATION_BACKENDS"] = presets[choosen_preset]
    else:
        data["AUTHENTICATION_BACKENDS"] = settings.get("AUTHENTICATION_BACKENDS", [])

```

galaxy_ng/app/dynaconf_hooks.py
```diff
@@ -557,6 +557,14 @@ def configure_authentication_backends(settings: Dynaconf) -> Dict[str, Any]:
...
 
+    # https://github.com/ansible/django-ansible-base/pull/611
+    data["RENAMED_USERNAME_PREFIX"] = settings.get("RENAMED_USERNAME_PREFIX", "galaxy_")
+    backends = data.get("AUTHENTICATION_BACKENDS", [])
+    backend = "ansible_base.lib.backends.prefixed_user_auth.PrefixedUserAuthBackend"
+    if backend not in backends:
+        backends.insert(0, backend)
```
AlanCoding, line 565: I said this in the eda-server PR too, and from what I understand about this problem, I'd prefer this to be at the _end_ of the list instead of the start. Ideally, this comes into play relatively rarely, and people just login with normal username/password most of the time.

galaxy_ng/app/dynaconf_hooks.py
```diff
@@ -543,19 +543,46 @@ def configure_ldap(settings: Dynaconf) -> Dict[str, Any]:
...
+    preset_name = settings.get("AUTHENTICATION_BACKEND_PRESET")
+    if (preset_list := settings.AUTHENTICATION_BACKEND_PRESETS_DATA.get(preset_name)) is not None:
+        for item in preset_list:
+            if item not in backends:
+                backends.append(item)
+
+    # insert the AAP migrated user backend
+    prefix_backend = "ansible_base.lib.backends.prefixed_user_auth.PrefixedUserAuthBackend"
+    if prefix_backend not in backends:
+        backends.append(prefix_backend)
```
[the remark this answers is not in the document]
rochacbruno: nm, I just saw https://github.com/ansible/galaxy_ng/pull/2261#pullrequestreview-2301328476

galaxy_ng/app/auth/ldap.py
```diff
@@ -40,3 +40,22 @@ class GalaxyLDAPBackend(LDAPBackend):
...
+
+    @property
+    def prefix(self):
+        return settings.get("RENAMED_USERNAME_PREFIX")
+
+    def authenticate(self, *args, **kwargs):
+        if username := kwargs.get("username"):
+            if username.startswith(self.prefix):
+                kwargs["username"] = username[len(self.prefix):]
+                return super().authenticate(*args, **kwargs)
```
rochacbruno, line 55: Python 3.9 added a removeprefix to `str` https://docs.python.org/3.9/library/stdtypes.html#str.removeprefix 

```suggestion
        if username := kwargs.get("username"):
            kwargs["username"] = username[len(self.prefix):]
            return super().authenticate(*args, **kwargs)
```

galaxy_ng/app/dynaconf_hooks.py
```diff
@@ -543,19 +543,46 @@ def configure_ldap(settings: Dynaconf) -> Dict[str, Any]:
...
-    if choosen_preset in presets:
-        data["AUTHENTICATION_BACKENDS"] = presets[choosen_preset]
+    # start with the default pulp settings
+    backends = settings.get("AUTHENTICATION_BACKENDS", [])
+
+    # merge in backends set by the previous hooks ...
+    if (default_list := data.get("AUTHENTICATION_BACKENDS")) is not None:
+        for item in default_list:
+            if item not in backends:
+                backends.append(item)
```
rochacbruno, line 564: suggestion.
```suggestion
        backends.extend([item for item in default_list if item not in backends])
```

galaxy_ng/app/auth/ldap.py
AlanCoding: This goes the opposite direction of the DAB mixin.

https://github.com/ansible/django-ansible-base/blob/devel/ansible_base/lib/backends/prefixed_user_auth.py

That adds the prefix to the username, as opposed to removing it. I can't find an explanation for why galaxy_ng specifically is doing this.

galaxy_ng/app/settings.py
```diff
@@ -341,7 +341,9 @@
     'ldap': [
         "galaxy_ng.app.auth.ldap.GalaxyLDAPBackend",
         "django.contrib.auth.backends.ModelBackend",
-        "pulpcore.backends.ObjectRolePermissionBackend"
+        "pulpcore.backends.ObjectRolePermissionBackend",
+        "galaxy_ng.app.auth.ldap.PrefixedLDAPBackend",
```
rochacbruno, line 345: Isnt this overriding the `GalaxyLDAPBackend` added on the top? does it mean the system will attempt LDAP authentication twice?

galaxy_ng/app/auth/ldap.py
rochacbruno: On LDAP users has no prefix, then the LDAP backend must remove it to succed the search?

AlanCoding: In the migration, a user "rick" is a _conflict_ because it exists both in controller & Hub. So the user in Hub will be renamed to "hub_rick", and "rick" will track the user from controller.

Scenario we care about is that someone logs in with "rick" username and the password which now only works for "hub_rick". So in the case of password auth, we would have to add the prefix to the username.

In the case of social auth, the client would still give "rick" username and expect it to work. Of course, this user will only be known as "rick" to the external auth provider. But IIRC, @newswangerd 's code sample still ran the same prefixing logic as password auth. I agree that might not make complete sense, but this diff doesn't make sense to me either.

galaxy_ng/app/auth/ldap.py
```diff
@@ -40,3 +40,22 @@ class GalaxyLDAPBackend(LDAPBackend):
...
+
+class PrefixedLDAPBackend(GalaxyLDAPBackend):
+
+    @property
+    def prefix(self):
+        return settings.get("RENAMED_USERNAME_PREFIX")
+
+    def authenticate(self, *args, **kwargs):
+        if username := kwargs.get("username"):
+            if username.startswith(self.prefix):
```
AlanCoding, line 53: Wouldn't this refuse to authenticate a user if their username didn't start with the prefix?

What if a user is not a conflict. It only ever existed in Hub, and then it is migrated. That user won't have the prefix. The way this is changing looks like it might block such a user from being able to log in with LDAP. Because this replaces the old `GalaxyLDAPBackend`, instead of adding to it.

I'm also somewhat expecting to save a variable on `self` to indicate whether or not the prefix was removed in this step, and then that's what we would use to determine whether or not the prefix is added back in `get_or_build_user`.

jctanner: The first problem has been fixed. The second comment is more complex...

Setting self.prefix_removed does the right thing when authenticating with prefix_username:password, but won't do the right thing when authenticating with username:password ... it ends up creating a second user with a new uid.

PR #2266, opened 2024-09-17
Make a simple/self-contained docker-compose to use for AAP dev.

jctanner:
We need a way to spin up the stack for AAP dev that is simple and doesn't involve oci-env or any third party deps besides docker-compose.


rochacbruno (approved):
🎉

PR #2275, opened 2024-09-25
The Big Cleanup

rochacbruno:
`rm ...`
- All related to pulp plugin_template (except those scripts reused by `ci_full`
- All related to the legacy `./compose` stack
- Useless Makefile targets


jctanner (approved):
finally ...

PR #2308, opened 2024-10-11
Community cleanup

jctanner:
All of this code was temporary for the beta switchover.


jctanner:
Given that so many integration tests are relying on the v1/sync endpoint, this isn't going to be feasible right now.

PR #2326, opened 2024-10-21
dev: add community compose

rochacbruno:
Add dev/compose/community.yaml

### DIFF
--- a/dev/compose/community.yaml
+++ b/dev/compose/community.yaml
@@ -60,7 +60,7 @@ x-common-env: &common-env
   # disable user/group modifications
   PULP_ALLOW_LOCAL_RESOURCE_MANAGEMENT: 'false'
 
-  # role content workaround ..
+  # role content workaround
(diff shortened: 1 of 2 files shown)


dev/compose/community.yaml
```diff
@@ -0,0 +1,303 @@
...
+      - "../../:/app"
+    environment:
+      <<: *common-env
+    user: root
+    command: |
+      bash -c "
+        /src/galaxy_ng/dev/compose/bin/wait /var/lib/pulp/.migrated;
+        /src/galaxy_ng/dev/compose/bin/devinstall;
+
+        cp -r /src/galaxy_ng/profiles/community/galaxy-importer /etc/galaxy-importer;
```
jctanner, line 236: We should be able to set GALAXY_IMPORTER_CONFIG to point directly at /src/galaxy_ng/profiles/community/galaxy-importer/galaxy-importer.cfg and then we have one less line here and dont have to worry about live reloads

jctanner (approved):

PR #2327, opened 2024-10-22
recreate gateway users during test setup as needed.

jctanner:
https://issues.redhat.com/browse/AAP-33988


jerabekjiri (approved):
LGTM, but question: why not use the [BasicAuthClient](https://github.com/ansible/galaxykit/blob/main/galaxykit/client.py#L545) for the GW requests?

PR #2335, opened 2024-10-24
Enable flake8-datetime plugin for ruff

cutwater:
But add `noqa` for existing code, since fixing naive to aware datetime may cause bugs.

### DIFF
--- a/dev/common/dependency_bumper.py
+++ b/dev/common/dependency_bumper.py
@@ -192,15 +192,16 @@ def main():
 
         # make a unique checkout for this new patch
         branch_checkout = os.path.join(workdir, f'galaxy_ng.{cfg.branch}')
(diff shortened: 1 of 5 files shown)


dev/common/dependency_bumper.py
```diff
@@ -192,15 +192,16 @@ def main():
 
         # make a unique checkout for this new patch
         branch_checkout = os.path.join(workdir, f'galaxy_ng.{cfg.branch}')
-        ts = datetime.datetime.now().isoformat().split('.')[0].replace('-', '_').replace(':', '_')
+        ts = datetime.datetime.now().strftime("%Y_%m_%dT%H_%M_%S")  # noqa: DTZ005
```
rochacbruno, line 195: There is also this option
```python
f"{datetime.now():%Y_%m_%dT%H_%M_%S}"
'2024_10_25T13_48_28'
```

rochacbruno (approved):

PR #2407, opened 2025-01-21
[docker-compose] certified-sync CI

jerabekjiri:


dev/compose/certified-sync.yaml
```diff
@@ -536,8 +536,10 @@ services:
       - '../nginx/certified-sync/insights.conf:/etc/nginx/nginx.conf:ro'
 
 volumes:
-  var_lib_pulp:
-    name: var_lib_pulp
+  var_lib_pulp_standalone:
+    name: var_lib_pulp_standalone
+  var_lib_pulp_insights:
+    name: var_lib_pulp_insights
```
rochacbruno, line 542: 👍🏻 nice! so we dont have conflicts when running multiple profiles in the same system.

we probably need to do that also on `aap` and `community` profiles.

rochacbruno (approved):

dev/compose/certified-sync.yaml
jerabekjiri: Good idea :+1: 
I did this mainly because two profiles shared the same data and that made no sense when testing the sync

PR #2428, opened 2025-03-05
Configure SELinux on compose env

fao89:
https://docs.docker.com/engine/storage/bind-mounts/#options-for---volume

No-Issue

Signed-off-by: Fabricio Aguiar <fabricio.aguiar@gmail.com>


rochacbruno:
It is a simple change, we probably want this on 4.9 and 4.10 branches

PR #2449, opened 2025-03-10
Use galaxy user instead of root

fao89:
After merging this PR, you might get some permission errors, as volumes might have root owned files.
Cleaning the volumes would fix that
```
docker system prune -a --volumes && docker volume rm $(docker volume ls -qf dangling=true)
```
You might also need to remove some "cache" dirs under galaxy_ng
```
# galaxy_ng
ls -al | grep root
# delete what you got from ^
sudo rm -rf <ROOT_OWNED>
```

### DIFF
--- a/Dockerfile.rhel8
+++ b/Dockerfile.rhel8
@@ -1,6 +1,7 @@
 FROM registry.access.redhat.com/ubi8
 
 ARG GIT_COMMIT
+ARG USER_ID=1000
 
 ENV LANG=en_US.UTF-8 \
     PYTHONUNBUFFERED=1 \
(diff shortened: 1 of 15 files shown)


Dockerfile.rhel8
```diff
@@ -1,6 +1,9 @@
 FROM registry.access.redhat.com/ubi8
 
 ARG GIT_COMMIT
+ARG USER_ID=1000
+ARG USER_NAME=galaxy
+ARG USER_GROUP=galaxy
```
rochacbruno, line 6: I am not sure if cloud deployments are using this image, so we need a review from @drodowic to check if this ok to add this defaults.

drodowic: Cloud deployments do use this image, however deployments to OpenShift will run the container with an arbitrary UID so permissions will be dependent on the group

rochacbruno: So is it ok to go with this change?

drodowic: I see lots of permissions errors in the devshift PR check. But it's hard to say if that's due to these changes or just because the tests are broken.
Can we get https://github.com/ansible/galaxy_ng/pull/2452 merged and then rebase this PR, so that it can be properly tested against the CRC ephemeral environment?

drodowic: https://github.com/ansible/galaxy_ng/pull/2452 has been merged. Please rebase.

fao89: I've rebased, did some tweaks and got CI green now

drodowic (approved):

rochacbruno (approved):

PR #2468, opened 2025-03-13
Remove unused files

fao89:


.bumpversion.cfg
rochacbruno, line 1: I think Makefile references this file 
https://github.com/ansible/galaxy_ng/blob/1d97bc4a0a262eb939414a2615dc37a4cf167b50/Makefile#L46-L65

setup.cfg
```diff
@@ -1,33 +0,0 @@
...
-search = version = "{current_version}"
-replace = version = "{new_version}"
-
-[bumpversion:file:./galaxy_ng/app/__init__.py]
-search = version = "{current_version}"
-replace = version = "{new_version}"
-
-[bumpversion:file:./galaxy_ng/__init__.py]
-search = __version__ = "{current_version}"
-replace = __version__ = "{new_version}"
```
rochacbruno, line 33: Or maybe this, not sure which one the bump2version uses

fao89: this one seems to be the one Makefile uses, as the version is right

.bumpversion.cfg
```diff
@@ -1,21 +0,0 @@
-[bumpversion]
-current_version = 4.10.0dev
```
fao89, line 2: outdated version, so it wasn't being used

rochacbruno (approved):

PR #2476, opened 2025-03-18
Replace setup.py with pyproject.toml

fao89:


pyproject.toml
```diff
@@ -1,6 +1,79 @@
...
+version = { attr = "galaxy_ng.__version__" }
+
+[tool.setuptools.packages.find]
+include = ["galaxy_ng*"]
+exclude = ["galaxy_ng.tests*"]
+
 [tool.black]
 line-length = 100
-target-version = ["py36", "py37"]
+target-version = ["py39", "py310", "py311", "py312"]
```
rochacbruno, line 76: we currently require minimum py311 so we can set to py311

rochacbruno:
Add the ui/build.py to https://github.com/ansible/galaxy_ng/blob/91c47c4602743b0b7262e072f3f3eeb6c046fb7f/sonar-project.properties#L5

galaxy_ng/ui/build.py
```diff
@@ -0,0 +1,71 @@
...
+class SDistCommand(_SDistCommand):
+    def run(self):
+        prepare_static()
+        return super().run()
+
+
+class BuildPyCommand(_BuildPyCommand):
+    def run(self):
+        prepare_static()
+        return super().run()
```
rochacbruno, line 71: I wonder if this custom build entrypoints for preparing UI are still required?

We are not releasing to PyPI anymore, we dont inspect people to rely on the UI coming from the package.

AFAIK galaxy.ansible.com has UI deployed directly from UI repo, can you confirm @drodowic 

Maybe we can just get rid of those custom entrypoints

fao89: I think we should have a separate PR for that then,
so we can have it highlighted on the changelog that the UI was removed

pyproject.toml
```diff
@@ -1,6 +1,79 @@
...
+    "dynaconf>=3.2.10",
+    "galaxy-importer>=0.4.27,<0.5.0",
+    "insights_analytics_collector>=0.3.0",
+    "marshmallow<4.0.0,>=3.6.1",
+    "pulp-container>=2.19.2,<2.20.0",
+    "pulp_ansible==0.23.1",
+    "pulpcore>=3.49.0,<3.50.0",
+    "social-auth-app-django>=5.2.0",
+    "social-auth-core>=4.4.2",
+]
```
fao89, line 52: we could do it dynamically  as:  https://github.com/ansible/aap-gateway/blob/devel/pyproject.toml#L27-L31

setup.py
```diff
@@ -1,179 +0,0 @@
...
-# https://softwareengineering.stackexchange.com/questions/223634/what-is-meant-by-now-you-have-two-problems
-def strip_package_name(spec):
-    operators = ["=", ">", "<", "~", "!", "^", "@"]
-    for idc, char in enumerate(spec):
-        if char in operators:
-            return spec[:idc]
-    return spec
-
-
-unpin_requirements = os.getenv("LOCK_REQUIREMENTS") == "0"
```
rochacbruno, line 129: It looks like we rely on `LOCK_REQUIREMENTS` on a lot of places and I am not sure how it would be with pyproject as it is a static file.

https://github.com/search?q=repo%3Aansible%2Fgalaxy_ng%20%20LOCK_REQUIREMENTS&type=code

setup.py
```diff
@@ -1,179 +0,0 @@
...
-        return super().run()
-
-
-class BuildPyCommand(_BuildPyCommand):
-    def run(self):
-        self.run_command("prepare_static")
-        return super().run()
-
-
-django_ansible_base_branch = os.getenv('DJANGO_ANSIBLE_BASE_BRANCH', '2025.3.7')
```
rochacbruno, line 90: AFAIK `DJANGO_ANSIBLE_BASE_BRANCH` may be set by some testing pipelines, for example to point to `devel` how would we do it with pyproject?

galaxy_ng/ui/build.py
drodowic: you are correct that galaxy.ansible.com deploys the UI seperately, and this built in UI in not utilized in the galaxy pipeline.

setup.py
fao89: not sure, I need to research more

setup.py
fao89: not sure, I need to research more

setup.py
fao89: actually, I realize both `DJANGO_ANSIBLE_BASE_BRANCH` and `LOCK_REQUIREMENTS` aren't `pyproject.toml` issues, we can handle those at dev environment level

PR #2520, opened 2025-04-01
Unlink Dockerfile

fao89:
Removing the symlink


rochacbruno:
Lets confirm with @drodowic if this symlink is used on cloud deployments

PR #2567, opened 2025-05-07
Add reload option to docker binaries

fao89:
Issue: AAP-44318

### DIFF
--- a/dev/compose/Dockerfile.dev
+++ b/dev/compose/Dockerfile.dev
@@ -10,7 +10,7 @@ USER root
 RUN LANG=C dnf -y --disableplugin=subscription-manager install nc
 
 USER galaxy
-RUN ${VIRTUAL_ENV}/bin/python3 -m pip install ipython ipdb rpdb django-extensions pulp-cli
+RUN ${VIRTUAL_ENV}/bin/python3 -m pip install hupper ipython ipdb rpdb django-extensions pulp-cli
 
 RUN /app/dev/compose/signing/setup_gpg_workarounds.sh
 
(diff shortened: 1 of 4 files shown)


dev/compose/Dockerfile.dev
```diff
@@ -10,7 +10,7 @@ USER root
 RUN LANG=C dnf -y --disableplugin=subscription-manager install nc
 
 USER galaxy
-RUN ${VIRTUAL_ENV}/bin/python3 -m pip install ipython ipdb rpdb django-extensions pulp-cli
+RUN ${VIRTUAL_ENV}/bin/python3 -m pip install hupper ipython ipdb rpdb django-extensions pulp-cli
```
fao89, line 13: https://github.com/pulp/pulpcore/blob/169ca42cfb2e16a72a1af9dd2f58bb09a8d71e14/pulpcore/tasking/entrypoint.py#L22

docker/bin/start-api
```diff
@@ -32,4 +33,8 @@ if [[ -n "${GUNICORN_LOGGER_CLASS}" ]]; then
     GUNICORN_OPTIONS+=(--logger-class "${GUNICORN_LOGGER_CLASS}")
 fi
 
+if [[ "${RELOAD_ENABLED}" == 'yes' ]]; then
+    GUNICORN_OPTIONS+=(--reload)
+fi
```
fao89, line 38: adding --reload to gunicorn

rochacbruno: Currently we are starting the API with the `reloader` script https://github.com/ansible/galaxy_ng/blob/62871b98e50499ef369f9165e5e4f9cbc0051dd3/dev/compose/aap.yaml#L202 

Will this replace the current reloader? or both can work together?

fao89: I haven't tested yet, the idea of this PR is having a way to reload from other dev environments: operator, aap-dev, ...

fao89: I tested it locally,
it can replace and it can work together.
So far it is disabled on this PR, should I enable it by adding `RELOAD_ENABLED: yes` on the compose files?

rochacbruno (approved):

PR #2656, opened 2025-07-30
Make sessionId secure and able to be overridden

bmclaughlin:
#### What is this PR doing:

Makes `sessionId` secure via galaxy_ng's settings.py and allows for it to be overridden by an environment variable `SESSION_COOKIE_SECURE`.

Issue: AAP-47264

### DIFF
--- a/galaxy_ng/app/settings.py
+++ b/galaxy_ng/app/settings.py
@@ -453,3 +453,5 @@ RENAMED_USERNAME_PREFIX = "galaxy_"
 
 # feature flags
 FLAGS = {}
+
+SESSION_COOKIE_SECURE = True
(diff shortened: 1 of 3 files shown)


galaxy_ng/app/settings.py
```diff
@@ -453,3 +453,5 @@
 
 # feature flags
 FLAGS = {}
+
+SESSION_COOKIE_SECURE = os.environ.get('SESSION_COOKIE_SECURE', True)
```
rochacbruno, line 457: Envvars are strings, so this will evaluate to truthy even if someone exports `SESSION_COOKIE_SECURE=False`

This kind of problem is already handled by dynaconf on pulp.

You can simply default it to `True` and then use `PULP_SESSION_COOKIE_SECURE=true` on the envvar.


```suggestion
SESSION_COOKIE_SECURE = True
```

rochacbruno (changes requested):

rochacbruno (approved):

